PR #80, opened 2015-09-18
[WIP] Cli Refactoring

kevgliss:
The number of commands and the code supporting Lemur's cli is getting a bit out of hand. This is
an attempt to decompose some of the commands into more concise commands. It also tries to move the code closer to the services it directly uses.


kevgliss:
Rebuild

PR #92, opened 2015-09-21
Rename SSL to TLS in many places

konklone:
This changes "SSL" to "TLS" in many places (not 100% comprehensive, but the important places).

Since the SSL protocol (last formally updated in 1999) is now fully deprecated in the browser and [by the IETF](https://tools.ietf.org/html/rfc7568), we should be doing our best to move to "TLS" as the primary descriptor.


kevgliss:
Thanks for this, there are many places where TLS should be used over SSL...

Issue #119, opened 2015-10-07
Key rotation


rpicard:
What is the process for rotating the encryption key used with Lemur? Lets say I accidentally copy it into an IRC chat. What are my next steps?

If there isn't any tooling built-in to help with this, I think it would be a great thing to have. I haven't done much research into the best way to do it, but a good starting point might be: https://cryptography.io/en/latest/fernet/#cryptography.fernet.MultiFernet

It might make sense to look into this at the same time as issue #117 is being addressed.

rpicard:
I can probably contribute some code to this if we come to a consensus on what this should look like.

kevgliss:
Closed by #123

PR #123, opened 2015-10-12
[WIP] Use MultiFernet for encryption

### LINKED ISSUE #117: Certificates encrypted with static IV per key
The IV is static per key at least. Lemur is using sqlalchemy_utils to encrypt certificates. This in turn
encrypts with AES in CBC mode.

https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/encrypted.py#L56

Given a single key, it will use the SHA256 hash of that key for all encryption. It looks like it will use the first 16 bytes of that hash as the IV for each operation.

rpicard:
Facilitates key rotation and uses more secure encryption than what
sqlalchemy-utils does.

Fixes #117 and #119.

This is still a work in progress. I'm having some trouble testing locally. When I try to create new certificates, authorities, roles, etc. I end getting "not created" errors for what seem to be unrelated reasons. I'm hoping you can take a look and see how it works yourself.

TODO:

- [ ] Migration script or something to deal with legacy users who have different keys that won't work with Fernet.
- [x] Should we pluralize the encryption key configuration key now that we support more than one? Maybe we could have a separate configuration key called "LEGACY_KEYS" or something to make it clear which is actively being used for encryption and which are just there for rotation.
- [x] Should we throw an exception or something when keys are not Fernet compatible?

### DIFF
--- a/docs/administration/index.rst
+++ b/docs/administration/index.rst
@@ -87,17 +87,23 @@ Basic Configuration
         >>> secret_key = secret_key + ''.join(random.choice(string.digits) for x in range(6))
 
 
-.. data:: LEMUR_ENCRYPTION_KEY
+.. data:: LEMUR_ENCRYPTION_KEYS
     :noindex:
 
-        The LEMUR_ENCRYPTION_KEY is used to encrypt data at rest within Lemur's database. Without this key Lemur will refuse
-        to start.
+        The LEMUR_ENCRYPTION_KEYS is used to encrypt data at rest within Lemur's database. Without a key Lemur will refuse
+        to start. Multiple keys can be provided to facilitate key rotation. The first key in the list is used for
+        encryption and all keys are tried for decryption until one works. Each key must be 32 URL safe base-64 encoded bytes.
 
-        See `LEMUR_TOKEN_SECRET` for methods of secure secret generation.
+        Running lemur create_config will securely generate a key for your configuration file.
+        If you would like to generate your own, we recommend the following method:
+
+            >>> import os
+            >>> import base64
+            >>> base64.urlsafe_b64encode(os.urandom(32))
 
     ::
 
-        LEMUR_ENCRYPTION_KEY = 'supersupersecret'
+        LEMUR_ENCRYPTION_KEYS = ['1YeftooSbxCiX2zo8m1lXtpvQjy27smZcUUaGmffhMY=', 'LAfQt6yrkLqOK5lwpvQcT4jf2zdeTQJV1uYeh9coT5s=']
 
 
 Certificate Default Options
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -4,8 +4,8 @@ Frequently Asked Questions
 Common Problems
 ---------------
 
-In my startup logs I see *'Aborting... Lemur cannot locate db encryption key, is LEMUR_ENCRYPTION_KEY set?'*
-  You likely have not correctly configured **LEMUR_ENCRYPTION_KEY**. See
+In my startup logs I see *'Aborting... Lemur cannot locate db encryption key, is LEMUR_ENCRYPTION_KEYS set?'*
+  You likely have not correctly configured **LEMUR_ENCRYPTION_KEYS**. See
   :doc:`administration/index` for more information.
 
 
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -13,9 +13,7 @@ from cryptography.hazmat.backends import default_backend
 from sqlalchemy.orm import relationship
 from sqlalchemy import event, Integer, ForeignKey, String, DateTime, PassiveDefault, func, Column, Text, Boolean
 
-from sqlalchemy_utils import EncryptedType
-
-from lemur.utils import get_key
+from lemur.utils import Vault
 from lemur.database import db
 from lemur.plugins.base import plugins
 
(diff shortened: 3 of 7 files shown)


rpicard:
@kevgliss If you wouldn't mind taking a look when you have a minute. I have a feeling at least part of the errors I'm seeing have to do with my local environment and not necessarily the changes I've made.

setup.py
```diff
@@ -32,7 +32,6 @@
     'Flask-Bcrypt==0.6.2',
     'Flask-Principal==0.4.0',
     'Flask-Mail==0.9.1',
-    'SQLAlchemy-Utils==0.30.11',
```
kevgliss: It looks like I also use the JSON type from SQLAlchemy-Utils so we should leave it in for the moment.

rpicard: :+1:

kevgliss:
I think this should be okay. I would just add back in SQLAlchemy and see if the tests pass. You could also post a stack trace to see if we can narrow down the errors you are seeing locally.

rpicard:
I still can't seem to save anything locally. Here is the content of `lemur.log` after trying to save a new issuer.

https://p.robert.io/FeudHiltsProvoke

kevgliss:
You will need to make an authority before you can issue any certificates. 

```
2015-10-12 16:27:28,996 ERROR: 400: Bad Request [in /Users/robert.picard/Code/rpicard/lemur/lemur/common/utils.py:60]
  File "/Users/robert.picard/Code/rpicard/lemur/lemur/common/utils.py", line 46, in wrapper
    resp = f(*args, **kwargs)
  File "/Users/robert.picard/Code/rpicard/lemur/lemur/authorities/views.py", line 302, in put
    args = self.reqparse.parse_args()
  File "/Users/robert.picard/.virtualenvs/lemur/lib/python2.7/site-packages/flask_restful/reqparse.py", line 295, in parse_args
    value, found = arg.parse(req, self.bundle_errors)
  File "/Users/robert.picard/.virtualenvs/lemur/lib/python2.7/site-packages/flask_restful/reqparse.py", line 215, in parse
    self.handle_validation_error(ValueError(error_msg), bundle_errors)
  File "/Users/robert.picard/.virtualenvs/lemur/lib/python2.7/site-packages/flask_restful/reqparse.py", line 144, in handle_validation_error
    flask_restful.abort(400, message=msg)
  File "/Users/robert.picard/.virtualenvs/lemur/lib/python2.7/site-packages/flask_restful/__init__.py", line 30, in abort
    original_flask_abort(http_status_code)
  File "/Users/robert.picard/.virtualenvs/lemur/lib/python2.7/site-packages/werkzeug/exceptions.py", line 605, in __call__
    raise self.mapping[code](*args, **kwargs)
```

Should return a message complaining what might be missing from the request. Which authority are you trying to use? The only one currently packaged with Lemur is the `verisign` plugin. Thus you would only be able to issue certificates if you had a `verisign` account. 

Are you doing this from the UI or the API? In either case ff you could give me the JSON you are sending to the API, I can probably spot whats missing.

rpicard:
Here's what I'm sending:

```
{
    "caDN": {
        "commonName": "something",
        "country": "US",
        "location": "Los Gatos",
        "organization": "Netflix",
        "organizationalUnit": "Operations",
        "state": "CA"
    },
    "caDescription": "something",
    "caName": "InternalCA",
    "caSensitivity": "medium",
    "caSigningAlgo": "sha256WithRSA",
    "caType": "root",
    "extensions": {
        "subAltNames": {
            "names": []
        }
    },
    "keyType": "RSA2048",
    "ownerEmail": "ca@ca",
    "pluginName": "verisign-issuer",
    "subAltType": null,
    "subAltValue": null,
    "validityEnd": "2015-10-22T07:00:00.000Z",
    "validityStart": "2015-10-02T07:00:00.000Z"
}
```

kevgliss:
Can you try with `California` instead of `CA` for state? I thought I changed all the defaults but may have missed it.

rpicard:
Tried with the same result.

kevgliss:
Which branch/commit are you working from? When you make the request there should be a more descriptive message in the body of the response. 

Also I noticed in your stack trace you are doing a PUT request. 

```
  File "/Users/robert.picard/Code/rpicard/lemur/lemur/authorities/views.py", line 302, in put
```

The JSON you are sending I would expect to be handled by a POST.

See: https://github.com/Netflix/lemur/blob/master/lemur/authorities/views.py#L302 for the variables allowed in a PUT request (editing an authority).

kevgliss:
I just tested this out and it looks good. Everything encrypts and decrypts like I would expect. I think throwing an exception for improperly formatted keys makes sense. After that is added I will go ahead and merge this in. I will put the migration script in a separate pull request. Thanks!

rpicard:
Thanks a lot. I'm not sure what my issue locally is. I'm doing things through the UI, so I would expect it to `POST` based on what you said.

In any case, I'll add in that exception and finish this up.

rpicard:
Fernet will actually throw an exception if the key passed to it is not valid. We could add some checks at startup or something to make it easier to debug, but I think as long as the requirement is documented we'll be alright. The exception thrown by Fernet is pretty clear too.

rpicard:
Maybe we should have a comment in the generated configuration file to document the requirement. I'll add it to the docs too.

rpicard:
Current status: All should be ready to go pending a decision on how best to handle the key requirements. As of now, I've documented the requirement on the administration page.

As it stands, the application will still start with a bad key, but Fernet will throw an exception when it tries to encrypt something. Is there a better channel to communicate errors with users?

kevgliss:
I think that should be enough - we generate good defaults and point out the requirements for those who wander off the beaten path. I am going to merge this in, thanks for the all the work. Feel free to hit me up on gitter if you are still struggling with creating authorities, or open an issue and I can see if I can tease out any more clues.

rpicard:
Awesome. Thanks for the help in putting this together.

PR #134, opened 2015-10-29
Removing hyphen from in-active.

monkeysecurity:
`inactive` is a word.  in-active is ... something else.


kevgliss:
This is so sad.

PR #155, opened 2015-11-27
Fix requires.io badge

Joe8Bit:
Update README.rst to point to correct requires.io badge. This is a SUPER minor change, but saw it and thought might as well fix it1


kevgliss:
Thanks!

Issue #160, opened 2015-12-01
Documentation


forkd:
I followed every installation steps described in Quickstart guide [1], but I found it quite confusing: could be improved/updated.

In fact, I reach the login screen, but I cannot get into Lemur.  Since the guide is deprecated (v.0.1.3) compared to the latest release (v.0.1.5), I could have missed something .

Despite of it, I'm sending you the installation steps I took.  This way, you could show me what I'm doing wrong, and improve that guide, if want so.

#### Steps I took

> I used XUbuntu 14.04 fully updated here.

Install packages:
```shell
$ sudo apt-get install nodejs-legacy python-pip python-dev libpq-dev build-essential libssl-dev libffi-dev nginx git supervisor npm postgresql
$ sudo pip install -U virtualenv
```

Create the destination directory:
```shell
$ sudo mkdir /www
$ cd /www
```

Download and "install" the latest release of Lemur (v.0.1.5):
```shell
$ sudo wget https://github.com/Netflix/lemur/archive/0.1.5.zip
$ sudo unzip 0.1.5.zip
$ sudo mv lemur-0.1.5.zip lemur
$ sudo rm 0.1.5.zip
$ sudo chown -R <YOUR_USER> lemur/
```

Prepare the virtualenv:
```shell
$ virtualenv lemur
$ source /www/lemur/bin/activate
$ cd lemur
```

Open `Makefile` and comment those lines related to `setup-git` rule.

Create Lemur's configuration file and prepare the database --`control-d` to exit Postgres' shell:
```shell
$ lemur create_config
$ sudo -u postgres psql postgres
# \password postgres
Enter new password: lemur
Enter it again: lemur
<control-d>
$ sudo -u postgres createdb lemur
$ lemur db init
```

> Here's one problem, because Lemur stopped with an authentication error:
> `FATAL: password authentication failed for user "lemur"`
> This is how I solved:
> ```
$ sudo -u postgres psql postgres
\password lemur
Enter new password: lemur
Enter it again: lemur
<control-d>
```

Initialize Lemur --I used the password `lemur`:
```shell
$ lemur init
```
> [+] Created 'admin' role
> [+] Added a 'lemur' user and added it to the 'admin' role!

Finally, edit `/etc/nginx/sites-available/default` --create a backup copy before making any changes--, accordingly to [1].

> Before starting Lemur, I restarted Nginx:
> `$ sudo service nginx restart`

Start Lemur:

```shell
$ lemur start
```

[...]

PR #164, opened 2015-12-01
updated quickstart guide

forkd:
Just updated quickstart to follow a logical sequence: first install all dependencies, then prepare environment, then start installation.

This was motivated after the resolution of issue #160.


kevgliss:
Thanks!

PR #178, opened 2015-12-10
[WIP] Adding ACME Support

kevgliss:
Would love feed back on this.

Open questions:

- Do you need auto registration of domains?
- Is registering a authenticated user before hand a big deal?
- Would each authority have it's own authenticated user (public key)?


harmw:
Do you need auto registration of domains?
No, Lemur should be able to use _some_ DNS API (Cloudflare/Route53) to create the required validation records. Considering DNS validation is fast and relatively easy to implement. To me, registration of domains is a completely different use-case.

    Is registering a authenticated user before hand a big deal?
Probably not, though eventually it'll be nicer to let Lemur do that when initializing a new CA which utilizes this plugin.

    Would each authority have it's own authenticated user (public key)?
Yes, as it is no extra burden for the operator. Lemur will take care of it in the background, right? Using the same user/key for the entire plugin implementation, and using that to talk with different endpoints is not save.

harmw:
Travis currently fails on python 3.3 and 3.4. @kevgliss are you planning to pick this up again? Support for ACME (or rather: letsencrypt) would be awesome!

kevgliss:
I was stalled because of a bug in the letsencrypt library, not sure if it's been resolved. Still interested in talking this on as I think it would be potentially very useful to quite a few folks. Will look into whether I can move forward with this plugin.

PR #186, opened 2015-12-21
Add Google SSO

robert-addepar:
This pull request adds Google SSO support. There are three main changes:

1. Add the Google auth view resource and update the Angular app.js to
use it
2. Change the "login with example" button to "login with google"
3. Make passwords optional when creating a new user. This allows an admin
to create a user without a password so that they can only login via Google.

I'm guessing you'll have some suggestions to make these changes more generally applicable.

### DIFF
--- a/lemur/static/app/angular/app.js
+++ b/lemur/static/app/angular/app.js
@@ -18,7 +18,7 @@ var lemur = angular
     'angular-clipboard',
     'ngFileSaver'
   ])
-  .config(function ($stateProvider, $urlRouterProvider, $authProvider) {
+  .config(function ($stateProvider, $urlRouterProvider, $authProvider, AuthenticationService) {
     $urlRouterProvider.otherwise('/welcome');
 
     $stateProvider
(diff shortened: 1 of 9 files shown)


kevgliss:
This looks good, I would ask that you remove the google specific login/client-id from the javascript until I can figure out a better method of doing this dynamically. That being said if you are up for attempting to make those changes in this or another pull request that would great as well.

robert-addepar:
Updated to remove the Google changes. I plan on doing another pull request for the server side stuff to support the idea of active providers.

rpicard:
Ended up just pushing to this same pull request. Added a providers endpoint to return the configuration details for active providers.

robert-addepar:
There are a couple of TODOs in the commit that adds the providers endpoint. I'm not sure what the appropriate values for those keys would be.

lemur/static/app/angular/app.js
```diff
@@ -27,17 +27,23 @@ var lemur = angular
...
-      url: 'http://localhost:8000/api/1/auth/ping',
-      redirectUri: 'http://localhost:3000/',
-      clientId: 'client-id',
-      responseType: 'code',
-      scope: ['openid', 'email', 'profile', 'address'],
-      scopeDelimiter: ' ',
-      authorizationEndpoint: 'https://example.com/as/authorization.oauth2',
-      requiredUrlParams: ['scope']
-    });
+    active_providers = AuthenticationService.get_providers();
```
kevgliss: Because get_providers returns a 'promise' or async result you probably want to wrap your logic after the providers have returned: 

```
AuthenticationService.get_providers().then(function (providers) {
    // do $authProvider manipulation
});
```
Otherwise when you call ```active_providers = AuthenticationService.get_providers();``` ```active_providers``` will be undefined until some later point

robert-addepar: Updated.

PR #187, opened 2015-12-27
Fixing some issues with dynamically supporting multiple SSO providers

kevgliss:
@rpicard I had to change a few things around (mostly JS) as this was a bit more difficult to pull off that I thought. If you get a chance I would interested to see how well this would work with a google config.

### DIFF
--- a/lemur/static/app/angular/app.js
+++ b/lemur/static/app/angular/app.js
@@ -20,7 +20,6 @@
       'ngFileSaver'
     ]);
 
-  fetchData().then(bootstrapApplication);
 
   function fetchData() {
     var initInjector = angular.injector(['ng']);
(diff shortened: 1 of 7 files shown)


lemur/static/app/angular/app.js
```diff
@@ -1,139 +1,158 @@
...
+      'angular-clipboard',
+      'ngFileSaver'
+    ]);
+
+
+  function fetchData() {
+    var initInjector = angular.injector(['ng']);
+    var $http = initInjector.get('$http');
+
+    return $http.get('http://localhost:8000/api/1/auth/providers').then(function(response) {
```
intgr: "It worked just fine on the developer's laptop"? :)

kevgliss: As part of the gulp `build` step it strips out `http://localhost:8000` leaving `/api/1/auth/providers` which should work fine when deployed. ;-)

intgr: Oh right you are, thanks. Turns out my problem was that I ran "gulp build package" in a single command and this way it doesn't do it for some reason.

kevgliss: Interesting... it would be nice to fix that if you figure out why that doesn't work.

Issue #224, opened 2016-01-19
npm postinstall uses global bower that might not exist


AlexClineBB:
While trying to dockerize lemur for running in our environment, I ran into an issue that I thought I'd document in case anyone else encounters it.

Following the installation instructions, you start up the python virtual environment and `make develop`. This takes a while in a docker container, but eventually will error out with this:

```
[19:16:07] 'build:inject' errored after 11 ms
[19:16:07] Error: Bower components directory does not exist at bower_components
```

I traced this to the fact that `bower install` fails to run during the setup phase, because package.json has this line:

```
"postinstall": "bower install --allow-root --config.interactive=false",
```

npm will use the global bower after installing the node modules and will fail (very quietly) if one doesn't exist:

```
npm install
npm WARN cannot run in wd Lemur@ bower install --allow-root --config.interactive=false (wd=/usr/local/src/lemur)
```

The way to workaround this error is to run bower install after the failure, using the local bower. I used something like this:

```
cd /usr/local/src/lemur && \
source /usr/local/src/lemur/bin/activate && \
make develop || \
node_modules/bower/bin/bower install --allow-root --config.interactive=false && \
make develop
```

If package.json is changed so the postinstall command uses the bower installation within the node_modules directory, this workaround won't be needed. Not sure if this should truly be a bug or if it's just a documentation clarification - I'll let the maintainers decide.

kevgliss:
Thanks for the info, I don't know a great deal how NPM but is there a way to specify local npm during runtime? You also have the option of pre-compiling the javascript. The installation script is smart enough to check for the static/dist folder and not try to re-compile it if it exists. This is actually how we deploy it, jenkins builds a deb with compiled javascript inside.

AlexClineBB:
Probably the cleanest way to use the local bower it to change the postinstall line to this:

```
"postinstall": "node_modules/.bin/bower install --allow-root --config.interactive=false",
```

As a thought exercise, do you have an example of how to pre-compile the javascript that I could use before `make develop`?

kevgliss:
Mind opening up a PR with that change? It looks good to me. 

This is essentially what I do in jenkins:
```
npm install
node_modules/.bin/gulp --cwd $WORKSPACE/root/apps/lemur build
node_modules/.bin/gulp --cwd $WORKSPACE/root/apps/lemur package
```

kevgliss:
Closing based on #227.

PR #227, opened 2016-01-20
Use the local bower instead of the global one.

AlexCline:
This change updates package.json's postinstall command to use the locally installed bower, rather than the global bower which might not exist or might not be in the current user's PATH. This addresses issue #224.


kevgliss:
Thanks!

Issue #221, opened 2016-01-15
Remove old key migration scripts


kevgliss:

mik373:
Can't you make the scripts idempotent by checking the table existence in the migration script itself?

kevgliss:
Possibly, but I think it still might be cleaner to just remove them, I don't know of any users that are running that old of a version. Removing old migration scripts is something that should probably be done periodically anyway.

PR #236, opened 2016-01-29
Removing per 2.0 migration scripts

kevgliss:
This should fix a number of issues people were seeing where the correct encryption tables were not created. 

It also adds some documentation to make Lemur initialization more straightforward.


kevgliss:
Closes #221

Issue #218, opened 2016-01-14
Key store password regex not compiling


kevgliss:
The password field on certificate export is failing to validate.

PR #237, opened 2016-01-29
relaxing keystore password validation

kevgliss:
It seems that no matter what I try I can't get a valid regex through to ng-pattern. Removing the requirement for the time being so that users can set their own passwords.


kevgliss:
Closes #218

PR #305, opened 2016-05-07
Slack plugin for notifications

harmw:
Uses the _simple_ incoming-webhooks Slack API to let Lemur send notifications to a Slack channel.


kevgliss:
This looks cool, thanks for the contribution.

PR #306, opened 2016-05-08
Support for pushing PEM certificates to S3

harmw:
**!! WIP !!** (although it does work fine)

This will add a new destination option: S3 buckets. The resulting file will be
stored using server-side encryption.

AWS access is done the usual way, and should that fail it's now also possible
to use plain old environment vars.

@kevgliss what are your thoughts on this one :smile:

### DIFF
--- a/lemur/plugins/lemur_aws/sts.py
+++ b/lemur/plugins/lemur_aws/sts.py
@@ -7,15 +7,29 @@
 """
 import boto
 import boto.ec2.elb
-
+import os
 from flask import current_app
 
 
-def assume_service(account_number, service, region=None):
-    conn = boto.connect_sts()
+class Credentials(object):
+    access_key = os.environ.get('AWS_ACCESS_KEY_ID')
+    secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
+    session_token = None
 
-    role = conn.assume_role('arn:aws:iam::{0}:role/{1}'.format(
-        account_number, current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')), 'blah')
+
+class Role(object):
+    credentials = Credentials()
+
+
+def assume_service(account_number, service, region=None):
+    try:
+        conn = boto.connect_sts()
+        role = conn.assume_role('arn:aws:iam::{0}:role/{1}'.format(
+            account_number, current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')), 'blah')
+    except:
+        current_app.logger.warn('Could not connect to STS, reading credentials from environment')
+        role = Role()
+        pass
 
     if service in 'iam':
         return boto.connect_iam(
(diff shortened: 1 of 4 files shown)


lemur/plugins/lemur_s3/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+            'required': False,
+            'validation': '/^\w+-\w+-\d+$/',
+            'helpMessage': 'Availability zone to use',
+        }
+    ]
+
+    """
+    Outputs the certificate data to a local file and returns its name.
+    """
+    def createPemFile(self, cert, key, chain):
```
kevgliss: Function names are usually under_score, not camelCase.

lemur/plugins/lemur_aws/sts.py
```diff
@@ -7,15 +7,29 @@
...
 import boto.ec2.elb
-
+import os
 from flask import current_app
 
 
-def assume_service(account_number, service, region=None):
-    conn = boto.connect_sts()
+class Credentials(object):
+    access_key = os.environ.get('AWS_ACCESS_KEY_ID')
```
kevgliss: Boto takes care credential access for you. You shouldn't have to specify these key.

lemur/plugins/lemur_aws/sts.py
```diff
@@ -7,15 +7,29 @@
...
+
 
-    role = conn.assume_role('arn:aws:iam::{0}:role/{1}'.format(
-        account_number, current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')), 'blah')
+class Role(object):
+    credentials = Credentials()
+
+
+def assume_service(account_number, service, region=None):
+    try:
```
kevgliss: I wonder if it doesn't make sense to keep the assume role architecture intact here, you would just add s3 permissions to the 'Lemur' role. This way you could also have the ability to write to s3 buckets in other accounts for free.

lemur/plugins/lemur_s3/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+                except S3CreateError as err:
+                    current_app.logger.error("Failed to create bucket %s in region %s: %s" % (bucket_name, find_value('region', options), err))
+                raise
+
+            current_app.logger.debug("Connected to S3 bucket %s in region %s" % (bucket_name, find_value('region', options)))
+
+            k = Key(bucket)
+            k.key = name + '.pem'
+            try:
+                k.set_contents_from_filename(pemfile, replace=True, encrypt_key=True)
```
kevgliss: You could use set_contents from string and avoid writing the file to disc. Alternatively the built in temp utility might accomplish what you are looking for: https://github.com/Netflix/lemur/blob/master/lemur/utils.py#L19

kevgliss:
See: #371

harmw:
awesome 😃 ! I sadly didn't have time to continue this.

PR #354, opened 2016-06-13
Using UTC time in JWT token creation

jroimartin:
As stated in PyJWT's documentation [1] and JWT specification [2][3], UTC
times must be used. This commit fixes JWT decoding in servers not using
UTC time.

[1] https://pypi.python.org/pypi/PyJWT/1.4.0
[2] https://tools.ietf.org/html/rfc7519#section-4.1.6
[3] https://tools.ietf.org/html/rfc7519#section-2


kevgliss:
Servers not using UTC? Madness! Good catch thanks!

jroimartin:
Indeed! In my case, it was my development machine (CEST time), so it may be helpful for developers :)

PR #359, opened 2016-06-14
Openssl authority

mik373:
OpenSSL authority allows to embed a private CA as a Lemur Authority plugin.

OpenSSL issuer and implemented some of the certificate extensions


kevgliss:
Rebase please.

kevgliss:
I am going to close this PR. I believe ( #378 ) fills your basic requirements with a few improvements:

- Instead of writing keys to disc, we store the authorities private key within Lemur itself
- Removed a subprocess call to openssl by using cryptography

I would be very interested to know if that PR is useful enough to switch, I *believe* it should be a drop in replacement.

mik373:
What about some of extensions we've implemented in https://github.com/Netflix/lemur/blob/master/lemur/certificates/service.py, can you please take that as well?  Some of those are critical for us.

kevgliss:
Ah, I missed that change. We should be able to add those in. Can you open a separate PR for the extra extensions?

PR #360, opened 2016-06-14
Letsencrypt aws route53 authority

mik373:
Let's Encrypt Lemur authority plugin for Route53 controlled domains.
The plugin is using ACME client and Route53 Boto API in order to create and answer DNS challenge in order for the certificate to be issued.


kevgliss:
Rebase please.

kevgliss:
Duplicate of #178.

mik373:
Ok, so the plugin does use ACME client but it's specific for AWS since it's using boto route53 client to answer the DNS challenge.  Is that really the same thing?

kevgliss:
We could potentially additional methods of answering challenges, I would expect that would be extended by adding an option to the ACME plugin allowing you to specify (r53, ultraDNS, etc).

The idea in the more generic name is such that we should be able to leverage the plugin to talk to all ACME based CAs not just letsencrypt. For instance if someone wished to leverage [Boulder](https://github.com/letsencrypt/boulder) for their own internal CA.

PR #419, opened 2016-09-02
AWS: allow AssumeRole or already assumed Role

donaldguy:
I am running Lemur in a docker container. 
I had set up https://github.com/swipely/iam-docker already to have the
whole container assume the appropriate role.

I didn't realize until later that lemur itself wanted to trigger
the role change.

This patch makes that optional for people like me, while also addressing
an issue for my next PR, namely the lack of an account_number value along
the ACME route53 path

With this refactor it would be an easy patch to also accept an expected
instance-profile

bumps version of boto3 to 1.4 ; 
though the changes that needed that ( for `boto3.session.Session#get_credentials()`)
could be factored out

### DIFF
--- a/lemur/plugins/lemur_aws/sts.py
+++ b/lemur/plugins/lemur_aws/sts.py
@@ -47,17 +47,57 @@ def assume_service(account_number, service, region='us-east-1'):
             security_token=role.credentials.session_token)
 
 
+def construct_role_arn(role_type='role', **kwargs):
+    account_number = kwargs.get('account_number') or \
+        current_app.config.get('AWS_DEFAULT_ACCOUNT_NUMBER') or \
+        boto3.client('sts').get_caller_identity().get('Account')
+    arn_data = {
+        'num': account_number,
+        'role_type': role_type,
+        'name': current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')
+    }
+    arn_data['service'] = 'sts' if role_type == 'assumed-role' else 'iam'
+
+    return 'arn:aws:{service}::{num}:{role_type}/{name}'.format(**arn_data)
+
+
+def already_assumed_target_role(**kwargs):
+    target_arn = construct_role_arn('assumed-role', **kwargs)
+    current_arn = boto3.client('sts').get_caller_identity().get('Arn')
+    return current_arn.startswith(target_arn)
+
+
+def _current_session_region():
+    _s = boto3._get_default_session()
+    return _s.region_name
+
+
+def _current_session_credentials_dict():
+    _c = boto3._get_default_session().get_credentials()
+    return {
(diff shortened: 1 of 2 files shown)


donaldguy:
moto doesn't implement any mock for `get_caller_identity` - its not a hard endpoint to fake though, at least for simple cases. I don't know anything about the mocking library situation in python and for some reason I've been staring at this project for literally 13 unbroken hours now ... so I'm gonna go to bed I guess 

cheers 😴

lemur/plugins/lemur_aws/sts.py
```diff
@@ -47,17 +47,57 @@ def assume_service(account_number, service, region='us-east-1'):
...
+        'num': account_number,
+        'role_type': role_type,
+        'name': current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')
+    }
+    arn_data['service'] = 'sts' if role_type == 'assumed-role' else 'iam'
+
+    return 'arn:aws:{service}::{num}:{role_type}/{name}'.format(**arn_data)
+
+
+def already_assumed_target_role(**kwargs):
```
donaldguy: Another option to make the test happy, and like could be correct depending on the situation, would just be
```python
if 'account_number' in kwargs:
    return False
```

lemur/plugins/lemur_aws/sts.py
```diff
@@ -47,17 +47,57 @@ def assume_service(account_number, service, region='us-east-1'):
...
-                current_app.config.get('LEMUR_INSTANCE_PROFILE', 'Lemur')
-            )
-            # TODO add user specific information to RoleSessionName
-            role = sts.assume_role(RoleArn=arn, RoleSessionName='lemur')
+
+            region_name = kwargs.get('region') or _current_session_region()
+            role = None
+
+            if already_assumed_target_role(**kwargs):
+                role = {'Credentials': _current_session_credentials_dict()}
```
donaldguy: Technically I think we could just pass through `None`s and have the same effect, but this is more explicit and leaves room for more selective logic

kevgliss:
This is interesting, your deployment looks something like: ECS (InstanceProfile w/assume) | IAM-docker (LemurRole) | Lemur-docker ?

How does that architecture work in a multi-region, multi-account world? Would the LemurRole still then assume role to other accounts?

I don't have anything against making assume-role more flexible (or optional). I will test this out and let you know.

lemur/plugins/lemur_aws/sts.py
kevgliss: I think the default behavior should remain, and only change if an `AWS_DEFAULT_ACCOUNT_NUMBER` is found. This way existing tests should pass and we can add new ones to test the presence of the modified environment.

PR #420, opened 2016-09-02
Attempts to fix ACME plugin

donaldguy:
The acme plugin in master does not work. This PR attempts
to fix it. After several hours I have not succeeded in getting
a succesful letsencrypt cert but I got to the point where I
had no more exceptions and just got
`[2016-09-02 07:34:25 +0000] [19] [CRITICAL] WORKER TIMEOUT (pid:30)`

Because I spent so long on it (partially for want of any other recent
python experience) I wanted to get my changes out there

I am curious if @kevgliss has a branch with some of these fixes that
just hasn't made its way into master yet

This PR is a superset of #419 because that change (or a simpler one
to compensate for lack of explicit account_number decleration) is
needed to get even this close to working

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ develop: update-submodules setup-git
 	npm install
 	pip install "setuptools>=0.9.8"
 	# order matters here, base package must install first
+	pip install -e 'git+https://github.com/certbot/certbot#egg=acme-0.9.0.dev0&subdirectory=acme'
 	pip install -e .
 	pip install "file://`pwd`#egg=lemur[dev]"
 	pip install "file://`pwd`#egg=lemur[tests]"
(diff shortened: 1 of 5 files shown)


donaldguy:
It also pulls in the github sourced acme-0.9.0dev0 because the last release on pypi doesn't have code for the DNS01 challenge.

kevgliss:
The error `WORKER TIMEOUT` would imply that Letsencrypt is simply taking too long to verify our domain and the worker thread is timing out. We could experiment with increasing the timeout of the thread. 

Unfortunately Lemur was originally designed to work with synchronous certificate APIs (organizational validation). This means for async APIs (like letsencrypt) that take upwards of minutes or more to validate we run in to timeouts. 

Thinking about it more, if we wanted to support the async use-case correctly we would most likely define a new plugin type `issuer-async` that would then have some pickup logic. This isn't on our near term roadmap as all of our issuers are synchronous in nature.

PR #421, opened 2016-09-02
Python 3 / alpine dockerfile + gliderlabs 12factor tools (well, a Rockerfile, to front load the npm build)

donaldguy:
In pursuit of my own testing and exploration, I found
myself in want of something a little fresher than
https://github.com/Netflix/lemur-docker

This is a proposal for a new Dockerfile, or in particular
https://github.com/grammarly/rocker file based off of alpine linux
and python 3

The goal of this config is to be usable immediately (i.e. run
`make docker`) for dev, but also more or less be usable as is
for production with the right command line flags.

it uses templating from https://github.com/gliderlabs/sigil
as part of https://github.com/progrium/entrykit to fill in environmental
variables to override toy configs. It builds both nginx and the api
into one small-ish image and leaves hooks for running behind a load
balancer to deal with the likes of SSL, etc.


kevgliss:
We aren't currently deploying Lemur with docker so I am not super up-to-date on all the (and quickly changing) best practices.

I'm a bit hesitant to add docker/deployment type metadata to the actual project itself as it then become opinionated in the way it's deployed. I wonder if this PR shouldn't be targeted at the `lemur-docker` repo itself? Is there a benefit to having these changes directly beside the source?

donaldguy:
This approach makes better use of the checksum based layered caching behavior of docker for development, e.g. On subsequent builds It doesn't reinstall the npm modules unless the package.json has changed. And I believe you could limit the whole js rebuild to just if lemur/static/app changes

My unfamiliarity with the python ecosystem made this effect hard to achieve on that side though (I thought I might only need the setup.py, but I apparently need at least the __about__.py's ?), so the brunt of the benefit is missing as currently any changes to the lemur/ subdir triggers a rebuild of all the python dependencies (in my dev I temporarily copied out specific files I was working on and had them COPY back in as a last step)

These things can be improved but more broadly, they cannot be achieved in a Dockerfile which just clones the repo inside the container during build

You could have a Makefile that e.g. cloned a copy of the lemur repo inside a subdir of the docker-lemur repo if it didn't already exist and then use that, though. This could achieve something similar. But then you are losing the "standard approach" of "docker build" or "docker-compose up" you currently have in that repo - obviously I already kind of gave way on that with the rocker usage, but that could be walked back.

donaldguy:
Also this may need further revision - though the issues may be more to do with python3 broadly than the docker set up.

For example: I was testing it against an already provisioned DB using the google auth plugin without auth-related issues, and I confirmed building from scratch you got the login screen; and bwtweeen the two I thought I was golden, but later I tried to actually login with lemur user / password and I hit something like https://github.com/maxcountryman/flask-bcrypt/issues/9

PR #424, opened 2016-09-07
feat(certificates): add support for restricted domains

terinjokes:
Lemur's documentation already mentions LEMUR_RESTRICTED_DOMAINS, a list
of regular expressions matching domains only administrators can issue
certificates for. An option to mark domains as sensitive existed in the
API, however the configuration option was not implemented.

Now both ways of sensitivity are checked in the same place.


kevgliss:
Thanks!

PR #427, opened 2016-09-19
Fix test certificates module hanging issue

chadhendrie:
As part of a new Lemur plugin project, I am planning to add addition unit tests to the lemur/tests/test_certificates.py module. 

When executing the test_certificates.py tests, all tests are executed, but the test process appears to hang and never completes with the results for the tests.

The hanging issue is traced to the two test methods in the module: test_import(logged_in_user) and test_upload(logged_in_user). The issue has to do with the test methods using the logged_in_user fixture from the conftest.py module as their method parameters.

The test methods at issue require the session, db, and app fixtures to be initialized for the tests to complete successfully. The logged_in_user() fixture only initializes the app fixture. Updating the test_import() and test_upload() methods parameters to be the "session" fixture fixes the hanging issue and the tests complete successfully when executed.

Here is the command being used to execute the certificate tests...

$ py.test -s -v lemur/tests/test_certificates.py

### DIFF
--- a/lemur/tests/conftest.py
+++ b/lemur/tests/conftest.py
@@ -141,7 +141,7 @@ def issuer_plugin():
 
 
 @pytest.yield_fixture(scope="function")
-def logged_in_user(app):
+def logged_in_user(session, app):
     with app.test_request_context():
         identity_changed.send(current_app._get_current_object(), identity=Identity(1))
         yield
--- a/lemur/tests/test_certificates.py
+++ b/lemur/tests/test_certificates.py
@@ -340,7 +340,7 @@ def test_create_csr():
     assert private_key
 
 
-def test_import(logged_in_user):
+def test_import(session):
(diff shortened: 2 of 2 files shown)


chadhendrie:
Would it be possible for someone to take a look at this pull request and let me know if you see any issues?

Thanks - Chad

kevgliss:
This looks fine. I wonder though, was there are reason why you switched the fixture to `session` instead of adding `session` to the logged_in_user fixture? Did that not stop the hanging?

chadhendrie:
I had not try to adding the `session` fixture to the `logged_in_user` fixture (I was trying to tread lightly with the change). But, I went back and added `session` to the `logged_in_user`. This change also fixes the hanging issue. I like this change better. If it's ok, I will update the pull request with a new commit.

kevgliss:
Looks good, thanks!

PR #428, opened 2016-09-22
fix(certificates): import re module

terinjokes:
Well, this is embarrassing. 😢  Seems to have lost the actual import at some point.


kevgliss:
Embarrassing for you and me! Sometimes the simplest ones are the easiest to miss.

PR #434, opened 2016-10-06
Fix Python3 import certificate private key encoding

chadhendrie:
When importing a certificate with a private key, it is passed to the import/upload process from the UI as a str object. In Python3, this raises two errors when processing the private key:

1. The private key validation fails
2. The database insert of the certificate fails

The fix in both cases is to correctly encode the private key as a bytes object.


kevgliss:
This looks good, we have slowly being trying to convert our production instance to python3 and there have been a great many of these edge cases, so this helps a lot.

Issue #431, opened 2016-09-29
Import regex should attempt to look or RSA in the header but also in end of the file.


kevgliss:
Currently the regex for importing certificates only looks at the beginning of the string for the "RSA" private key. We should be more specific and look for RSA in the beginning and the end of the text.

PR #435, opened 2016-10-07
Update the private key regex validation

chadhendrie:
We have a need to import Let's Encrypt certificates into Lemur. The private keys provided by the Let's Encrypt certificate authority as part of their certificate bundle fail the import/upload certificate private key validation in the UI. The validation is looking for a specific character sequence at the begin of the private key.

In order to support valid Let's Encrypt private keys, we would like to updated the private key regex validation to check for both the existing character sequence and the Let's Encrypt character sequence.

Examples

Existing: -----BEGIN RSA PRIVATE KEY-----

Let's Encrypt: -----BEGIN PRIVATE KEY-----

### DIFF
--- a/lemur/static/app/angular/certificates/certificate/upload.tpl.html
+++ b/lemur/static/app/angular/certificates/certificate/upload.tpl.html
@@ -60,7 +60,7 @@
 
                 <div class="col-sm-10">
                 <textarea name="privateKey" ng-model="certificate.privateKey" placeholder="PEM encoded string..."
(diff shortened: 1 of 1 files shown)


kevgliss:
This looks fine, we should also try an close #431 with this change is well. If you don't mind adding it that would be awesome otherwise, we can merge it and I will open up a PR the other half of the validation.

chadhendrie:
We will go ahead and update the regex to check both the header and the footer of the private key and add the commit to this pull request.

PR #436, opened 2016-10-07
Log request data on error

mikegrima:
Improves debugging by logging the request data in the event there was an exception.


kevgliss:
I agree this would be nice to have, however I think it should be lower level. Perhaps utilizing `@app.before_request`. Something like:

    @app.before_request
    def log_requests():
        from flask import request
        app.logger.debug('|'.join([
            request.remote_addr,
            request.method,
            request.url,
            str(request.data)
        ]))

We would have to vary of is accidentally logging private keys and passwords. I don't think whitelist is the correct approach here but I don't have any better ideas.

PR #441, opened 2016-10-10
Fix the Java export default password generator

chadhendrie:
When exporting a certificate to a Java Keystore or Java Truststore, the password is an optional parameter. When a password is not supplied by the caller, a default password is generated by the export method. The generation library creates the random password as a bytes object. The bytes object raises an error in the '`keytool`' command used to export the certificate. The `keytool` is expecting the password to be a str object.

The fix is to decode the generated password from a bytes object to a str object.

The associated Java plugin tests have been updated to verify the export method returns the password as a str object. In addition, the tests have been updated to correctly test the export methods response object. The original tests treated the response as a single object. The current export methods return a tuple of data (type, password, data).

In order to make the tests compatible with both Python2 and Python3, the 'six' library was used to test the password is in fact a string.

### DIFF
--- a/lemur/plugins/lemur_java/plugin.py
+++ b/lemur/plugins/lemur_java/plugin.py
@@ -178,7 +178,7 @@ class JavaTruststoreExportPlugin(ExportPlugin):
         if self.get_option('passphrase', options):
             passphrase = self.get_option('passphrase', options)
         else:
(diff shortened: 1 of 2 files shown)


kevgliss:
Lgtm

Issue #450, opened 2016-10-14
Cannot install application


neilschelly:
Updating gulp-protractor dependency
Following the quickstart instructions at http://lemur.readthedocs.io/en/latest/quickstart/index.html, I ran into issues at `make develop` because several dependencies couldn't be resolved. Several errors like this were in the output in the output:

```
npm ERR! TypeError: Cannot read property 'latest' of undefined
npm ERR!     at next (/usr/share/npm/lib/cache.js:687:35)
npm ERR!     at /usr/share/npm/lib/cache.js:675:5
npm ERR!     at saved (/usr/share/npm/node_modules/npm-registry-client/lib/get.js:142:7)
npm ERR!     at /usr/lib/nodejs/graceful-fs/polyfills.js:133:7
npm ERR!     at Object.oncomplete (fs.js:107:15)
npm ERR! If you need help, you may report this log at:
npm ERR!     <http://github.com/isaacs/npm/issues>
npm ERR! or email it to:
npm ERR!     <npm-@googlegroups.com>

npm ERR! System Linux 3.13.0-92-generic 
npm ERR! command "/usr/bin/nodejs" "/usr/bin/npm" "install"
npm ERR! cwd /home/lemur/lemur
npm ERR! node -v v0.10.25
npm ERR! npm -v 1.3.10
npm ERR! type non_object_property_load
```

`npm list` yielded this output at the bottom:
 
```
npm ERR! missing: @types/jasmine@^2.2.31, required by protractor@4.0.9
npm ERR! missing: @types/node@^6.0.35, required by protractor@4.0.9
npm ERR! missing: @types/q@^0.0.30, required by protractor@4.0.9   
npm ERR! missing: @types/selenium-webdriver@~2.53.30, required by protractor@4.0.9
npm ERR! missing: adm-zip@0.4.7, required by protractor@4.0.9
npm ERR! missing: chalk@^1.1.3, required by protractor@4.0.9
npm ERR! missing: glob@^7.0.3, required by protractor@4.0.9
npm ERR! missing: jasmine@2.5.2, required by protractor@4.0.9
npm ERR! missing: jasminewd2@0.0.10, required by protractor@4.0.9
npm ERR! missing: optimist@~0.6.0, required by protractor@4.0.9
npm ERR! missing: q@1.4.1, required by protractor@4.0.9
npm ERR! missing: saucelabs@~1.3.0, required by protractor@4.0.9
[...]

PR #451, opened 2016-10-14
Updating gulp-protractor dependency

neilschelly:
https://github.com/Netflix/lemur/issues/450 describes a problem I ran into trying to install lemur that I tracked down to dependency issues on protractor. This is how I resolved the problem. I am unsure what testing or scrutiny is required to get this accepted, but I'm open to reading or doing more if necessary.


kevgliss:
As this is a minor point release, I don't see a problem in updating this dependency.

PR #452, opened 2016-10-14
Implement a CFSSL issuer plugin

chadhendrie:
As part of our adoption of Lemur as our certificate manager, we are planning to use CFSSL (https://github.com/cloudflare/cfssl) as our internal certificate authority.

The new plugin interacts with CFSSL through the CFSSL REST API. The CFSSL configuration is defined in the lemur.conf.py property file using property names prefixed with "CFSSL_". The CFSSL certificates have been successfully created and tested using both python 2.7 and 3.5.

Please review and let us know if you see any issues with the plugin.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -51,7 +51,7 @@ Basic Configuration
         CORS = False
 
 
-.. data:: SQLACHEMY_DATABASE_URI
+.. data:: SQLALCHEMY_DATABASE_URI
     :noindex:
 
         If you have ever used sqlalchemy before this is the standard connection string used. Lemur uses a postgres database and the connection string would look something like:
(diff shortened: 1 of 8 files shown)


kevgliss:
Please update: https://raw.githubusercontent.com/Netflix/lemur/master/docs/administration.rst with the available options and usages.

chadhendrie:
@kevgliss  I am working on updating the administration.rst documentation and I noticed there is also plugin documentation in the lemur/docs/developer/internals package.

Should I also create a lemur.plugins.lemur_cfssl.rst file and update the lemur.plugins.rst file to include the CFSSL plugin?

Thanks

kevgliss:
That would be good to have... that section is a bit neglected. I believe it was auto created by a documentation script I used to bootstrap the docs as it looks like the majority of plugins aren't there. 

After you create them for cfssl, if you think they are generally useful we should open up an issue and get the rest of the plugins in there.

kevgliss:
Thanks!

PR #460, opened 2016-10-19
Fixing test to take python3 into account.

kevgliss:


kevgliss:
These do pass locally, i'll be damned if I can figure out what travis is doing...

PR #463, opened 2016-10-24
fix python2.x branch of get_rsa_public_key (auth/service.py)

cviecco:
Two issues are present
1. The value returned by the json is utf-8 and base64url_decode requires a string. An explicit conversion is needed. (lines 47 and 48)
2. The base64url_decode function takes a single paramters not two as in python 3 (lines 48)


kevgliss:
Good catch. Thanks!

PR #464, opened 2016-10-24
fix(web): send JSON for all errors

terinjokes:
Configure werkzeug to output JSON error messages for the benefit of
downstream clients. This also allows for metrics collection in all cases
where werkzeug is outputting an exception.


terinjokes:
@kevgliss The CI failures don't seem to be a result of my changes.

kevgliss:
Thanks!

Issue #474, opened 2016-11-02
LEMUR_DEFAULT_ORGANIZATION ignored for creating new authority


neilschelly:
When creating a new Authority out of the box (following the quickstart instructions), I expected the LEMUR_DEFAULT_ORGANIZATION variable to be the default organization for the new CA. The other LEMUR_DEFAULT_* variables work this way, but that field will default to Netflix. The defaults all work as expected through the for creating a Certificate already.

![create-authority](https://cloud.githubusercontent.com/assets/3343289/19912600/fa55196a-a071-11e6-9b48-138d476a1327.png)

PR #475, opened 2016-11-02
Fixing default organization for a new authority.

neilschelly:
This change fixes https://github.com/Netflix/lemur/issues/474 with the default organization in a new authority. Previously, the configuration variables were ignored and "Netflix" was auto-filled.


kevgliss:
thanks!

PR #570, opened 2016-12-07
Certificate rotation enhancements

kevgliss:
- Moving cli to new home under the module that owns it.
- Adding additional logic for conditional certificate rotation.
- Refactoring messaging for certificate rotation and expiration.


lemur/notifications/messaging.py
```diff
@@ -0,0 +1,102 @@
...
+    :license: Apache, see LICENSE for more details.
+
+.. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
+
+"""
+
+import arrow
+from flask import current_app
+
+from lemur import database, metrics
```
jchuong: How does this work?  Everything else uses `from lemur.extensions. import metrics`, otherwise this just a module.

kevgliss: It works because lemur's __init__ imports those two modules. That being said they should be imported from `lemur.extensions` to avoid circular imports.

PR #598, opened 2016-12-14
Plugin base classes: update method signatures & fix raise

intgr:
This way IDEs can verify method overrides in subclasses, otherwise these
are flagged as erroneous.

Changed base classes to properly raise NotImplementedError; previously
they would cause "TypeError: exceptions must derive from BaseException"

Also fixed exception handling in sources.service.clean().

### DIFF
--- a/lemur/sources/service.py
+++ b/lemur/sources/service.py
@@ -198,7 +198,7 @@ def clean(source):
 
     try:
         certificates = s.clean(source.options)
-    except NotImplemented:
+    except NotImplementedError:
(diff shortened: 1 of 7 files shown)


lemur/sources/service.py
```diff
@@ -198,7 +198,7 @@ def clean(source):
 
     try:
         certificates = s.clean(source.options)
-    except NotImplemented:
+    except NotImplementedError:
```
kevgliss: You are correct I don't believe this is a hot code path so I probably missed this.

PR #600, opened 2016-12-14
Fix check for sensitive and restricted domains

intgr:
This is a fix for a potential security issue. Fix submitted as a pull request per discussion with @kevgliss.

There was much wrong with this code:
* If there were no matches in the Domain table then
  LEMUR_RESTRICTED_DOMAINS was ignored.
* If there were multiple Domain matches, the LEMUR_RESTRICTED_DOMAINS
  checks were ran multiple times.
* It's unnecessary to query Domain rows if the user is an admin.
* If a domain failed the check, the error would say
  "Domain Domain(name=example.org) has been marked as sensitive ..."

### DIFF
--- a/lemur/common/validators.py
+++ b/lemur/common/validators.py
@@ -42,20 +42,20 @@ def private_key(key):
 
 def sensitive_domain(domain):
     """
-    Determines if domain has been marked as sensitive.
+    Determines if domain has been marked as sensitive or matches restricted patterns.
     :param domain:
     :return:
     """
+    if SensitiveDomainPermission().can():
+        # User has permission, no need to check anything
+        return
+
     restricted_domains = current_app.config.get('LEMUR_RESTRICTED_DOMAINS', [])
-    if restricted_domains:
-        domains = domain_service.get_by_name(domain)
-        for domain in domains:
-            # we only care about non-admins
-            if not SensitiveDomainPermission().can():
-                if domain.sensitive or any([re.match(pattern, domain.name) for pattern in restricted_domains]):
-                    raise ValidationError(
-                        'Domain {0} has been marked as sensitive, contact and administrator \
-                        to issue the certificate.'.format(domain))
+    if any(re.match(pattern, domain) for pattern in restricted_domains) or \
+            any(d.sensitive for d in domain_service.get_by_name(domain)):
+        raise ValidationError(
+            'Domain {0} has been marked as sensitive or restricted, contact and administrator '
+            'to issue the certificate.'.format(domain))
 
 
 def encoding(oid_encoding):


lemur/common/validators.py
```diff
@@ -42,20 +42,20 @@ def private_key(key):
...
-        domains = domain_service.get_by_name(domain)
-        for domain in domains:
-            # we only care about non-admins
-            if not SensitiveDomainPermission().can():
-                if domain.sensitive or any([re.match(pattern, domain.name) for pattern in restricted_domains]):
-                    raise ValidationError(
-                        'Domain {0} has been marked as sensitive, contact and administrator \
-                        to issue the certificate.'.format(domain))
+    if any(re.match(pattern, domain) for pattern in restricted_domains) or \
+            any(d.sensitive for d in domain_service.get_by_name(domain)):
```
intgr: This could be rewritten to filter Domain.sensitive on database-side. But I guess the old code was written in this style to keep the generality of `get_by_name()`.

lemur/common/validators.py
```diff
@@ -42,20 +42,20 @@ def private_key(key):
...
-    Determines if domain has been marked as sensitive.
+    Determines if domain has been marked as sensitive or matches restricted patterns.
     :param domain:
     :return:
     """
+    if SensitiveDomainPermission().can():
+        # User has permission, no need to check anything
+        return
+
     restricted_domains = current_app.config.get('LEMUR_RESTRICTED_DOMAINS', [])
```
kevgliss: For us, we generally allow all domains (as validated by issuers) and only blacklist them very specifically. I think at some point we should do away with this setting altogether and rely on the `sensitive` attribute on the domain model. The domains UI in particular has been neglected so it might take a bit of work to get it going.

lemur/common/validators.py
```diff
@@ -42,20 +42,20 @@ def private_key(key):
 
 def sensitive_domain(domain):
     """
-    Determines if domain has been marked as sensitive.
+    Determines if domain has been marked as sensitive or matches restricted patterns.
     :param domain:
     :return:
     """
+    if SensitiveDomainPermission().can():
+        # User has permission, no need to check anything
```
kevgliss: We used to have something like this:

https://github.com/Netflix/lemur/blob/0.4.0/lemur/tests/conftest.py#L183

I had removed that fixture when I removed most of the permission logic from the services and into the views. This may be an edge case where this particular fixture is useful again.

lemur/common/validators.py
kevgliss: It depends, we can either keep the regex matching or rely entirely on the domain.sensitive field for this logic. I would prefer the latter going forward as I believe it would be more flexible.

kevgliss:
If we choose to not only fix this functionality, but improve it. I think adding a event log similar to: https://github.com/Netflix/lemur/blob/master/lemur/certificates/views.py#L449
when a request is rejected might make sense for an auditing perspective.

kevgliss:
@intgr Any update on this PR?

intgr:
@kevgliss Sorry, I've been busy with other things. Hopefully next week I can return to working on Lemur.

kevgliss:
Closing this now for now. Feel free to re-open.

PR #607, opened 2016-12-15
Fall back to CN for CA name when organization is not available

intgr:
In-house CAs may not have the organization field filled out.

### DIFF
--- a/lemur/common/defaults.py
+++ b/lemur/common/defaults.py
@@ -186,19 +186,23 @@ def bitstrength(cert):
 
 def issuer(cert):
     """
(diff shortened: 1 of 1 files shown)


kevgliss:
Organization *should* be set either via the `LEMUR_DEFAULT_ORGANIZATION` or via the UI. Is the use case that: it is not set, it cannot be set or it should be set dynamically?

intgr:
The use case is importing legacy CA certificates into Lemur. I can't change them.

kevgliss:
Makes sense. Cheers.

PR #643, opened 2017-01-08
Adjusting the way that certificates are requested.

kevgliss:
Attempts to resolve issues found by #605


harmw:
I see you dropped the _ACME_PRIVATE_KEY_ in favor of generating a new one per cert/request - sounds ok. It's all still pretty much tied to AWS Route53 though... (which I personally don't like considering _my_ plans on getting a different DNS provider in place)

Not sure if I can test this but I'll give it a try. Many thanks!

kevgliss:
I wouldn't be against breaking out the DNS registration logic from r53... It's mostly segmented as is.  Route53 is simply what I have access to.

Issue #650, opened 2017-01-18
Authority Key Identifier schema inconsistency


neilschelly:
The Authority Key Identifier extension has two options presented in the UI and the schema on both certificates and authorities. Both are options in the Authority Key Identifier extension.

For authorities in the UI:
authority.extensions.authorityKeyIdentifier.useKeyIdentifier
authority.extensions.authorityKeyIdentifier.useAuthorityCert

For certificates in the UI:
certificate.extensions.authorityKeyIdentifier.useKeyIdentifier
certificate.extensions.authorityIdentifier.useAuthorityCert

The certificates representation above doesn't make sense, as the useAuthorityCert option isn't part of the authorityKeyIdentifier extension options. The authorities representation matches the usage better.

The schemas also recognize these as two different attributes in the ExtensionSchema, each with separate nested schemas. That causes the authorities values to get mangled/dropped. Both options should be handled in the same nested schema called AuthorityKeyIdentifierSchema.

This would have the effect of making the overlapping authority and certificate schemas consistent with each other, and align the Authority Key Identifier options like the other extensions.

PR #651, opened 2017-01-18
Combining Authority Key Identifier extension options in the schema.

neilschelly:
This makes processing them in the cert/csr generation stage make more sense because they are two options in the same x.509 extension. They were already in the same part of the schema for authorities, but this makes the certificates follow the same pattern, and it allows them to share the same schema/validation layout.

Please see issue #650 for more detail.

### DIFF
--- a/lemur/schemas.py
+++ b/lemur/schemas.py
@@ -170,12 +170,9 @@ class BasicConstraintsSchema(BaseExtensionSchema):
     pass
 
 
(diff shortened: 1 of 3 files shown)


neilschelly:
I removed a test because I removed the thing that it was testing, hence the decrease in coverage percentage.

kevgliss:
lgtm thanks!

Issue #652, opened 2017-01-18
UI Handling of Custom OIDs


neilschelly:
There's a lot of unexpected behavior with these, both in certificates and authorities:
* The is_critical/isCritical value is not in the marshmallow schemas, so whatever is submitted for it is dropped.
* In the angular JS code, if isCritical isn't submitted (unchecked), then it is not being interpreted as false.
* As custom OIDs are added in the UI, they are displayed in a table that is pretty self-explanatory without a table header row, except for the critical value. I recommend "critical" or "" instead of True/False.
* The checkbox for critical has got the wrong namespace. It's certificate.extensions.custom.isCritical (or authority.extensions.custom.isCritical), when the other custom OID names (and the underlying JS code) suggest it should be just certificate.customIsCritical (or authority.customIsCritical).

PR #653, opened 2017-01-18
Fixing handling of adding custom OIDs in UI

neilschelly:
* is_critical wasn't in the schema, so was getting dropped.
* isCritical in the Javascript wasn't getting assigned if it was unchecked. Now, it will be assumed false if missing.
* The display of critical or not in the list of added custom OIDs was unclear when it was just true/false with no heading. Now it will be displayed as critical or nothing instead.
* The namespace for the checkbox for isCritical was wrong, and didn't get processed with the oid/type/value variables.

See Issue #652 for details.


kevgliss:
lgtm thanks!

Issue #654, opened 2017-01-18
Almost all extensions are wiped out when creating an authority


neilschelly:
See this conditional statment:
https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/authorities/services.js#L17

Basically, if there are no subjectAlternativeNames submitted from the JS form, then the extensions object will be initialized to a mostly empty extensions structure with no SANs. Any other checked boxes for various extensions will be wiped out unless a SAN was setup.

The same function appears in the certificates workflow, except it works. It will only create an empty extensions structure if extensions is not already defined.
https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/certificates/services.js#L21

PR #655, opened 2017-01-18
Fixing extension handling when creating authorities

neilschelly:
I'm stealing this code from the attachSubAltName function in the certificates JS workflow. Prior to this, the authorities version was wiping out any extensions if their weren't SAN names from the authority UI. See Issue #654 for more details.


kevgliss:
lgtm thanks!

Issue #656, opened 2017-01-18
Pedantic typo in error message


neilschelly:
I was confused more than a few times about this error when I ran into it:
https://github.com/Netflix/lemur/blob/master/lemur/authorities/schemas.py#L63

It's missing a comma between "subca" and "parent" to make that read more clearly.

PR #657, opened 2017-01-18
Correcting grammar for subca ValidationError message for clarity

neilschelly:
See issue #656 for explanation. I got stuck on this error message a bit before digging into the code deeper here.


kevgliss:
lgtm

Issue #663, opened 2017-01-19
KeyAgreement, Encipher Only, and Decipher Only are related


neilschelly:
If either _Encipher Only_ or _Decipher Only_ are selected for KeyUsage in the UI, then _Key Agreement_ should also be checked. Also, _Encipher Only_ and _Decipher Only_ are mutually exclusive.

Further, the backend schemas and code should ensure this is enforced as well for API requests, as per the comment from @kevgliss found here: https://github.com/neilschelly/lemur/commit/aef68193426d2b3828c3a7d3310ad949707fa298

neilschelly:
I'm unsure how to make a schema validator that makes decisions based on the values of more than one input. I need to read up more on marshmallow's validator/schema docs to better understand more advanced techniques.

neilschelly:
Two PRs are ready. I like the second one using radio buttons:
1. https://github.com/Netflix/lemur/pull/664 (original checkbox version)
2. https://github.com/Netflix/lemur/pull/692 (newer radio button version)

PR #664, opened 2017-01-19
UI adjustments for mutually exclusive (checkbox version) encipher/decipher-only Key Usage

neilschelly:
UI adjustments to make Key Agreement, Encipher Only, and Decipher Only relationship more user-friendly. This contributes the UI half of issue #663.

### DIFF
--- a/lemur/common/fields.py
+++ b/lemur/common/fields.py
@@ -162,6 +162,9 @@ class KeyUsageExtension(Field):
             elif k == 'useCRLSign':
                 keyusages['crl_sign'] = v
 
+            elif k == 'useKeyAgreement':
+                keyusages['key_agreement'] = v
+
             elif k == 'useEncipherOnly' and v:
                 keyusages['encipher_only'] = True
                 keyusages['key_agreement'] = True
--- a/lemur/static/app/angular/authorities/authority/extensions.tpl.html
+++ b/lemur/static/app/angular/authorities/authority/extensions.tpl.html
@@ -67,17 +67,17 @@
       </div>
       <div class="checkbox">
         <label>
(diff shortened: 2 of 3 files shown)


kevgliss:
If we want to allow for mutual exclusivity wouldn't a radio button be better here instead of forcing checkboxes to be disabled?

neilschelly:
I could take a stab at that.

neilschelly:
The backend validation/schema stuff for this actually got wrapped into the work at https://github.com/Netflix/lemur/pull/666 just because the entirety of handling input got refactored there.

kevgliss:
The only work left on this would be if we decided to use radio buttons instead of checkboxes... did you want me to merge this in for you to tackle that (if you want) with another PR?

neilschelly:
I played with radio buttons a bit. It's complicated in a variable naming context since those wouldn't "look" like all the other extension checkboxes in the data that comes back to the backend. That could be massaged from the radio buttons in Angular JS so that it still reaches the backend "looking" like it came from checkboxes to keep that consistent? Maybe consistent isn't that important, but if not, that's changing the API call inputs too.

I also had trouble with unchecking the radio buttons. A certificate can have neither of these options checked, but a radio button implies at least one of them will be checked. I'm sure it could be done by someone more familiar with Angular and/or JS, but I was unable to do it after an hour of playing. I opted to focus on the functional code contributions first, and I planned to revisit this later.

neilschelly:
I propose rejecting this PR in favor of https://github.com/Netflix/lemur/pull/692. That version accomplishes the same thing with radio buttons.

kevgliss:
Agreed, radio buttons do feel a lot nicer. Thanks!

Issue #646, opened 2017-01-09
Support more KeyUsage and ExtendedKeyUsage options


neilschelly:
The UI includes checkboxes to set all these options, but they mostly get ignored now. Many of the x509 extensions are pseudo-coded in a comment beginning here: https://github.com/Netflix/lemur/blob/master/lemur/certificates/service.py#L365

* ~~Some options will be mutually exclusive, like KeyUsage encipherOnly and decipherOnly.~~ separated to #663 
* ~~Some options will imply others, like encipherOnly and decipherOnly requiring keyAgreement.~~ separated to #663
* keyCertSign will imply the BasicConstraint for CA:TRUE. 
* ~~Some extensions requested in the CSR state can only be granted or completed in the signing phase within the CA plugin, like inserting CRL and/or OCSP service URIs.~~ Out of scope. Plugins/providers will ignore what they want to and add what they need to when signing CSRs.
* ~~Since Authorities are essentially a Certificate with specific extensions and usages, some of these options may also need to be applied to the Authority object and data structures.~~ Implemented in #658 

EDIT: I'm reducing the scope of this. I want this ticket to be about encoding the checkboxes/values from the authority/certificate creation pages into the resulting CSR (that gets submitted to the authority for signing). This is primarily all the KeyUsage and ExtendedKeyUsage stuff.

Out of scope:
* Custom OIDs: These are troublesome because they are inherently not supported by the underlying python-cryptography library. It's probably not too much effort to support them in a non-critical state, but the library explicitly _can't_ support custom OIDs with the critical flag, because that would violate the RFC. Issue #665 can work on this.
* CRL Distribution Points: The UI currently offers 3 options (Yes/No/Default), but those values get dropped. It's hard to be clear on what _should_ be here, but the end goal is to include the description/URLs/etc that would be included in the certificate. At the very least, it's unlikely the CSR creation will know these values and the authority plugin/provider would be necessary to do handle this. Issue #662 was created to better deal with this.

kevgliss:
This sounds good, all those checkbox options should already be passed to `issuer_options` into any given plugin.  You're right that a lot of those options are for the CA might not even be respected if inserted into the CSR. Looking forward to it.

Issue #665, opened 2017-01-19
Custom OIDs not encoded into CSR


neilschelly:
These are troublesome because they are inherently not supported by the underlying python-cryptography library. It's probably not too much effort to support them in a non-critical state, but the library explicitly can't support custom OIDs with the critical flag, because that would violate the RFC for x509. An x509 implementation (python cryptography here) has to raise an error upon handling a critical extension it doesn't natively recognize/support. So, the x509.UnrecognizedExtension type inherently cannot support critical extensions.

In the create_csr() function, a custom extension will look like:
```{'oid': u'1.3.6.1.5.5.7.3.25', 'value': u'blah', 'is_critical': True, 'encoding': u'string'}```
The value should be DER encoded (unsure about how just now) and added to the CSR in create_csr like:
```
builder = builder.add_extension(
    x509.UnrecognizedExtension(
        oid=custom['oid'],
        value=custom['value']
    )
)
```

kevgliss:
It seems to me that we should respect the RFC here. If custom OIDs should not be allowed to be critical we shouldn't support setting it as such.

PR #666, opened 2017-01-19
X509 extensions issue#646

neilschelly:
This implements Issued #646 by processing all the extensions in the create_csr function to ensure those checkboxes get encoded to the CSR before passing it along to the CA for signing.

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -9,6 +9,7 @@ import arrow
 
 from flask import current_app
 
+from cryptography import x509
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 from sqlalchemy.orm import relationship
@@ -229,16 +230,39 @@ class Certificate(db.Model):
 
     @property
     def extensions(self):
-        # TODO pull the OU, O, CN, etc + other extensions.
-        names = [{'name_type': 'DNSName', 'value': x.name} for x in self.domains]
+        return_extensions = {}
+        cert = lemur.common.utils.parse_certificate(self.body)
+        for extension in cert.extensions:
+            if isinstance(extension, x509.BasicConstraints):
+                return_extensions['basic_constraints'] = extension
+            elif isinstance(extension, x509.SubjectAlternativeName):
+                return_extensions['sub_alt_names'] = extension
+            elif isinstance(extension, x509.ExtendedKeyUsage):
+                return_extensions['extended_key_usage'] = extension
+            elif isinstance(extension, x509.KeyUsage):
+                return_extensions['key_usage'] = extension
+            elif isinstance(extension, x509.SubjectKeyIdentifier):
+                return_extensions['subject_key_identifier'] = {'include_ski': True}
+            elif isinstance(extension, x509.AuthorityInformationAccess):
+                return_extensions['certificate_info_access'] = {'include_aia': True}
+            elif isinstance(extension, x509.AuthorityKeyIdentifier):
+                aki = {
+                    'use_key_identifier': False,
+                    'use_authority_cert': False
+                }
+                if extension.key_identifier:
+                    aki['use_key_identifier'] = True
+                if extension.authority_cert_issuer:
+                    aki['use_authority_cert'] = True
+                return_extensions['authority_key_identifier'] = aki
+            elif isinstance(extension, x509.CRLDistributionPoints):
+                # FIXME: Don't support CRLDistributionPoints yet https://github.com/Netflix/lemur/issues/662
+                pass
+            else:
+                # FIXME: Not supporting custom OIDs yet. https://github.com/Netflix/lemur/issues/665
+                pass
 
-        extensions = {
-            'sub_alt_names': {
-                'names': names
-            }
-        }
-
-        return extensions
+        return return_extensions
 
     def get_arn(self, account_number):
         """
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -345,109 +345,23 @@ def create_csr(**csr_config):
         x509.NameAttribute(x509.OID_EMAIL_ADDRESS, csr_config['owner'])
     ]))
 
-    # Assume that we're not creating a CA certificate, but allow setting
-    # certificate_authority in csr_config to True to switch this.
-    if csr_config.get('certificate_authority', False):
-        builder = builder.add_extension(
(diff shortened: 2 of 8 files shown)


kevgliss:
Have you thought about using marshalling instead of giant if statements? We could potentially use the existing Extension schema to correctly marshall into keys into the proper oid objects. The only drawback I can think of is this would then put the onus on the issuer plugin to correctly convert these values if they do not respect the CSR (via issuer_options). But that is already generally done with existing issuer-plugins.

kevgliss:
For instance, we could define custom fields fields at https://github.com/Netflix/lemur/blob/master/lemur/schemas.py#L198 that way underlying code could assume they have proper oid object instead of recreating them for all of the different flows.

neilschelly:
It did not occur to me to have marshmallow handle this. Would it be possible to do it there without breaking the input of options to everywhere further down the line? Presumably, some plugins may already depend on issuer_options looking like they do now, right?

kevgliss:
It doesn't look like it would affect any public plugins, they are mostly naive and do no look beyond SubAltNames. Our internal plugin would be affected but I could modify that relatively easily if we thought it would make the code cleaner.

neilschelly:
I'll have to read up on the schemas/validators/etc a lot more next week to see about doing this. I know I already struggled to find a backend mechanism to validate the mutual exclusion for the encipher/decipher options. I mostly just have to sit down and learn how it works better I think.

Would you expect that this should organize all the extensions _except_ the SubAltNames then? Or would I wrap that in too and expect all the plugins to catch up? Or make sure that both the structured Extensions co-exist with the current issuer_options keys?

kevgliss:
Yes, I think it should be progressive. Starting with the functionality you are adding and going from there... I wouldn't expect you to do it for all extensions right away. 

Custom fields should be straightforward. https://github.com/Netflix/lemur/blob/master/lemur/common/fields.py#L8 Is a needless complex one for marshalling iso date strings into arrow objects. Essentially some places I would use arrow and others datetime so I just ensured that all schemas where creating Arrow objects.

neilschelly:
I documented some of the concerns that make custom fields a bit more complicated in Issue #665. I don't know any way around the critical bit.

neilschelly:
The processing of some extensions can be moved into the schema validation step, like KeyUsage, ExtendedKeyUsage, SubjectAlternativeNames, Custom OIDs. The data necessary to construct these extensions is all submitted with the form. Some of these cannot be constructed in the schema because the data isn't available until the CSR phase. SubjectKeyIdentifier is an example of that. That can only be calculated in or after the CSR phase. Other extensions can't happen until the CA signs the certificate. This includes AuthorityKeyIdentifier, CertificateInfoAccess, and CRLDistributionPoints.

If we go with the mechanism in this PR already, these extensions are handled/processed in the CSR phase as much as possible. The only exception is those which must happen in the CA signing phase, and that will usually happen outside this codebase anyway.

If we go with your suggestion, some extensions will be handled in the data schema, some will be handled in the CSR phase, and some will be handled in the CA phase. Assuming the CA is still mostly outside this codebase, it means that the processing of x509 extensions into x509 ExtensionType objects is necessarily split between two primary places.

Personally, I have a preference toward leaving it the way it is in this PR as the cleaner option. I'm still trying to wrap my head around the marshmallow custom fields stuff to see if I can make a proof of concept of that though.

neilschelly:
I have a working mockup (needs cleaning) of putting schema custom types in charge of KeyUsage and ExtendedKeyUsage here: https://github.com/neilschelly/lemur/compare/bbb843cb53c00bf4377248d56fef7147cb8d9a1d...test

kevgliss:
I agree that having validation in multiple places isn't great. I wonder if it's not possible to use marshmallow fields to marshall and validate data in the CSR phase itself. 

I'm thinking that we still use marshmallow for serialization/validation but don't actually attempt to validate until we have all the data. 

Traditionally I have done validation in the views, but for this particular use case I think we could define extensions as a `fields.Dict()` in the view and then only attempt to validate/marshall the data inside the `create_csr` function using `ExtensionSchema().load()`.

neilschelly:
I'm nearly done with doing it in the views, and I don't dislike it as much as I expected. Really, the only extension that would be handled here in the CSR phase is the SubjectKeyIdentifier, and that's basically just added once it can be calculated.

Take a look at create_csr() here: https://github.com/neilschelly/lemur/blob/921b7617003f0436ffab8f277de6fd0026b529cc/lemur/certificates/service.py

If you don't like it, I can try another refactor to move the marshaling back into create_csr, using the marshmallow functions as a 3rd attempt. I worry that makes the code in create_csr() less readable. It potentially requires more understanding of marshmallow and the marshaling by anyone trying to read/understand create_csr().

kevgliss:
Your example looks good to me!

neilschelly:
Alright, I'll keep focusing on getting tests updated to pass this then before updating that branch to be a cleaner merge to this PR. There appear to be a couple pathways where tests expect to be able to submit dicts to create_csr, causing it to fail now that those are expected to be ExtensionTypes. And I think some of my serialize/deserialize relationships aren't as two-way-equal as they should be.

neilschelly:
So the reissue_certificate stuff (https://github.com/Netflix/lemur/blob/master/lemur/certificates/service.py#L533) expects to be able to duplicate a Certificate object, and re-submit it's data, modified slightly, to the create_csr() function. That is a problem since the Certificate.extensions object property (https://github.com/Netflix/lemur/blob/master/lemur/certificates/models.py#L231) is really limited. It only does SANs, doesn't do any SANs except DNSNames, and it returns them in the dict format instead of the x509 ExtensionType objects.

I'll have to revisit tomorrow to extend the Certificate class so it can properly extract all the supported ExtensionType objects out of the list of extensions. This also bumps into an oddity in the data model for a certificate, where many fields of data are stored twice, effectively, since the body of the certificate encodes all the things like issuer, serial, cn, extensions, dates, etc. That's okay from an ease-of-use standpoint in accessing Certificate objects, but which is the preferred canonical source of truth when reading a Certificate object?

neilschelly:
Alright, passes tests and the x509 ExtensionTypes are now created in the validation marshmallow. I'm ready for this to merge if you like it.

kevgliss:
Looks good, will deploy it to our test env to shake out any integration issues.

PR #674, opened 2017-01-24
Gracefully handle importing certificates with missing data

nevins-b:
Some certificates we have, which are generated by CFSSL, are missing the common name attribute. 

This PR should allow for handling that case as well as other missing attributes.

### DIFF
--- a/lemur/certificates/verify.py
+++ b/lemur/certificates/verify.py
@@ -7,6 +7,7 @@
 """
 import requests
 import subprocess
+from requests.exceptions import ConnectionError
 from cryptography import x509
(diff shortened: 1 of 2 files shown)


lemur/certificates/verify.py
```diff
@@ -7,6 +7,7 @@
 """
 import requests
 import subprocess
+from requests.exceptions import ConnectionError
```
kevgliss: Why was this added?

kevgliss: Ah I see now. It's has to do with a different issue.

kevgliss:
lgtm thanks!

PR #676, opened 2017-01-27
Added migration to cover what seem to be missing fields on the 'endpoints' object

tlianza:


kevgliss:
Thanks for the catch!

PR #685, opened 2017-02-03
adding generic OAuth2 provider

nevins-b:


kevgliss:
Could you also add some documentation about okta? It looks very much like vanilla oauth2, perhaps highlight differences (if any?)

nevins-b:
Sure thing! I think it is basically the vanilla oauth2 so if it makes more sense renaming this something more generic I'd be fine with that.

kevgliss:
Yeah, perhaps renaming it simple oauth2 would be best.

kevgliss:
Awesome, thanks!

PR #689, opened 2017-02-10
Issue 688 cert templates

### LINKED ISSUE #688: Certificate templates have unexpected consequences
I found a few inconsistencies with using the templates for client and server templates in creating a certificate.

1. If you create some SAN records on the first page of the certificate wizard, and then get to the second page and select a template, those SAN records will be cleared.
2. If you select a template, and extra Key Usage and an extra Extended Key Usage called isCritical will be submitted in the POST data. The critical bit of those extensions isn't exposed in the UI, so it shouldn't be here. If it were exposed, it wouldn't be modeled as if it were another checkbox among the Key Usage or Extended Key Usage options. The extra variable name is ignored, but does trigger a logged error in parsing.

neilschelly:
Closes #688 .

### DIFF
--- a/lemur/static/app/angular/certificates/certificate/certificate.js
+++ b/lemur/static/app/angular/certificates/certificate/certificate.js
@@ -166,11 +166,9 @@ angular.module('lemur')
         'extensions': {
           'basicConstraints': {},
           'keyUsage': {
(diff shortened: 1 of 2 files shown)


kevgliss:
lgtm thanks!

PR #690, opened 2017-02-13
allow attributes to be excluded from a cert subject

gesquive:
Currently when creating a certificate via the API, if a user omits or passes an empty string on an attribute like "location". The API returns errors even though a certificate does not require these attributes to be part of the subject.

I needed to be able to create certificates without certain NameAttributes in the subject. This PR allows users to specify empty strings for the `organization`, `organizational_unit`, `country`, `state`, `location` attributes via the API.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -337,15 +337,15 @@ def create_csr(**csr_config):
     builder = x509.CertificateSigningRequestBuilder()
     name_list = [x509.NameAttribute(x509.OID_COMMON_NAME, csr_config['common_name']),
                  x509.NameAttribute(x509.OID_EMAIL_ADDRESS, csr_config['owner'])]
-    if 'organization' in csr_config and len(csr_config['organization'].strip()) > 0:
+    if 'organization' in csr_config and csr_config['organization'].strip():
         name_list.append(x509.NameAttribute(x509.OID_ORGANIZATION_NAME, csr_config['organization']))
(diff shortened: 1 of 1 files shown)


lemur/certificates/service.py
```diff
@@ -335,15 +335,19 @@ def create_csr(**csr_config):
...
-        x509.NameAttribute(x509.OID_ORGANIZATION_NAME, csr_config['organization']),
-        x509.NameAttribute(x509.OID_ORGANIZATIONAL_UNIT_NAME, csr_config['organizational_unit']),
-        x509.NameAttribute(x509.OID_COUNTRY_NAME, csr_config['country']),
-        x509.NameAttribute(x509.OID_STATE_OR_PROVINCE_NAME, csr_config['state']),
-        x509.NameAttribute(x509.OID_LOCALITY_NAME, csr_config['location']),
-        x509.NameAttribute(x509.OID_EMAIL_ADDRESS, csr_config['owner'])
-    ]))
+    name_list = [x509.NameAttribute(x509.OID_COMMON_NAME, csr_config['common_name']),
+                 x509.NameAttribute(x509.OID_EMAIL_ADDRESS, csr_config['owner'])]
+    if 'organization' in csr_config and len(csr_config['organization'].strip()) > 0:
```
kevgliss: I believe this can just be `if 'organization' in csr_config and csr_config['organization'].strip()`

gesquive:
The latest commit includes the suggestions.

kevgliss:
Awesome, thanks!

PR #692, opened 2017-02-16
UI adjustments for mutually exclusive (radio button version) encipher/decipher-only Key Usage #664 

neilschelly:
This is an alternate solution to https://github.com/Netflix/lemur/issues/663 using radio buttons instead of mutually exclusive checkboxes. This PR is an alternate to https://github.com/Netflix/lemur/pull/664. Both PRs cannot be accepted. I like this one more.

The code underlying them is a bit more complicated, but the UI should look more comfortably aligned with the options available for KeyUsage.

### DIFF
--- a/lemur/common/fields.py
+++ b/lemur/common/fields.py
@@ -162,6 +162,9 @@ class KeyUsageExtension(Field):
             elif k == 'useCRLSign':
                 keyusages['crl_sign'] = v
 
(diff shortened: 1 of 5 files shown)


kevgliss:
This looks good, at some point we should really refactor this to be it's own component to be shared between the authority and certificate workflows. That doesn't need to be a part of this change though. Thanks!

Issue #703, opened 2017-03-03
RFC 5280 compliance with SAN extension


neilschelly:
From RFC 5280:

```
   If the subjectAltName extension is present, the sequence MUST contain
   at least one entry.  Unlike the subject field, conforming CAs MUST
   NOT issue certificates with subjectAltNames containing empty
   GeneralName fields.  For example, an rfc822Name is represented as an
   IA5String.  While an empty string is a valid IA5String, such an
   rfc822Name is not permitted by this profile.  The behavior of clients
   that encounter such a certificate when processing a certification
   path is not defined by this profile.
```

Certificates created with Lemur without any SANs are created with an empty list of SANs in the SubjectAlternativeName extension. See here these spots where that's essentially a hard-coded default:

* https://github.com/Netflix/lemur/blob/master/lemur/common/fields.py#L396
* https://github.com/Netflix/lemur/blob/master/lemur/certificates/models.py#L244

I find these work in most places, but I have a suspicion they do not work in OpenJDK 7+ keystores. I am working on confirming this now, but it at least violates the RFC.

kevgliss:
The cryptography plugin is probably most affected by this. External CAs (Digicert, Symantec, etc.) will enforce this behavior by adding whatever is in the common name as a subAltName entry I would assume we could do same.

neilschelly:
Here's some proof:
* CA certificate with no SAN component: https://gist.github.com/neilschelly/1f22eef17e88fb589954b1ccf785424d
* CA certificate with an empty list of SANs: https://gist.github.com/neilschelly/eb51cc6ecb3adc48b16101e59ddbdbd7

The following is the command output of using the Java keytool to parse the empty SAN certificate:

```
$ keytool -printcert -v -file ca-emptysan.crt
keytool error: java.lang.Exception: Failed to parse input
java.lang.Exception: Failed to parse input
        at sun.security.tools.keytool.Main.printCertFromStream(Main.java:2208)
        at sun.security.tools.keytool.Main.doPrintCert(Main.java:2337)
        at sun.security.tools.keytool.Main.doCommands(Main.java:1052)
        at sun.security.tools.keytool.Main.run(Main.java:337)
        at sun.security.tools.keytool.Main.main(Main.java:330)
Caused by: java.security.cert.CertificateParsingException: java.io.IOException: No data available in passed DER encoded value.
        at sun.security.x509.X509CertInfo.<init>(X509CertInfo.java:168)
[...]

PR #711, opened 2017-03-07
Issue #703 bugfix

neilschelly:
These fixes resolve the concerns in https://github.com/Netflix/lemur/issues/703.

1. If the SAN extension does not contain the CN part of the subject, it will be added as a DNSName. The side effect here is that you can never have an empty SAN extension.
1. Found a bug in creating certificates where the subAltNames tracked in the web UI were getting dropped. If you added SANs on the first page, then changed the template drop-down on the second page, you'd lose the SANs.
1. I also removed the allowance lemur_cryptography used to have where you could have both types of AuthorityKeyIdentifier (keyid and serial/issuer) at once. I have been unable to make a certificate validate there, and the RFC does suggest it should be one or the other.

### DIFF
--- a/lemur/plugins/lemur_cryptography/plugin.py
+++ b/lemur/plugins/lemur_cryptography/plugin.py
@@ -157,7 +157,7 @@ def normalize_extensions(csr):
 
     # Remove original san extension from CSR and add new SAN extension
     extensions = list(filter(filter_san_extensions, csr.extensions._extensions))
-    if not san_extension is None and len(san_extension.value._general_names) > 0:
+    if san_extension is not None and len(san_extension.value._general_names) > 0:
         extensions.append(san_extension)
 
     return extensions
--- a/lemur/plugins/lemur_cryptography/plugin.py
+++ b/lemur/plugins/lemur_cryptography/plugin.py
@@ -131,14 +131,13 @@ def normalize_extensions(csr):
         san_extension = csr.extensions.get_extension_for_oid(x509.oid.ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
         san_dnsnames = san_extension.value.get_values_for_type(x509.DNSName)
     except x509.extensions.ExtensionNotFound:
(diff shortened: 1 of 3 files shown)


neilschelly:
Currently, this PR doesn't fix another problem I've uncovered with creating an authority with lemur_cryptography. I created an issue for that here: https://github.com/Netflix/lemur/issues/712. I'm unsure how to fix it immediately.

neilschelly:
Running into issues duplicating the Subject CN into the SAN extension if it has spaces. That won't really happen with traditional servername certificates, but could happen with client certs or CA certs where the CN will commonly have spaces. Trying to gracefully skip adding the CN to the SAN in those cases and then remove the SAN extension if nothing is left.

I've decided I'm over-complicating this and running out of talent today. Will take a further stab tomorrow.

neilschelly:
This should be good to commit now. It provides a bit more functionality in the certificate signing process in lemur_cryptography to parse/modify other extensions too, if needed. The biggest thing to work around was to ensure no common names with spaces in them somehow snuck into the SAN extension as DNS Names. The ASN1 parsing way underneath things would continually fail with the most useless exception messages imaginable.

kevgliss:
The only other thing you might also want to check for is `_` in the common name, idns/ASN1 parsing will barf on these as well. 

This might be added to Lemur at a higher level than your code though, I'm just not quite sure how many people are using underscores in their certificates as some CAs allow it.

neilschelly:
The problem with making sure that the Common Name is a valid hostname is that it doesn't actually have to be. That only applies to certificates used for an FQDN. CAs and client certificates and code-signing certificates are all pretty unlikely to put hostnames in the CommonName field.

The SAN fields _must_ be valid for the type of data they hold, so you cannot always assume that the CommonName should be a valid DNSName to insert into the SAN extension. That is common practice for server name certificates of course, but it's a bit of a hack as a result of world abusing the CommonName as server name for so long.

PR #723, opened 2017-03-13
specify python3 when creating virtualenv

RickB17:
Lemur is developed against Python3.5. If you do not specify the Python version it is possible the virtualenv will be built on a different version.


kevgliss:
Thanks!

PR #725, opened 2017-03-14
added steps for loading custom plugin

RickB17:
added steps for loading a custom plugin into Lemur once the files have been put into place (/www/lemur/lemur/plugins/) and the setup.py file (/www/lemur/setup.py) has been modified.


kevgliss:
Thanks!

PR #735, opened 2017-03-24
adding plugin/lemur_linuxdst

RickB17:
A destination plugin that will will export the certificate to a remote linux host. A sub-folder will be created under the destination directory based on the certificates cn and it will contain two files, cert.pem and priv.key.


lemur/plugins/lemur_linuxdst/remote_host.py
```diff
@@ -0,0 +1,32 @@
...
+        certOut.write(lemCert.body)
+    certOut.close()
+    keyOut = open(keyFile, "w+")
+    keyOut.write(lemCert.private_key)
+    keyOut.close()
+    return {'certDir': '{0}/{1}'.format(tempFolder, lemCert.cn)}
+
+
+def copyCert(dstUser, dstHost, dstDir, certDir, options, **kwargs):
+    os.system('scp -r {0} {1}@{2}:{3}'.format(certDir, dstUser, dstHost, dstDir))
```
RickB17: Yes. There is an assumption that the lemur host has the private key matching the public key in ~/.ssh/authorized_keys on the remote host.

lemur/plugins/lemur_linuxdst/remote_host.py
```diff
@@ -0,0 +1,32 @@
...
+
+def createCert(name, tempFolder, exportType):
+
+    lemCert = service.get_by_name(name)
+    if not os.path.exists(tempFolder):
+        os.mkdir(tempFolder)
+    if not os.path.exists('{0}/{1}'.format(tempFolder, lemCert.cn)):
+        os.mkdir('{0}/{1}'.format(tempFolder, lemCert.cn))
+    certFile = '{0}/{1}/cert.pem'.format(tempFolder, lemCert.cn)
+    keyFile = '{0}/{1}/priv.key'.format(tempFolder, lemCert.cn)
```
RickB17: I agree. I'll check it out.

RickB17: @kevgliss  have you had any luck using this library with a non-standard sshd port?  It keeps yelling at me and I cannot find out what known_hosts it is referencing. I can use ssh to connect to the target machine with the standard ssh command no problem. I tired /etc/ssh/known_hosts and ~/.ssh/knownhosts

paramiko.ssh_exception.SSHException: Server '[target.domain.com]:2299' not found in known_hosts

RickB17: I was able to resolve it with : 
image
my concern is that it's automatically accepting the host key :-(

RickB17: @kevgliss  when walking though the link you provide it appears it still wants a file to transfer. Am I overlooking something?

RickB17:
cloned by https://github.com/Netflix/lemur/pull/736

PR #736, opened 2017-03-25
Destination Plugin/Lemur_linuxdst

RickB17:


RickB17:
Still working on this code.

RickB17:
@kevgliss  When you have a moment, could you please take a look at this update? Let me know if you feel these changes address the security concerns of using a local temp.

lemur/plugins/lemur_linuxdst/remote_host.py
```diff
@@ -0,0 +1,56 @@
...
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    if dst_priv_key is None:
+        priv_key = paramiko.RSAKey.from_private_key_file(dst_priv)
+    else:
+        priv_key = paramiko.RSAKey.from_private_key_file(dst_priv, dst_priv_key)
+    ssh.connect(dst_host, username=dst_user, port=dst_port, pkey=priv_key)
+    sftp = ssh.open_sftp()
+    try:
+        sftp.mkdir(dst_dir)
+    except IOError:
```
kevgliss: Is the assumption here that the directory already exists? If this fails for some other IO related reason what will happen?

lemur/plugins/lemur_linuxdst/plugin.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def upload(self, name, body, private_key, cert_chain, options, **kwargs):
+
+        export_type = self.get_option('exportType', options)
+        dst_host = self.get_option('dstHost', options)
+        dst_host_port = self.get_option('dstPort', options)
+        dst_user = self.get_option('dstUser', options)
+        dst_priv = self.get_option('dstPriv', options)
+        dst_priv_key = self.get_option('dstPrivKey', options)
+        if len(dst_priv_key) < 1:
```
kevgliss: The comparison is unnecessary. `If len(dst_priv_key):` should be enough.

lemur/plugins/lemur_linuxdst/plugin.py
```diff
@@ -0,0 +1,84 @@
...
+    requires_key = False
+
+    def upload(self, name, body, private_key, cert_chain, options, **kwargs):
+
+        export_type = self.get_option('exportType', options)
+        dst_host = self.get_option('dstHost', options)
+        dst_host_port = self.get_option('dstPort', options)
+        dst_user = self.get_option('dstUser', options)
+        dst_priv = self.get_option('dstPriv', options)
+        dst_priv_key = self.get_option('dstPrivKey', options)
```
kevgliss: You might want to add some simple encryption to this value. For other use cases we use MultiFernet with keys Lemur already has access to. See: https://github.com/Netflix/lemur/blob/71ddbb409ce3d49e9fb33bea297ea1d777d95539/lemur/utils.py#L102

lemur/plugins/lemur_linuxdst/remote_host.py
RickB17: The assumption is that the directory doesn't exist and it will be created. So far I have tested two cases:
1) the directory doesn't exist yet
2) the directory exists because it was previously created from the plugin

lemur/plugins/lemur_linuxdst/plugin.py
RickB17: @kevgliss Could you help point me in the direction of where I can "catch" the input values passed as options to run them through the encryption process before writing them to the data base in the destination options? Can this be done in the plugin code, or does it need to be outside in the destination code?

kevgliss: There's no way to do this currently as we haven't had anything sensitive stored in plugin options to date. If you're interested in adding support for this you would 'catch' the value via the destination schema:
https://github.com/Netflix/lemur/blob/master/lemur/destinations/schemas.py#L28

Right now we don't do anything to the values before storing them if we wanted to add encryption support you could add it via a custom marshmallow field that would encrypt and decrypt the values during serialization and deserialization:
http://marshmallow.readthedocs.io/en/latest/custom_fields.html

You could then add an optionally add a value to the options dictionary `encrypt: True` that would only encrypt plugin options that are sensitive.

Thinking through this, it is quite a bit of work. If you decide to pursue it should be an in a separate PR.

RickB17: @kevgliss  is this a dependency to getting this merged?

kevgliss: No, if you are happy with the plugin. I would be happy to merge this in.

RickB17: @kevgliss I added a short note about this in the read me file. Looks like we should be good to go.

kevgliss:
Sound good, I will fix the PEP8 errors.

PR #737, opened 2017-03-28
adding url context path to build

henry-megarry-tr:
This will replace all requests for resources and api calls to include a context path. 
Example: 
  urlContextPath=lemur
  /api/1/auth/providers -> /lemur/api/1/auth/providers


kevgliss:
Interesting, is this in the case you are running multiple applications on a single server?

henry-megarry-tr:
We're running several tools behind Zuul and we need the context path for routing.

kevgliss:
Sounds good! Thanks!

Issue #741, opened 2017-04-03
Unable to create PR for 0.4.0 tag


treacher:
Hello :)

I'm looking to create a PR against version 0.4.0 it seems there is no 0.4.x branch for me to create a pull request against. I'm just wondering would this be the normal approach for creating this type of PR, this is a patch for fixing an issue with the Digicert plugin.

Currently it always sets the `custom_expiration_date` even if you only define `validity_years`. This is a problem as some setups in digicert will reject creation requests if they have a `custom_expiration_date` set. These setups only allow `validity_years` to be passed through.

The change I'm looking to make is on the following branch: https://github.com/Netflix/lemur/compare/0.4.0...treacher:custom_expiration_date-fix

Please advise :)

kevgliss:
Does the problem still exist in master? I would do a PR against HEAD, once that looks okay and is merged I will create a 0.4 branch and cherry pick your PR to backport it.

treacher:
Thanks for the fast response, yes it does, here is the PR: https://github.com/Netflix/lemur/pull/742

kevgliss:
I created the following branch with your PR applied. Please let me know if you into any issues.

https://github.com/Netflix/lemur/tree/0.4.1

treacher:
Is it possible to also cherry-pick the following commit into 0.4.x as well as it causes issues when spinning up a stack: https://github.com/Netflix/lemur/commit/d68b2b22e067066b010885b615d6e06ebab3933c due to a dependency issue.

kevgliss:
Done.

treacher:
It seems cherrypicking that commit so far back has caused quite a few issues a couple to name:
- Factory Boy needs upgrading `ImportError('The ``fake-factory`` package is now called ``Faker``.\n\nPlease update your requirements.\n',)`
- There is a Flakey time dependent test that seems to be failing: `AssertionError: assert not {'_schema': ['Validity end must not be after 2020-01-01 20:59:06']}` 
- It's now referencing methods that don't exist: `ImportError: cannot import name validate_conf`

I suspect those wont be the only or last ones, I wonder what you think we should do in this case? Is there a release coming soon where I could just roll forward with this rather than spending time on fixing something in a branch that's already fixed in the future? Or should I do a PR to fix these things on this branch?

kevgliss:
Master is currently pretty stable (it's what we've been running for a month or so), I just haven't gotten around to cutting a release. Let me try to find some time this weekend and I will get a release tagged.

[...]

PR #742, opened 2017-04-04
Don’t set ‘custom_expiration_date’ if validity years is set in the UI.

treacher:
This PR is for updating the Digicert Plugin so that when you define `validity_years` in the UI it no longer sets the `custom_expiration_date`. The reason for this is because not every digicert account allows for you to define a `custom_expiration_date` as it's a restriction on the account, it's also not required if you don't need a custom expiration date on your cert.

Other small changes:
- Explicitly set the `sub_alt_names` in the tests to be unicode (this is to fix a breaking test)
- Always define `validity_years` in `map_fields` as it's required in the [Digicert API](https://www.digicert.com/services/v2/documentation/order/order-ssl-determinator)
- Unset `validity_end` as it's no longer required if `validity_years` is already defined, by default `validity_end` seems to be getting set upstream, doing this makes it easier to determine whether or not to use `custom_expiration_date`.

This PR is in relation to this issue: https://github.com/Netflix/lemur/issues/741

Is it possible to also cherry-pick the following commit into `0.4.x` as well as it causes issues when spinning up a stack:  https://github.com/Netflix/lemur/commit/d68b2b22e067066b010885b615d6e06ebab3933c

### DIFF
--- a/lemur/plugins/lemur_digicert/tests/test_digicert.py
+++ b/lemur/plugins/lemur_digicert/tests/test_digicert.py
@@ -10,7 +10,7 @@ from cryptography import x509
 def test_map_fields_with_validity_end_and_start(app):
     from lemur.plugins.lemur_digicert.plugin import map_fields
 
-    names = [u"one.example.com", u"two.example.com", u"three.example.com"]
+    names = [u'one.example.com', u'two.example.com', u'three.example.com']
 
     options = {
         'common_name': 'example.com',
--- a/lemur/plugins/lemur_digicert/tests/test_digicert.py
+++ b/lemur/plugins/lemur_digicert/tests/test_digicert.py
@@ -7,10 +7,10 @@ from lemur.tests.vectors import CSR_STR
 from cryptography import x509
 
 
-def test_map_fields(app):
+def test_map_fields_with_validity_end_and_start(app):
     from lemur.plugins.lemur_digicert.plugin import map_fields
 
-    names = ['one.example.com', 'two.example.com', 'three.example.com']
+    names = [u"one.example.com", u"two.example.com", u"three.example.com"]
 
     options = {
         'common_name': 'example.com',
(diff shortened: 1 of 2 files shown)


lemur/plugins/lemur_digicert/tests/test_digicert.py
```diff
@@ -7,10 +7,10 @@
 from cryptography import x509
 
 
-def test_map_fields(app):
+def test_map_fields_with_validity_end_and_start(app):
     from lemur.plugins.lemur_digicert.plugin import map_fields
 
-    names = ['one.example.com', 'two.example.com', 'three.example.com']
+    names = [u"one.example.com", u"two.example.com", u"three.example.com"]
```
kevgliss: These should be single quotes.

lemur/plugins/lemur_digicert/tests/test_digicert.py
```diff
@@ -35,14 +35,47 @@ def test_map_fields(app):
...
-        'custom_expiration_date': arrow.get(2017, 5, 7).format('YYYY-MM-DD')
+        'custom_expiration_date': arrow.get(2017, 5, 7).format('YYYY-MM-DD'),
+        'validity_years': 1
+    }
+
+
+def test_map_fields_with_validity_years(app):
+    from lemur.plugins.lemur_digicert.plugin import map_fields
+
+    names = [u'one.example.com', u'two.example.com', u'three.example.com']
```
kevgliss: What error did you experience that required the change to Unicode?

treacher:
```
>>> from cryptography import x509
>>> x509.DNSName('foo.bar.com')
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/cryptography/x509/general_name.py", line 96, in __init__
    raise TypeError("value must be a unicode string")
TypeError: value must be a unicode string
>>> x509.DNSName(u'foo.bar.com')
<DNSName(value=foo.bar.com)>
>>>
```

kevgliss: Ah, I see this is a python2.7 issue. That sounds reasonable for the 0.4 branch (as python 2.7) as it's still supported there.

PR #744, opened 2017-04-05
supervisor ; cause services not to start

RickB17:
the ; in the supervisor/conf.d/app.conf file cause the service not to start.


kevgliss:
lgtm! Thanks.

PR #754, opened 2017-04-19
Update cron commands in Production documentation.

treacher:
Seems the api for these actions have changed. Thought I would update the documentation around this. Let me know if I've misunderstood something.


kevgliss:
Lgtm. Thanks!

PR #762, opened 2017-04-27
Return correct intermediate certificate on digicert creation.

pvdvreede:
At the moment when we create a new cert through the digicert plugin, the export created has the created certificate twice rather than the certificate and intermediate. This commit corrects the return value of the plugin for digicert.

This commit also removes the unused DIGICERT_INTERMEDIATE env
var as it is not used.


kevgliss:
lgtm thanks!

PR #763, opened 2017-04-27
Hexify cert serial

treacher:
It seems as though majority of CA's represent the serial as a hex value rather than a decimal value as does the `openssl` command. I suspect this is due to readability. This PR hexifies the serial at the serialization layer so that it's represented in the UI the same way it is within the CA UI. The reason I've done it at the serialization layer is this allows for reconciliation of certificate serials via the API if needed. I didn't update the serial in the database as it's nice to keep the original intended value.

Other changes:
- Fix for flakey test, the ordering of the keys is random so by sorting we remove the randomness.


kevgliss:
lgtm thanks!

PR #771, opened 2017-04-28
Add docker setup for running tests on a docker enabled dev environment.

pvdvreede:
We use docker and docker-compose for development. This PR adds the requisite files to enable development without needing to install python, node, postgres. It should not alter a non-docker dev workflow, but merely enable docker dev as an option if that is your thing.


kevgliss:
This seems cool if you have time/inclination I wouldn't mind some documentation around developing Lemur with docker. 

https://github.com/Netflix/lemur/blob/master/docs/developer/index.rst

PR #772, opened 2017-04-28
config LEMUR_MAIL to LEMUR_EMAIL

RickB17:
I referenced https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_email/plugin.py and it appears this configuration option should be "LEMUR_EMAIL"


kevgliss:
lgtm thanks!

Issue #752, opened 2017-04-12
Internal Server Error - Adding certificates to notifications


kevgliss:
```
2017-04-12 17:35:33,074 ERROR: Exception on /api/1/notifications/950 [PUT] [in /apps/lemur/lib/python3.5/site-packages/flask/app.py:1560]
Traceback (most recent call last):
  File "/apps/lemur/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/apps/lemur/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/apps/lemur/lib/python3.5/site-packages/flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "/apps/lemur/lib/python3.5/site-packages/flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "/apps/lemur/lib/python3.5/site-packages/flask_restful/__init__.py", line 587, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/apps/lemur/lemur/auth/service.py", line 110, in decorated_function
    return f(*args, **kwargs)
  File "/apps/lemur/lemur/common/schema.py", line 150, in decorated_function
    data, errors = input_schema.load(request_data)
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/schema.py", line 580, in load
    result, errors = self._do_load(data, many, partial=partial, postprocess=True)
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/schema.py", line 660, in _do_load
    index_errors=self.opts.index_errors,
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/marshalling.py", line 295, in deserialize
    index=(index if index_errors else None)
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/marshalling.py", line 68, in call_and_store
    value = getter_func(data)
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/marshalling.py", line 288, in <lambda>
    data
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/fields.py", line 265, in deserialize
    output = self._deserialize(value, attr, data)
  File "/apps/lemur/lib/python3.5/site-packages/marshmallow/fields.py", line 465, in _deserialize
    data, errors = self.schema.load(value)
[...]

PR #774, opened 2017-04-30
fix unit and interval transposition in schemas.py (#752)

pr8kerl:
I found the same error as in #752 when trying to add a new slack notification.
**interval** and **unit** appear to be transposed in validate_options.
I've tested briefly; after this fix I could successfully add a certificate to a notification, as well as add the new slack notification.


kevgliss:
Awesome thanks! Closes #752

PR #776, opened 2017-05-02
Updated apache config in documentation

csv7:
You guys asked for one that worked... It took me a little while to tweak, esp. since I'm not a guru with python.  The comment about needing mod_wsgi isn't true, unless you want to run lemur as a cgi program... I suspect that's from an older version that ran as cgi and not as a standalone webserver.

### DIFF
--- a/docs/production/index.rst
+++ b/docs/production/index.rst
@@ -217,6 +217,23 @@ An example apache config::
         # HSTS (mod_headers is required) (15768000 seconds = 6 months)
         Header always set Strict-Transport-Security "max-age=15768000"
         ...
+        
+     # Set the lemur DocumentRoot to static/dist
+     DocumentRoot /www/lemur/lemur/static/dist
+     
+     # Uncomment to force http 1.0 connections to proxy
+     # SetEnv force-proxy-request-1.0 1
+     
+     #Don't keep proxy connections alive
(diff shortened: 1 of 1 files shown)


kevgliss:
Thanks, if you're pretty confident perhaps we should remove the note about it being incomplete?

csv7:
I think it'd be ok to remove. I don't see where mod_wsgi is needed -- unless I'm missing something completely unobvious. We're running a similar python application -- Internet2's FileLocker https://sourceforge.net/projects/filelocker2/  -- which I had to tweak heavily to get it to work "just right" in our environment... Since both of them are stand-alone webservers (filelocker uses cherryPi), I used it's apache proxy config as a starting point. We've been running it for almost a year.

csv7:
I'm going to try my hand at making a couple of plugins for Lemur... I've been using a home-grown solution to nmap our network on known ports and import ssl certs into a database and try to keep up with it... but maintaining that codebase sucks. openssl coding is the devil. Also, I need some way of automating making Internet2 Incommon certs, and Lemur could do it with a little help.

kevgliss:
Cool, I think an Nmap source plugin could be cool... we have traditionally kept that functionality out of Lemur and had other applications handle scanning infrastructure. But given your use case, a sync plugin would fit the bill.

PR #778, opened 2017-05-05
Fix for status representation in the view

treacher:
Fixes:
- Change status to revoked instead of invalid as that's what the frontend is looking for.
- Change the schema to represent the status as a string rather than a boolean as otherwise it's always true and never shows anything in the UI.


kevgliss:
lgmt thanks!

PR #780, opened 2017-05-07
Check for unknown as status is no longer represented as a boolean

treacher:
This PR is in relation to https://github.com/Netflix/lemur/pull/778 it updates the view to represent 'unknown' correctly.


kevgliss:
lgtm thanks!

PR #781, opened 2017-05-08
Log the audit logs

treacher:
This PR is for logging when we create log records in the database. The reason for this is because if an attacker infiltrates your system they would be able to change the database entries where as if it's going to your central logging system you'll have a trail.


kevgliss:
lgtm thanks!

Issue #792, opened 2017-05-12
Adding domain fails on unselectable "sensitive"


arthurlogilab:
Client side

![screenshot_2017-05-12_11-06-54](https://cloud.githubusercontent.com/assets/445200/25991405/528d417a-3703-11e7-9e6c-d70beb6d38e2.png)

Server side

```
May 12 09:05:48 lemur supervisord: lemur-web [2017-05-12 09:05:48,892] ERROR in schema: 'sensitive'
May 12 09:05:48 lemur supervisord: lemur-web Traceback (most recent call last):
May 12 09:05:48 lemur supervisord: lemur-web   File "/var/www/lemur/lemur/common/schema.py", line 158, in decorated_function
May 12 09:05:48 lemur supervisord: lemur-web     resp = f(*args, **kwargs)
May 12 09:05:48 lemur supervisord: lemur-web   File "/var/www/lemur/lemur/domains/views.py", line 126, in post
May 12 09:05:48 lemur supervisord: lemur-web     return service.create(data['name'], data['sensitive'])
May 12 09:05:48 lemur supervisord: lemur-web KeyError: 'sensitive'
May 12 09:05:48 lemur supervisord: lemur-web
May 12 09:05:48 lemur supervisord: lemur-web 'sensitive'
May 12 09:05:48 lemur supervisord: lemur-web Traceback (most recent call last):
May 12 09:05:48 lemur supervisord: lemur-web   File "/var/www/lemur/lemur/common/schema.py", line 158, in decorated_function
May 12 09:05:48 lemur supervisord: lemur-web     resp = f(*args, **kwargs)
May 12 09:05:48 lemur supervisord: lemur-web   File "/var/www/lemur/lemur/domains/views.py", line 126, in post
May 12 09:05:48 lemur supervisord: lemur-web     return service.create(data['name'], data['sensitive'])
May 12 09:05:48 lemur supervisord: lemur-web KeyError: 'sensitive'
```

mikegrima:
Made a PR for this is at #795

PR #795, opened 2017-05-15
Possible fix for #792

mikegrima:
This resolved the error seen in #792.

However, after adding domains, we were unable to search for them using the domain UI.  We were able to, however, see the domains in the database.


kevgliss:
This has been pulled in via #796. I have also fixed the secondary issue to mention in #797.

PR #812, opened 2017-05-22
Add Check of DB connections on healthcheck URL

paulieborg:


kevgliss:
What was the use case for this? Perhaps we should pull an actual value from the database, for instance, if that database could connect but there are not tables is it health?

paulieborg:
Heh @kevgliss 
The AWS ELB LoadBalancer health check will still pass even if the DB is down.
For sure we can add an actual table to check further. This PR is a basic 'eyes open' check to start off with and not necessarily intend at this point to do a functional check.
Paul

treacher:
@kevgliss from my understanding by default when the application starts if the schema is not in the state it's expecting the application will fail to start, so I don't believe we need to be checking that as frequently as the connection to the database. I think `SELECT 1` is a pretty common query to use for a health check to make sure the application has access to the database as it's a light weight query.

kevgliss:
Sounds good to me.

PR #814, opened 2017-05-24
adding url context path to html templates

henry-megarry-tr:
Something changed with how the pager.html was retrieved. This is a fix for the urlContextPath changes I put in a while ago. Also adding a clarification to the documentation.


kevgliss:
lgtm

PR #830, opened 2017-06-05
Fix AWS S3 Destination plugin 

paulieborg:
AWS s3.put function wants named arguments for the following values:
account_number, region, bucket_name, prefix, data and encrypt

Previously they were inline args which did not work with existing functionality.


paulieborg:
Also, I need to rename the value of `accountName` in the `upload` function to `account_name` so it can be properly used in the sts_client function

kevgliss:
lgtm thanks!

Issue #791, opened 2017-05-12
[featurerequest] LDAP support for users


arthurlogilab:
It would be useful to have an LDAP connector to handle users.

douardda:
Or any way to delegate auth to the reverse proxy in front of the flask app, in fact.

kevgliss:
This looks like it would be pretty easy to add.

- Add a new LDAP login provider [Examples](https://github.com/Netflix/lemur/blob/master/lemur/auth/views.py#L30)
- Add required configuration parameters
- Add additional dependency on `python-ldap`

I'd be happy to merge something like this.

PR #842, opened 2017-06-28
simple ldap config

pr8kerl:
@kevgliss  Just after some feedback atm. This is for #791 ldap support.
I've got it working for me against Active Directory. Haven't tested against anything else.

In the issue you mention 'provider' which confused me a bit. LDAP can't be handled quite like Google or Oauth. As LDAP is a backend authentication server, I've put the code within the auth Login class. This way the standard basic auth form can be used. If the user has a local database definition with matching password there, that will be used. Then if LDAP is configured and no local user matches, ldap will be tried.i Not sure if you will be happy with this approach. Seemed reasonable to me.

Also with the dependencies, this will now need the following packages installed to build successfully: libldap2-dev libsasl2-dev libldap2-dev libssl-dev
Any issues with that??

I also can't see any relevant test to update. Can try to create something if you think its needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ RUN apt-get update
 RUN apt-get install -y make python-software-properties curl
 RUN curl -sL https://deb.nodesource.com/setup_7.x | bash -
 RUN apt-get update
-RUN apt-get install -y nodejs
+RUN apt-get install -y nodejs libldap2-dev libsasl2-dev libldap2-dev libssl-dev
 RUN pip install -U setuptools
 RUN pip install coveralls bandit
 WORKDIR /app
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -251,7 +251,108 @@ Lemur supports sending certification expiration notifications through SES and SM
 
 Authentication Options
 ----------------------
-Lemur currently supports Basic Authentication, Ping OAuth2, and Google out of the box. Additional flows can be added relatively easily.
+Lemur currently supports Basic Authentication, LDAP Authentication, Ping OAuth2, and Google out of the box. Additional flows can be added relatively easily.
+
+LDAP Specific Options
+~~~~~~~~~~~~~~~~~~~~~
+
+Lemur supports the use of an LDAP server in conjunction with Basic Authentication. Lemur local users can still be defined and take precedence over LDAP users. If a local user does not exist, LDAP will be queried for authentication. Only simple ldap binding with or without TLS is supported.
+
+LDAP support requires the pyldap python library, which also depends on the following openldap packages.
+
+.. code-block:: bash
+
+      $ sudo apt-get update
+      $ sudo apt-get install libldap2-dev libsasl2-dev libldap2-dev libssl-dev
+
+
+To configure the use of an LDAP server, the following settings must be defined.
+
+.. data:: LDAP_AUTH
+    :noindex:
+
+        This enables the use of LDAP
+
+        ::
+
+            LDAP_AUTH = True
+
+.. data:: LDAP_BIND_URI
+    :noindex:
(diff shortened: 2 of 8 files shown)


kevgliss:
This looks really good, I don't really have too much to add. 

Looking at that file it has grown quite a bit and will probably need some refactoring soon (user creation, role attachment). 

I would also add that I don't think the function `ldap_login` should be attached to the resource class as that function doesn't look to be using anything from `self`. At the very least I'd expect it to be a standalone function, perhaps even in its own .py file.

Testing wise, I don't really have very many good tests for those particular paths. Obviously, anything you can reasonably unit-test would be helpful as I don't have an LDAP environment available for testing. 

Let me know when you're happy with this PR and I will get it merged in.

pr8kerl:
Thanks. That's good feedback. I'll try to clean things up in line with your comments.

Cheers.

kevgliss:
@pr8kerl any update on this PR?

pr8kerl:
@kevgliss Apologies for my tardiness. I haven't forgotten - just other priorities getting in the way.
I've only recently got this installed on our internal server. So far so good. And I want to try and do some more unit tests - something I'm learning as I'm not a python guy usually.
Will try to knock it off within the next week.

kevgliss:
Sounds good. I am planning to do a release in week or so. Would love to include this!

pr8kerl:
@kevgliss I just pushed my latest. Think it's good. My tests are probably suspicious, but I think they do a little bit of valid testing. I haven't done much testing in python before.

However, one thing that is a little disconcerting is that now on my internal lemur server, I am now having some db migration issues. To get to where I am I rebased from master and added in my ldap changes. I have an existing database there with quite a few certs already.
On startup it seems to do a migration and create the new rotation policy stuff.
```
/usr/local/lib/python3.5/site-packages/alembic/util/messaging.py:69: UserWarning: Revision a1dbe3bf65a7 is present more than once
   warnings.warn(msg)
INFO  [alembic.runtime.migration] Running stamp_revision 8ae67285ff14 -> a1dbe3bf65a7
```
Then if I restart, it seems to run the migration again and the lemur service fails due to the rotation policy column already there. Any thoughts?? I can raise another issue with all the info if you think.
Thanks.

kevgliss:
Will take a look. Yes please file a separate issue with your migration issue and we'll get you sorted.

Dockerfile
```diff
@@ -3,7 +3,7 @@ RUN apt-get update
 RUN apt-get install -y make python-software-properties curl
 RUN curl -sL https://deb.nodesource.com/setup_7.x | bash -
 RUN apt-get update
-RUN apt-get install -y nodejs
+RUN apt-get install -y nodejs libldap2-dev libsasl2-dev libldap2-dev libssl-dev
```
kevgliss: I wonder if we should pull all of these in by default, maybe documentation only is a better approach here?

lemur/auth/views.bak
```diff
@@ -0,0 +1,720 @@
+"""
```
kevgliss: This whole file looks like it's unneeded.

Dockerfile
pr8kerl: Yes, I'm also unsure of the best way to handle these. They are required by the python ldap module.
This Dockerfile was put there to allow easy running of the tests. Without installing these in the docker image, the tests fail to run - pyldap fails to install.
I guess the travis servers have everything installed already so no issue there.
I've left this as is for now.

lemur/auth/views.bak
pr8kerl: oops - my bad. tx.
Gone now.

kevgliss:
lgtm thanks!

PR #845, opened 2017-07-03
Add log message on creation or update of certificate.

pvdvreede:
We would like to be able to put audit logs out of band of the lemur servers, so this PR will create a log entry when a cert is created or updated. This in turn will log an entry to the logs which will go to a central logging point.


kevgliss:
lgtm thanks!

PR #876, opened 2017-08-16
Prevent unintended access to sensitive fields (passwords, private keys)

intgr:
Make sure that fields specified in filter, sortBy, etc. are model fields
and may be accessed. This is fixes a potential security issue.

The filter() function allowed guessing the content of password hashes
one character at a time.

The sort() function allowed the user to call an arbitrary method of an
arbitrary model attribute, for example sortBy=id&sortDir=distinct would
produce an unexpected error.


intgr:
Sorry about the long hiatus. This time truly I hope I can commit significant time to upstreaming our Lemur modifications.

kevgliss:
This looks like a great improvement thanks!

PR #878, opened 2017-08-16
Reworked sensitive domain name and restriction logic

intgr:
Finally a follow-up to pull request #600.

* This is a fix for a potential security issue; the old code had edge
  cases with unexpected behavior.
* LEMUR_RESTRICTED_DOMAINS is no more, instead LEMUR_WHITELISTED_DOMAINS
  is a list of *allowed* domain name patterns. Per discussion in PR #600
* Domain restrictions are now checked everywhere: in domain name-like
  CN (common name) values and SAN DNSNames, including raw CSR requests.
* Common name values that contain a space are exempt, since they cannot
  be valid domain names.


lemur/common/validators.py
```diff
@@ -41,22 +42,33 @@ def private_key(key):
         raise ValidationError('Private key presented is not valid.')
 
 
+def common_name(value):
+    """If the common name could be a domain name, apply domain validation rules."""
+    # Common name could be a domain name, or a human-readable name of the subject (often used in CA names or client
+    # certificates). As a simple heuristic, we assume that human-readable names always include a space.
+    # However, to avoid confusion for humans, we also don't count spaces at the beginning or end of the string.
```
intgr: I hope you are comfortable with this heuristic; matches my experience for how the subject CN field is used.

kevgliss: I'm a bit confused by this, are you saying that all of your users that enter a common name happen to have a space in them? We often see myCommonName in client certificates (java programmers). So I'm not sure this would work for us. Could this heuristic be changed to look for '.' in the common name and then attempt to validate? 

```
if '.' in value:
    return sensitive_domain(value)
```
I think there is a strong assumption that if you are entering periods into a common name an FQDN is expected.

intgr: The problem with your proposal is that `myCommonName` or `com` is also a valid domain name, and if implemented that way, it would be impossible to restrict that. In a corporate setting, where workstations have configured a domain search path, simple strings like that could resolve to an actual service.

My point was to err on the side of caution: only skip whitelist check if it's clear that it couldn't be a domain name.

kevgliss: Okay, that makes more sense.

intgr:
Thanks!

PR #879, opened 2017-08-17
Fix roles display on user screen and fix removing user roles

intgr:


lemur/users/schemas.py
```diff
@@ -26,6 +26,7 @@ class UserOutputSchema(LemurOutputSchema):
     username = fields.String()
     email = fields.Email()
     active = fields.Boolean()
+    roles = fields.Nested(AssociatedRoleSchema, many=True)
```
intgr: It seems this was entirely broken since #633.

kevgliss:
Thanks!

PR #880, opened 2017-08-17
Fix ability to remove all roles from authority

intgr:


lemur/authorities/service.py
```diff
@@ -26,12 +26,11 @@ def update(authority_id, description=None, owner=None, active=None, roles=None):
     """
     authority = get(authority_id)
 
-    if roles:
-        authority.roles = roles
```
intgr: Why was this "if" here in the first place? To support partial updates? Looks like that isn't possible as the code is written now, so I decided to make all the function arguments required too.

PR #881, opened 2017-08-17
Disable unused Flask Principal sessions

intgr:
Lemur uses its own auth token for authentication; logging out doesn't
properly dispose of the Flask Principal session.


kevgliss:
Thanks!

PR #882, opened 2017-08-17
Create signal hooks and handler for dumping CSR and certificate details

intgr:
This has been quite useful for debugging various Lemur integrations and CA configuration: who sends/requests what and how the certificate looks like in the end. Since it shows more information than Lemur itself.

I don't know if you want to merge this whole. Alternatively you I could only submit the signals part and I can refactor the rest out into a plugin for myself.


kevgliss:
This looks generally useful, thanks!

PR #883, opened 2017-08-18
Prevent creation of empty SubjAltNames extension in CSR

intgr:
I don't remember exactly how, but the empty SubjAltNames extension caused problems. And it serves no useful purpose anyway.


kevgliss:
lgtm thanks!

PR #906, opened 2017-09-08
Improve certificate name normalization: remove Unicode characters, etc.

intgr:
* Accented characters are replaced with non-accented version (ä -> a)
* Spaces are replaced with '-' (previously they were removed)
* Multiple non-alphanumeric characters are collapsed into one '-'

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -62,7 +62,6 @@ def get_sequence(name):
 
 
 def get_or_increase_name(name):
-    name = '-'.join(name.strip().split(' '))
     certificates = Certificate.query.filter(Certificate.name.ilike('{0}%'.format(name))).all()
 
     if not certificates:
@@ -138,7 +137,7 @@ class Certificate(db.Model):
 
         # when destinations are appended they require a valid name.
         if kwargs.get('name'):
-            self.name = get_or_increase_name(kwargs['name'])
+            self.name = get_or_increase_name(defaults.text_to_slug(kwargs['name']))
         else:
             self.name = get_or_increase_name(
                 defaults.certificate_name(self.cn, self.issuer, self.not_before, self.not_after, self.san))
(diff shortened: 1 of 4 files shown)


lemur/certificates/models.py
```diff
@@ -138,7 +137,7 @@ def __init__(self, **kwargs):
 
         # when destinations are appended they require a valid name.
         if kwargs.get('name'):
-            self.name = get_or_increase_name(kwargs['name'])
+            self.name = get_or_increase_name(defaults.text_to_slug(kwargs['name']))
```
intgr: This change I wasn't sure about. On the one hand, it's a compatibility break. On the other hand, users were table to submit any arbitrary character in the 'name' field (with the sole exception of space) and it would be accepted without normalization.

lemur/certificates/models.py
```diff
@@ -62,7 +62,6 @@ def get_sequence(name):
 
 
 def get_or_increase_name(name):
-    name = '-'.join(name.strip().split(' '))
```
intgr: It seems the only purpose of this line was `name = name.replace(' ', '-')`?

kevgliss: Agreed.

lemur/certificates/models.py
kevgliss: I'm trying to think what this might break, I can't think of anything and having that kind of naming expectations could be bad. It was mainly just an attempt to make all certificate names valid AWS IAM certificate names.  I'm okay merging this in an letting it bake in our test env for a while.

intgr: My concern was that maybe some integrations have an assumption that whatever "name" they provide will be used verbatim by Lemur. It's a bad assumption already, but after this change, it's much more likely that the name changes.

kevgliss: I think in addition to this change we should outright reject the custom names that have these special characters. 

That way there are no surprises for users of the API that the name given is not returned. For auto-generated names, I don't think a consumer of an API should have any expectations of the name returns (they didn't specify).

 https://github.com/Netflix/lemur/blob/master/lemur/certificates/schemas.py#L57

PR #910, opened 2017-09-10
Sync docs requirements.txt

fbvs:
The `pyldap` package is missing causing the autodoc generation to fail (e.g [REST API docs)](http://lemur.readthedocs.io/en/latest/developer/index.html#rest-api) 

This adds the `pyldap` and updates the versions of the other packages


kevgliss:
Cheers thanks!

Issue #899, opened 2017-09-06
Issue with Database Upgrade


csmithson12345:
Hello,
I'm new to GitHub and this project, so please let me know if i have left out any useful information.

I have an issue with the upgrade process for lemur. When performing the command from the virtual environment
```
(lemur) lemur@COMPUTERNAME:/www/lemur/lemur lemur -c /home/lemur/.lemur/lemur.conf.py db upgrade
```
I get the below output. Could you please offer any suggestions for a resolution?

```
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 7f71c0cea31a -> 932525b82f1a, Changing the column name to the more accurately named 'notify'.
INFO  [alembic.runtime.migration] Running upgrade 932525b82f1a -> e3691fc396e9, Adding logging database tables.
INFO  [alembic.runtime.migration] Running upgrade e3691fc396e9 -> 131ec6accff5, Ensuring we have endpoint updated times and certificate rotation availability.
Traceback (most recent call last):
  File "/www/lemur/bin/lemur", line 11, in <module>
    load_entry_point('lemur', 'console_scripts', 'lemur')()
  File "/www/lemur/lemur/manage.py", line 545, in main
    manager.run()
  File "/www/lemur/lib/python3.5/site-packages/flask_script/__init__.py", line 412, in run
    result = self.handle(sys.argv[0], sys.argv[1:])
  File "/www/lemur/lib/python3.5/site-packages/flask_script/__init__.py", line 383, in handle
    res = handle(*args, **config)
  File "/www/lemur/lib/python3.5/site-packages/flask_script/commands.py", line 216, in __call__
    return self.run(*args, **kwargs)
  File "/www/lemur/lib/python3.5/site-packages/flask_migrate/__init__.py", line 259, in upgrade
    command.upgrade(config, revision, sql=sql, tag=tag)
  File "/www/lemur/lib/python3.5/site-packages/alembic/command.py", line 254, in upgrade
    script.run_env()
  File "/www/lemur/lib/python3.5/site-packages/alembic/script/base.py", line 425, in run_env
    util.load_python_file(self.dir, 'env.py')
  File "/www/lemur/lib/python3.5/site-packages/alembic/util/pyfiles.py", line 93, in load_python_file
    module = load_module_py(module_id, path)
[...]

PR #913, opened 2017-09-11
server_default from "False" to sa.false()

RickB17:
in support of https://github.com/Netflix/lemur/issues/899

Reference:
https://bitbucket.org/zzzeek/alembic/issues/347/default-value-doesnt-work-on-boolean


kevgliss:
lgtm

PR #917, opened 2017-09-13
Fixes 873 by explicitly declaring pyopenssl version.

caiges:


kevgliss:
@caiges Do you need this backported to the 0.5 release as well?

caiges:
It would probably be helpful if it was.

PR #932, opened 2017-09-21
Add hack to workaround cfssl ignoring csr ext'd params

BrionStone:
Here's a hack I used to workaround cfssl ignoring the csr generated by lemur.  Useful as a workaround only (don't merge!), as it assumes the 'server' -profile exists in cfssl.

### DIFF
--- a/lemur/plugins/lemur_cfssl/plugin.py
+++ b/lemur/plugins/lemur_cfssl/plugin.py
@@ -13,6 +13,8 @@ import requests
 
 from flask import current_app
 
+from cryptography import x509
+
 from lemur.plugins.bases import IssuerPlugin
 from lemur.plugins import lemur_cfssl as cfssl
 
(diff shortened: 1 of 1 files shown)


kevgliss:
Does cfssl make extensive use of profiles (Haven't used it personally)? We might need to make profiles a first class attribute in Lemur, wondering if that would help with the cfssl's integration.

BrionStone:
From what I've read, profiles are the way to instruct cfssl on what type (i.e., x509v3 Key Usage) of certs to create. The profile names are left to the implementer.   Examples are given in various places:
https://github.com/cloudflare/cfssl/blob/master/doc/cmd/cfssl.txt
with more functional ones in blogs: 
https://medium.com/@vrmvrm/setup-cloudflare-cfssl-with-ocsp-responder-aba44b4134e6
----
If  ['CA', 'email, 'ocsp', 'intermediate', 'server', 'client' ] make sense as profiles on CAs other than cfssl then sure, it should become a first class attribute.  (I'd extend the "certificate template" field to  include each of the profile types. In the lemur.conf.py file, map each profile type to the string used by the CA. On the service, move the ExtendedKeyUsage check to a function that checks for the unique attribute that designates a given profile type, adding the correct value string to the JSON.) 
However, it depends on getting a valid error message from the CA when the profile name doesn't match, and explaining how to fix it (which probably varies by CA).

kevgliss:
Closing this for now. Please reopen or create a new PR when you're ready.

PR #934, opened 2017-09-22
Updated Quikstart

aaronmell:
Got some failures doing a clean install on Ubuntu 17.04 Zesty Zapus (Final) from virtualboxes.org


kevgliss:
Thanks!

PR #936, opened 2017-09-25
Disallow issuing certificates from inactive authority

intgr:
It seems that this was the intent of the 'active' flag on authorities, but never actually implemented?

### DIFF
--- a/lemur/static/app/angular/authorities/services.js
+++ b/lemur/static/app/angular/authorities/services.js
@@ -107,13 +107,7 @@ angular.module('lemur')
     AuthorityService.findActiveAuthorityByName = function (filterValue) {
       return AuthorityApi.getList({'filter[name]': filterValue})
         .then(function (authorities) {
-          var activeAuthorities = [];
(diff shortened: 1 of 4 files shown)


lemur/static/app/angular/authorities/services.js
```diff
@@ -107,13 +107,7 @@ angular.module('lemur')
...
       return AuthorityApi.getList({'filter[name]': filterValue})
         .then(function (authorities) {
-          var activeAuthorities = [];
-          _.each(authorities, function (authority) {
-              if (authority.active) {
-                activeAuthorities.push(authority);
-              }
-          });
-          return activeAuthorities;
+          return authorities.filter(function(authority) { return authority.active; });
```
intgr: Internet Explorer 8 and older don't support Array.filter, is that OK?

kevgliss: That's fine by me. We don't make any guarantees around browser support.

kevgliss:
lgtm thanks!

PR #937, opened 2017-09-25
Avoid redundant key_view log entries

intgr:
Don't re-request private key when it's already loaded in frontend.


kevgliss:
lgtm thanks!

PR #939, opened 2017-09-25
Add option to disable owner email address in CSR subject

intgr:


lemur/certificates/service.py
```diff
@@ -344,8 +344,9 @@ def create_csr(**csr_config):
     private_key = generate_private_key(csr_config.get('key_type'))
 
     builder = x509.CertificateSigningRequestBuilder()
-    name_list = [x509.NameAttribute(x509.OID_COMMON_NAME, csr_config['common_name']),
-                 x509.NameAttribute(x509.OID_EMAIL_ADDRESS, csr_config['owner'])]
+    name_list = [x509.NameAttribute(x509.OID_COMMON_NAME, csr_config['common_name'])]
+    if current_app.config.get('LEMUR_OWNER_EMAIL_IN_SUBJECT', True):
```
intgr: Do you think this should even default to `True`?

kevgliss: Have you encountered CAs that don't allow this? Or is this purely from a privacy consideration?

intgr: It's for privacy, yes.

kevgliss: I would then leave the default as `True` and let them opt out of it. So as not to break the existing functionality.

intgr: Sure, works for me

kevgliss:
lgtm thanks!

PR #945, opened 2017-09-27
Use Cloudflare as DNS provider for LE certs

harmw:
There are some bugs in the registration process (beyond this PR) but at least the Cloudflare stuff just works 👍 

You can now register SSL certificates for a domain (DNS) hosted by Cloudflare. While they can also provide you with certs for communicating with an _origin_ or for consumers directly hitting their edge, this feature will allow one to host public LE certificates anywhere - using a domain at CF.

I'm running into various LE specific issues that make me wonder if LE itself currently even works with Lemur though.


kevgliss:
Yes, you are correct. The LE integration was very early and was not feature complete. The ACME library is still very much a moving target. Unfortunately I haven't had time to invest in fixing it.

harmw:
great, that's somewhat comforting :) I'll fix _this_ PR up first, guess I forgot to run some tests.

Let me see if I can fix the LE stuff, I'll branch *this* branch (since I'm using Cloudflare DNS) and start from there. Thanks 👍

kevgliss:
Awesome!

kevgliss:
FYI it looks like #986 has fixed some issues with ACME as well.

kevgliss:
Any update on this PR @harmw ?

harmw:
let me see 👍

kevgliss:
I'm merging this as-is, we will be looking at using the let encrypt plugin more so I'd like to reduce merge conflicts.

harmw:
super cool 👍 terribly sorry for not keeping up with this branch - the DNS API calls work though, so it wouldn't surprise me if things start rolling now :)

PR #971, opened 2017-10-12
Added ability to ignore cert for oauth2 provider

johanneslanger:
This is useful for development environments where the OAuth provider
doesn't have a valid cert!


lemur/auth/views.py
```diff
@@ -289,9 +289,10 @@ def post(self):
         # you can either discover these dynamically or simply configure them
         access_token_url = current_app.config.get('OAUTH2_ACCESS_TOKEN_URL')
         user_api_url = current_app.config.get('OAUTH2_USER_API_URL')
+        verify_cert = current_app.config.get('OAUTH2_VERIFY_CERT')
```
kevgliss: Should we default this value to "True"?

johanneslanger:
Thanks good point, added default to True. Also removed the mistakenly inserted default for OAUTH2_SECRET.

kevgliss:
lgtm thanks!

PR #985, opened 2017-11-09
Added ability to set custom roles for users logging in via oauth provider

johanneslanger:
So far it was not possible to set roles for lemur users via oauth when logging in. This pull requests adds the ability to add roles to a user on login via oauth.

The user gets assigned roles in case there is a claim "roles" in the users userinfo!


kevgliss:
This looks good to me, thanks for the PR!

PR #986, opened 2017-11-09
Fixes for using ACME with Route53

praymann:
We're looking to use Lemur heavily internally here at LAKANA with our AWS infrastructure.

We want mainly to use it with LetsEncrypt with Route53 domain validation, however, I ran into issues with the lemur_acme plugin during my proof of concept for the team.

Included in this PR are the changes I made to allow functional ACME operations with Route53.

Python is not my strongest suit, we (Ops) heavily use Ruby at LAKANA so if there are changes to be made to this please let me know!


kevgliss:
Looks good to me thanks!

Issue #987, opened 2017-11-13
Can we generate CSR for submitting ssl enrollment 


selatamil:
@kevgliss Is it possible to generate CSR in lemur WEB GUI to manually submit SSL enrollment with third party certificate providers like comodo.  If yes please let me know the procedure? 

Regards,
Tamilventhan S

kevgliss:
@selatamil this is currently not possible.

jchuong:
I have a plugin for exporting CSR already.  I can make a pull request later.

PR #988, opened 2017-11-14
CSR Export Plugin

jchuong:
This plugin allows a certificate to be exported as a CSR via OpenSSL
x509.  The workflow will be:
* Create self-signed cert via Cryptography authority
* Export CSR via this plugin
* Sign your own cert outside of Lemur
* Import new cert with private key

### DIFF
--- a/lemur/plugins/lemur_csr/__init__.py
+++ b/lemur/plugins/lemur_csr/__init__.py
@@ -1,0 +1,5 @@
+try:
+    VERSION = __import__('pkg_resources') \
+        .get_distribution(__name__).version
+except Exception as e:
+    VERSION = 'unknown'
--- a/lemur/plugins/lemur_csr/plugin.py
+++ b/lemur/plugins/lemur_csr/plugin.py
@@ -1,0 +1,104 @@
+"""
+.. module: lemur.plugins.lemur_csr.plugin
+
+An export plugin that exports CSR from a private key and certificate.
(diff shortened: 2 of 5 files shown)


jchuong:
For #987 but you have to create the self-signed first, then import the actual cert.

kevgliss:
This looks good, although I am working on the use-case for this? Generally speaking, I'd expect that Lemur be directly integrated with the CA itself. That way the user doesn't have to handle key material. Are there CA's out there that a purely manual and have no APIs at all?

jchuong:
I also prefer having Lemur request the cert from the CA itself.  There are two cases why you may want CSR however.

1.  I work at a third party company that serves traffic for a customer website (eg Akamai, Fastly, other CDNs).  A new customer wants a third party that we don't support in Lemur, because of reasons (maybe they negotiated some kind of deal, or prefer one over the other).  Then we can give the customer the CSR, and we keep the key to ourselves.  The customer gets the certificate from our CSR and we just import it back into Lemur.  

Downside is the person creating the CSR has to copy the private key into the newly imported certificate (and remove the original).  I have a change that also copies the original key, so the user doesn't have to do that (just use replaces field).

2. The CA doesn't have an API (or we just want to do a one-off with this CA, or can't create the plugin quickly enough or something).

kevgliss:
Interesting, thanks for the insight!

Issue #856, opened 2017-07-17
User associatable API keys


kevgliss:
For service applications using the Lemur API we should allow API keys to be created and allow for multiple API keys to be in use at any one time. This will help with the rotation of keys as required by the service application. 

cc @andrewachen

SecurityInsanity:
Where is this along? Would it be possible for me to take a look at this? This is something we'd love to have, but I don't want to step on anyone's toes if it's already being actively worked on.

kevgliss:
@SecurityInsanity I haven't started work on this just yet, but would love to see it included as well. Let me know if you have any questions.

SecurityInsanity:
Hey @kevgliss ,

Finally starting to take a look into this. It looks like we're already generating JWTs for the users. I'm thinking just re-using this logic for API Keys. Does this sound alright to you? Or were you thinking of something else when filing this ticket?

kevgliss:
I think JWT would work fine. Would we just make them very long lived? I think we would also need another table/column containing the list of tokens that were issued/revoked, adding in the necessary checks beyond simple signature validation.

SecurityInsanity:
Sorry for the delay on response for this. I thought I had responded, but noticed today I actually never had.

I was thinking of having a configurable timeout. Then for tracking when certain issues where issued/revoked we could simply store an id field inside the JWT, that we can use for determining if it's been revoked, etc.

kevgliss:
Sounds good to me.

SecurityInsanity:
Figured I'd post an update. I've got the backend for API Keys working + Tested. I'm now digging into the angular side of things, and should hopefully have a PR up this week.

PR #995, opened 2017-11-16
add per user api keys to the backend

SecurityInsanity:
this commit adds in per user api keys to the backend of lemur.

refs #856

the basics are:
  - API Keys are really just JWTs with custom second length TTLs.
  - API Keys are provided in the exact same ways JWTs are now.
  - API Keys can be revoked/unrevoked at any time by their creator
    as well as have their TTL Change at anytime.
  - Users can create/view/list their own API Keys at will, and
    an admin role has permission to modify all api keys in the
    instance.

### DIFF
--- a/lemur/api_keys/views.py
+++ b/lemur/api_keys/views.py
@@ -19,7 +19,8 @@ from lemur.auth.permissions import ApiKeyCreatorPermission
 from lemur.common.schema import validate_schema
 from lemur.common.utils import paginated_parser
 
-from lemur.api_keys.schemas import api_key_input_schema, api_key_revoke_input_schema, api_key_output_schema, api_keys_output_schema, api_key_described_output_schema
+from lemur.api_keys.schemas import api_key_input_schema, api_key_revoke_schema, api_key_output_schema, \
+    api_keys_output_schema, api_key_described_output_schema, user_api_key_input_schema
 
 mod = Blueprint('api_keys', __name__)
 api = Api(mod)
@@ -71,7 +72,7 @@ class ApiKeyList(AuthenticatedResource):
            :query sortDir: asc or desc
            :query page: int default is 1
            :query count: count number. default is 10
-           :query userId: a user to filter by.
+           :query user_id: a user to filter by.
            :query id: an access key to filter by.
            :reqheader Authorization: OAuth token to authenticate
            :statuscode 200: no error
@@ -130,6 +131,110 @@ class ApiKeyList(AuthenticatedResource):
         return dict(jwt=create_token(access_token.user_id, access_token.id, access_token.ttl))
 
 
+class ApiKeyUserList(AuthenticatedResource):
+    """ Defines the 'api_keys' endpoint on the 'users' endpoint. """
+    def __init__(self):
+        super(ApiKeyUserList, self).__init__()
+
+    @validate_schema(None, api_keys_output_schema)
+    def get(self, user_id):
+        """
+        .. http:get:: /users/:user_id/api_keys
+
+           The current list of api keys for a user, that you can see.
+
+           **Example request**:
+
+           .. sourcecode:: http
+
+              GET /users/1/api_keys HTTP/1.1
+              Host: example.com
+              Accept: application/json, text/javascript
+
+           **Example response**:
+
+           .. sourcecode:: http
+
+              HTTP/1.1 200 OK
+              Vary: Accept
+              Content-Type: text/javascript
+
+              {
+                "items": [
+                    {
+                      "id": 1,
+                      "name": "custom name",
+                      "user_id": 1,
+                      "ttl": -1,
+                      "issued_at": 12,
+                      "revoked": false
+                    }
+                ],
+                "total": 1
+              }
+
+           :query sortBy: field to sort on
+           :query sortDir: asc or desc
+           :query page: int default is 1
+           :query count: count number. default is 10
+           :query id: an access key to filter by.
+           :reqheader Authorization: OAuth token to authenticate
+           :statuscode 200: no error
+           :statuscode 403: unauthenticated
+        """
+        parser = paginated_parser.copy()
+        args = parser.parse_args()
+        args['has_permission'] = ApiKeyCreatorPermission().can()
+        args['requesting_user_id'] = g.current_user.id
+        args['user_id'] = user_id
+        return service.render(args)
+
+    @validate_schema(user_api_key_input_schema, api_key_output_schema)
+    def post(self, user_id, data=None):
+        """
+        .. http:post:: /users/:user_id/api_keys
+
+           Creates an API Key for a user.
+
+           **Example request**:
+
+           .. sourcecode:: http
+
+              POST /users/1/api_keys HTTP/1.1
+              Host: example.com
+              Accept: application/json, text/javascript
+
+              {
+                "name": "my custom name"
+                "ttl": -1
+              }
+
+           **Example response**:
+
+           .. sourcecode:: http
+
+              HTTP/1.1 200 OK
+              Vary: Accept
+              Content-Type: text/javascript
+
+              {
+
+                "jwt": ""
+              }
+
+           :reqheader Authorization: OAuth token to authenticate
+           :statuscode 200: no error
+           :statuscode 403: unauthenticated
+        """
+        if not ApiKeyCreatorPermission().can():
+            if user_id != g.current_user.id:
+                return dict(message="You are not authorized to create tokens for: {0}".format(user_id)), 403
+
+        access_token = service.create(name=data['name'], user_id=user_id, ttl=data['ttl'],
+            revoked=False, issued_at=int(datetime.utcnow().timestamp()))
+        return dict(jwt=create_token(access_token.user_id, access_token.id, access_token.ttl))
+
+
 class ApiKeys(AuthenticatedResource):
     def __init__(self):
         self.reqparse = reqparse.RequestParser()
@@ -174,7 +279,7 @@ class ApiKeys(AuthenticatedResource):
                 return dict(message="You are not authorized to view this token!"), 403
         return dict(jwt=create_token(access_key.user_id, access_key.id, access_key.ttl))
 
-    @validate_schema(api_key_revoke_input_schema, api_key_output_schema)
+    @validate_schema(api_key_revoke_schema, api_key_output_schema)
     def put(self, aid, data=None):
         """
         .. http:put:: /api_keys/1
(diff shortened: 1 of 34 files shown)


SecurityInsanity:
This commit adds in the backend, but I'm still having trouble understanding parts of the frontend so I haven't added that in yet.

@kevgliss I was wondering could you help me with the frontend? Specifically I'm confused how the `services.js` files work that specifically fetch data from the APIs (at least from what I think is going on). Is there perhaps documentation I'm missing? Thanks.

kevgliss:
This looks great, will leave a few comments as I go through it in more detail.

As far the UI stuff, I'd expect potentially two views. 

1) The first and easiest view is to simply have a table of all the current keys, we should probably display the creator, ttl and any other important attributes. 

    A good starting point would be to simply copy the [domain](https://github.com/Netflix/lemur/tree/master/lemur/static/app/angular/domains) service.
   
   You will probably get quite far simply renaming the endpoint to match what'd you expect. If you're still confused I would take a quick peek at these [docs](https://github.com/mgonto/restangular#lets-code) for some straightforward examples how data fetching works in Lemur (which uses Restangular).

    Each of these Javascript modules aligns very similarly how things are defined in python, there are `services` and `views` for just about every endpoint model. `services` are responsible for fetching data from the API. `views` are responsible for massaging the data and presenting it to the user. 

2) The second view would be a per-user view listing only the API keys for that are associated with that user. You could hang this endpoint off of the user endpoint.

     We do something similar with `roles` the endpoint for `/api/1/user/1/roles` returns only the roles for that account. Likewise I'd imagine the endpoint could be `/api/1/user/1/keys`. Take a look at the [user](https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/users/services.js#L23) service.

    In the the above scenario we would add a `getKeys` function to the user service `user.getList('keys')` and potentially ,`user.createKeys()`, `user.deleteKeys()`, `user.updateKeys()`. etc., depending on how far you want to go with it.

SecurityInsanity:
Hey @kevgliss ,

I've been making some process (though not to the point where I feel it's safe to spin up, and test in the UI yet). I think I've got all the UI bits in place, but I'm missing some context on where certain things are initialized.

For example take: https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/users/services.js#L6

The line currently is:
```js
.service('UserApi', function (LemurRestangular) {
```

However I've changed it too:
```js
.service('UserApi', function (LemurRestangular, ApiKeyService) {
````

How do I ensure this `UserApi` gets `ApiKeyService` populated correctly?

**Comment by kevgliss** [comment]
It's provided by angular's [dependency injection](https://docs.angularjs.org/guide/di). So as along as you define your service correctly angular will take care of it for you.

**Comment by SecurityInsanity** [comment]
Hey @kevgliss ,

So I've got the UI Started, but I'm stuck on how to show the JWT to a user once they've created an API Key in the creation modal. So far I see multiple ways to do it:

  1. Pop up a modal inside the modal. This just seems like inception in inception. I also was told by our designers I should never do this, but if you want too. It's techincally an option.
  2. Just say the UI is "management only" and if you want to get a JWT you have to make an API Call.
  3. Some other UI pattern that I'm not thinking of.

I'm not expecting an answer here til after the holidays, but wanted to drop it before I forgot. Happy Thanksgiving!

**Comment by kevgliss** [comment]
Instead of a modal-inception, why not just have a div with an "ng-show=apiKey". When apiKey===True it will show the div otherwise it will not be rendered. I would also suggest the API key be in a code block div (https://stackoverflow.com/questions/11742907/why-does-html5-recommend-putting-the-code-element-inside-pre)

**Comment by SecurityInsanity** [comment]
Hmm Where would this div kep be? If it's inside the create/edit modal window that might be a bit weird to have to re-edit the key to see that div (since the modal auto closes on creation).

Although I guess it could be just on the list page. I'll play around with that.

Thanks!

**Comment by kevgliss** [comment]
If you take a look at the export plugin I think we do something similar where a passphrase is automatically generated for you, showing the key after it's been created. I believe it just replaces the create form with the div, you can similarly do an ng-hide for those divs when apiKey===True.

**Comment by SecurityInsanity** [comment]
Thanks! I'll take a look at the export plugin.

**Comment by SecurityInsanity** [comment]
Alright @kevgliss I think this is finally ready for review (since frontend, and backend are now done)!

**Comment by kevgliss** [comment]
This looks good, but it's quite a large PR and might take me a little bit to go through. This was actually an internally request feature so I am super pumped about getting this merged.

**Comment by SecurityInsanity** [comment]
Yea no worries, I'm aware it's quite large. We're not too time crunched on our end. However, we would as you like it 😄 but it's nothing to lose sleep over for us.

File: lemur/api_keys/views.py [hunk 1]
```diff
@@ -0,0 +1,558 @@
...
+                  "issued_at": 1222222,
+                  "revoked": false
+              }
+
+           :reqheader Authorization: OAuth token to authenticate
+           :statuscode 200: no error
+           :statuscode 403: unauthenticated
+        """
+        access_key = service.get(aid)
+        if access_key is None:
```
kevgliss: I would consider reversing these two, throw a 403 if the user truly doesn't have permission then throw a 404 if the user has permission but the API key does not exist.

lemur/static/app/angular/api_keys/services.js
```diff
@@ -0,0 +1,24 @@
...
+  .service('ApiKeyApi', function (LemurRestangular) {
+    return LemurRestangular.all('api_keys');
+  })
+  .service('ApiKeyService', function ($location, ApiKeyApi) {
+    var ApiKeyService = this;
+    ApiKeyService.update = function(apiKey) {
+      return apiKey.put();
+    };
+
+    ApiKeyService.updateRevoked = function(apiKey) {
```
kevgliss: Is this needed? Is `update` enough?

lemur/static/app/index.html
```diff
@@ -53,6 +53,7 @@
                     <li><a ui-sref="authorities">Authorities</a></li>
                     <li><a ui-sref="endpoints">Endpoints</a></li>
                     <li><a ui-sref="notifications">Notifications</a></li>
+                    <li><a ui-sref="api_keys">Api Keys</a></li>
```
kevgliss: I think the listing of API keys would fall under "Admin" functionality and should be moved to the Admin dropdown.

lemur/api_keys/views.py
```diff
@@ -0,0 +1,314 @@
...
+
+class ApiKeyList(AuthenticatedResource):
+    """ Defines the 'api_keys' endpoint """
+    def __init__(self):
+        super(ApiKeyList, self).__init__()
+
+    @validate_schema(None, api_keys_output_schema)
+    def get(self):
+        """
+        .. http:get:: /api_keys
```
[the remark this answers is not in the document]
SecurityInsanity: Sure thing.

lemur/api_keys/views.py
SecurityInsanity: Hmm, that's a good idea. I just took the 403 always to not leak info (like S3, etc.), but I like throwing a 404 if they do have permission, more.

lemur/static/app/angular/api_keys/services.js
SecurityInsanity: Probably, to be honestI'm not sure why I added two here.

lemur/static/app/index.html
SecurityInsanity: I originally had that, but then I realized there'd be no way for a non-Admin user to view their API Keys. So I extracted it to the root level.

lemur/api_keys/views.py
SecurityInsanity: Actually while looking at doing this, there isn't a good way of doing this. Because we only get the access key id passed in. So we can't determine if it belongs to the user or not (and thus has permission) without the access key existing.

SecurityInsanity:
Hey @kevgliss , 

I've gone ahead an renamed `api_keys` to `keys`, and removed the needless `updateRevoked` method (since update works fine).

lemur/static/app/index.html
kevgliss: I believe all users can view other users? Additionally, I don't believe we hide the Admin dropdown from regular users.

SecurityInsanity: Oh if we don't hide the admin dropdown for regular users we can move it there. I'll go ahead, and move it there (then confirm a regular user can see it).

kevgliss:
One additional thing I noticed is that you let the user specify a userId during creation/modification, I would have assumed that the created token would be automatically tied to the user creating it. 

Did you have a use case where tokens would be moved around? It seems to me if that were the case, it would be better if new tokens would be created and the old ones revoked at a later time, otherwise if you wanted to track actions of a token back to a user you would have difficulty.

SecurityInsanity:
The userId only needs to be specified when you use the root `/keys` endpoint (because there's no way to know who the user is for). If you use the endpoint: `/users/:user_id/keys` then this will use the user id specified in the request. The reason the UI asks for this is because a use case for us (and this may only be for us) is an admin creating api keys for users. Plus shoe-horning it into the users page to get a user ID felt weird to me.

We ourselves wouldn't have a need to move an API Key from a user to a user, no. However actions would still be attributed to the user they were created for (even if they were updated at a later time). And would be revoked for everyone who had an iteration of that token (if it were ever revoked). Ideally yes people would just create new api keys, but moving api keys from user to user (keeping the same api key id) could be done, and would still allow users to be tied to themselves. (You'd just have to make sure to give them the updated JWT each time).

lemur/api_keys/views.py
kevgliss: Could we do something similar to what we do with fetching private keys? 

https://github.com/Netflix/lemur/blob/master/lemur/certificates/views.py#L434

Here we determine first if the Id exists and then if you have access. I suppose you could leak information about the aid itself, but aren't those sequentially issued anyway?

kevgliss:
Interesting, I don't think I would mind having that field on the user/edit page. One thing we may add in the future (doesn't have to be in this PR) is a typeahead for users during key creation, figuring out a user's Id isn't great.

lemur/api_keys/views.py
SecurityInsanity: Yea that could be done, but as you mentioned it would leak the access key id. However, as you also mentioned those are also issued sequentially. Technically a legitimate user should never see this message so I aired on the side of safety.

I can change it though if you'd like.

SecurityInsanity:
Yea I wouldn't mind it either, I'm just not sure how that would look I guess? Perhaps we should file an issue for adding it to that page some time in the future?

lemur/api_keys/views.py
kevgliss: Yeah, let's go ahead and make that change. I'd like to maintain the parity we have with viewing private keys.

SecurityInsanity: Alright, that change has been made. Non-existent keys will now show 404.

kevgliss:
Thanks! Will try to get this out in the 0.7 release in the next week or so!

PR #1007, opened 2017-11-29
Added revoke support to cfssl plugin

johanneslanger:
This pull requests adds support for revoking for the lemur cfssl plugin.

One thing i noticed with revoke functionality.
- When a cert is revoked there is no status indicator in Lemur UI 
- You can revoke a certificate multiple times (might be intentional?)


kevgliss:
Thanks for the PR!

We don't immediately update the revocation status, relying instead on a verification command: 

https://github.com/Netflix/lemur/blob/97d83890e05c2c31fb97d36e8e18609f2783b1d4/lemur/certificates/verify.py#L84

To give us some idea when a certificate is *actually* revoked by the CA. 

This approach was taken to externally verify when and if something had actually been revoked. OCSP updates pretty quickly but CRL often took quick awhile before revocation was actually in place. Additionally, some resource like Chrome CRLSets have a different time period for revocation (which we should start checking). 

Essentially I didn't want to give a false sense of security. 

I am open to changing this if there is a general consensus for immediately marking them as revoked in Lemur; the revocation feature was added quite recently. 

If we do go that route. We would be able to stop the user from issuing multiple revocation requests.

johanneslanger:
Thanks for the update and merging . Makes sense to me!

PR #1012, opened 2017-12-05
Adding some niceties around the way users are associated with tokens.

kevgliss:
- Includes user typeahead
- Tooltips
- User information displayed in table
- Default to current user when no user is passed

cc @SecurityInsanity


SecurityInsanity:
Oh man, this was fast! There are some great things here. Once tests pass, I think I'll check this out locally, though just a quick glance through seems fine.

PR #1015, opened 2017-12-06
Don't send notifications that are marked inactive

intgr:
Apparently previously Lemur ignored the "active" flag of notifications.


kevgliss:
lgtm thanks!

PR #1017, opened 2017-12-08
fix roles assigned in the ui for sso

### LINKED ISSUE #966: Ensure that Lemur roles are assigned correctly via Ping provider.
There are cases where the Ping provider is assigning all of the known google group roles, but Lemur specific roles are being lost during login.

SecurityInsanity:
fixes #966

this commit fixes the ability to assign roles to people in the ui
when the user is SSO. the idea is if a role is ever assigned via
SSO it becomes a "SSO Role" or a "Third Party" Role. by setting
third_party to true on the role object.

Once a role is marked as third party it can no longer be controlled
through the ui for SSO Users. (for ui users this poses no functional
change). It must be controlled via SSO.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ is 30 days. Every certificate will gain a policy regardless is auto-rotation is 
 
 Adds per-user API Keys, requires a database migration.
 
+Adds third_party to roles for external authentication roles, requires a database migration.
+
 .. note:: This version is not yet released and is under active development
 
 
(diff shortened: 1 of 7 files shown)


SecurityInsanity:
Just want to make sure we're not waiting on me to do anything here, and that this is just a "not enough hours in the day" problem. If there is something I can do please let me know.

Thanks!

kevgliss:
Nope this looks good, just need some time to do a review.

PR #1019, opened 2017-12-12
Allow sorting and filtering by camelCase field names

intgr:
The API exposes camelCase field names everywhere, but only accepted
underscore_field_names in 'filter' or 'sort' GET attributes. Now both
are allowed.


kevgliss:
lgtm thanks!

PR #1020, opened 2017-12-12
Don't write files from the test suite

intgr:
The lemur_email.tests.test_render test would fail when running unittests
from a read-only source tree.

Feel free to reject this PR if writing these files served some useful purpose for you. I could also rewrite this and write to some temporary directory. What do you think?


kevgliss:
This is fine, I believe they were only written to a file for debug purposes. (aka I could open them up in the browser easily this way)

PR #1025, opened 2017-12-15
Add DIGICERT_ORDER_TYPE to Digicert plugin

jchuong:
This allows lemur.conf.py to control which kind of certificate to
order.  User defined options are not currently supported in the the UI,
so we cannot create multiple Digicert authorities at runtime for
separate certificate types.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -581,6 +581,12 @@ The following configuration properties are required to use the Digicert issuer p
             This is the url for the Digicert API (e.g. https://www.digicert.com)
 
 
+.. data:: DIGICERT_ORDER_TYPE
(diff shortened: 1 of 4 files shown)


jchuong:
You can see valid types with the API documentation here https://www.digicert.com/services/v2/documentation/order/overview-submit

kevgliss:
Can you add what this variable does to the docs? http://lemur.readthedocs.io/en/latest/administration.html?highlight=digicert#digicert-issuer-plugin

jchuong:
@kevgliss ping, I've updated the doc for this variable.

kevgliss:
lgtm, thanks!

PR #1027, opened 2017-12-22
CRL verify: skip unknown URI schemes like ldap:// and add unit tests

intgr:


lemur/tests/test_verify.py
```diff
@@ -0,0 +1,50 @@
...
+    cert = (cert_builder
+            .add_extension(x509.CRLDistributionPoints([crl_dp]), critical=False)
+            .sign(private_key, hashes.SHA256(), default_backend()))
+
+    with mktempfile() as cert_tmp:
+        with open(cert_tmp, 'wb') as f:
+            f.write(cert.public_bytes(serialization.Encoding.PEM))
+
+        # Must not raise exception
+        crl_verify(cert_tmp)
```
intgr: Ugh, this is inconsistent currently. A cert without CRL distribution points will get status "unknown", but a cert with unsupported distribution points would be "valid".

I'll have to rethink this.

kevgliss:
lgtm thanks!

PR #1028, opened 2017-12-22
CRL verify: handle "Remove from CRL" status as not revoked

intgr:
Per RFC 5280 section 6.3.3 (k):
https://tools.ietf.org/html/rfc5280#section-6.3.3


kevgliss:
lgtm thanks!

PR #1030, opened 2018-01-02
Refactoring authentication to support GET and POST requests. 

### LINKED ISSUE #990: Allow both GET and POST for auth endpoints
Whether GET or POST is sent to our endpoints we should correctly handle the callback. 

This is particularly a problem on newer versions of FF.

kevgliss:
Closes #990.


kevgliss:
cc @nevins-b

Issue #755, opened 2017-04-21
Synchronous issuer plugin design


treacher:
Recently we had an issue where NGINX timed out a request as it was taking longer than x seconds. When this happened, it created the certificate in Digicert but we lost the private key. In generally this is not ideal as you really want to store the private key as soon as you possibly can.

I have a proposal for an idea and I want to see what your thoughts are as to whether this is something you'd accept as a PR.

**Current workflow:**
CSR is created with Private Key -> Request made via IssuerPlugin to create/accept/retrieve certificate -> Store Certificate with Private key

**Proposed workflow**
CSR is created with Private Key -> Store Certificate in database with private key and no certificate
Sync cron runs -> Updates certificate

**Pros:**
- Asynchronous workflow
- You can't lose the private key
- You can now use approval workflows with your CA

**Cons:**
- If your request to the CA fails you have a certificate with only a private key in Lemur

kevgliss:
I wouldn't have any problems merging something like this in as I think it is a very common workflow for low volume certificates, but it will take quite a bit of work. 

For historical context, most of our downstream clients and customers expect a response immediately such that they don't have to build in retry/fetching logic. We get around the asynchronous workflow by explicitly using APIs from CAs that are also synchronous (e.g. Digicert CIS). I don't know if that is an available to all customers.

I see two ways of adding this functionality:

**Workflow Option 1**
CSR is created with Private Key -> Request made via AsyncIssuerPlugin to create create/accept/retrieve certificate -> Task Status/Id is returned to caller -> Caller periodically checks for status. 

Components:
- Create a new plugin type that would be `AsyncIssuer`. All plugins of this type would use celery/rq to run a background process to request a certificate and then wait for it to be issued. 
- Create a new table to keep track of the certificate's private key and the current status of its issuance task.
- Modify certificate creation flow control, perhaps returning an empty certificate dictionary and task ID

**Workflow Option 2**
Create request -> Async? (Yes/No) -> Create background task -> return task id

Components:
- Make an explicit call to new endpoint `create/async` re-use existing plugins 
- Wrap the `create` function call in celery/rq background task.  

I'm not in love with either option, but I am currently leaning on Option 1.

pjmichael:
Another intermediate Option we were considering was going from a single Certificate commit to a 2 stage commit. 
* The initial Certificate commit would contain the private key and any other data not derived from the cert_body & cert_chain.
[...]

PR #1037, opened 2018-01-04
Async Certificate Issuing using Pending Certificates

jchuong:
For #755

I wrote this with Digicert plugin in mind, as I don't use other issuers right now, so I haven't checked if this applies to other issuers (but everything is handled by plugin, so just implement that when someone finds one they need to support).

The general work flow is as follows:

1. Create cert as normal, if you get the response immediately, then you continue as normal
2. If you don't get a certificate from the issuing plugin, then you return empty cert body, chain and external_id. 
3. Then a PendingCertificate is created (PendingCertificate pretty much a subset of Certificate, with a couple extra fields such as renaming [certs are named based on certificate, which doesn't exist for Pending Certificate, so a dummy name needs to be generated in the meanwhile).
4. Like Sources, there is a CLI command to fetch one/many/all pending certificate by id.  It goes through each pending cert, goes to issuer plugin and does `get_ordered_certificate()` and attempts to get the order via external_id.  If it does get a cert, then a real cert is uploaded.  Otherwise, the attempt counter goes up (in case we want to do something if too many failed attempts).
`python manage.py pending_certs fetch -i all` 
5. Then delete the used PendingCertificate.  Update the count of syncs for every certificate (perhaps fail and notify if a PendingCert has sat around for too long).

I have one more commit in code review on my end for the Cancel button (cancel an order that is pending), and I don't want things to diverge more so I'll add it in once I get it done.


jchuong:
One test failed, would this be the right way of doing it?
`+def test_pending_cancel(client, pending_certificate, token, status):
+    assert client.delete(api.url_for(PendingCertificates, pending_certificate_id=pending_certificate.id),`

kevgliss:
This is a big PR! It looks pretty good from a cursory glance but might take me a little while to review.  Since this shouldn't impact those that wait/are synchronous that should make it easier to merge in.

jchuong:
I think this is missing the alembic revision FYI.

kevgliss:
Did you already generate a revision? Could you add it to this PR if so?

jchuong:
I haven't generated a revision.  I can add it in a bit

jchuong:
I found a bug that doesn't copy the relationship from pending to full certificate, but I'm breaking UTs so don't merge just yet.

kevgliss:
Is this ready for me to look at?

jchuong:
Yeah, it should be fine now.  Build seems to have stalled but might as well update the branch.

kevgliss:
This looks great. Would you mind adding a blurb to http://lemur.readthedocs.io/en/latest/developer/plugins/index.html#issuer about what implimenting a `get_ordered_certificate` and `cancled_ordered_certificate` funcs mean for someone writing a new issuer plugin? And squash your commits? I plan to roll it out and monitor it internally. After this looks stable for a bit this is probably a big enough change to warrant a 0.7 release.

kevgliss:
Yup, I will squash them down. Thanks again!

PR #1041, opened 2018-01-11
Fix DigiCert issuer plugin revoke URL

jchuong:
The URL for revoking DigiCert certificates was incorrect.

See: https://www.digicert.com/services/v2/documentation/certificate/revoke


kevgliss:
lgtm thanks!

PR #1055, opened 2018-01-26
Update psycopg2 to 2.7.3.2 for glibc 2.26 compatibility

pincushionman:
Respectfully request to update psycopg2 to make it work with Ubuntu 17.10 and Archlinux (which use glibc 2.26).  
On Ubuntu 17.10, `lemur init` or `lemur db create` will fail with a backtrace, unless psycopg2 is upgraded with pip.


kevgliss:
TravisCI is happy, I'm happy. Thanks!

Issue #1045, opened 2018-01-17
Lemur should be able to revoke certs pulled in via a source plugin.


illegalhex:
Short version: can Lemur currently revoke a cert it found via a source plugin, and if not is this a feature that could be implemented? 

Currently, it seems that only certificates that are issued via an issuer plugin can be revoked despite the the revoke button being exposed in the UI for all certificates regardless of how the certificate ended up in Lemur. After digging through the code this seems to because the `authority_id` value is set as part of the certificate object when the cert is created in Lemur, however there doesn’t seem to be a way — aside from some making changes to the core Lemur code on how the certificates are imported into Lemur via a source plugin — to pass an `authority_id` when pulling the cert in via the source plugin. We would like to be able to solve for the case of someone generating a cert outside of lemur, having the cert be visible to lemur and then(if ever needed) revoked by Lemur. 

It would be very useful to allow Lemur to revoke any certificate that it knows about and could revoke if the certificate were to have been created via an issuer plugin (and thus had an authority stored), but was instead created outside of lemur and imported via a source plugin.

kevgliss:
You are correct, there is no way to do this currently. 

You're also correct about the solution we would need to modify the source plugin to actually pull in and store this information along side the certificate in question. For some issuers this can be derived (usually the sha1 of the cert) but that's not always the case. 

This should be a relatively straightforward change. I can't comment on when I might get to this... but would be more than willing to merge in a PR to support that functionality.

kevgliss:
Looking at:
https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_digicert/plugin.py#L402 

It seems like we already do what you'd like for the DigicertCISSource plugin, I'm not sure which issuer plugin you're using but you would need to figure out the `external_id` of the given certificate and store that alongside the sourced certificate.

plopezpivotal:
@kevgliss  We are using an EJBCA plugin we created for internal use. We plan on putting in a PR for that as well. From what we see here https://github.com/Netflix/lemur/blob/master/lemur/sources/service.py#L141 The authority_id is not being passed. This causes errors when you go to revoke. The plugin does pass external_id as defined. We validated this to be the case in the DB. @illegalhex has a working branch that we will send your way. 
Thanks.

kevgliss:
Ah, I see. Cool looking forward to the PR.

kevgliss:
Fixed in #1056. Thanks again.

PR #1056, opened 2018-01-26
Fix for Lemur should be able to revoke certs pulled in via a source plugin

pincushionman:
Fix for [Issue #1045](https://github.com/Netflix/lemur/issues/1045) by illegalhex (Thanks @illegalhex)


kevgliss:
lgtm thanks!

PR #1087, opened 2018-03-19
requirements.txt

castrapel:
Addition of requirements.txt and requirements.in with pinned versions for all dependencies, sub-dependencies, along with hashes


castrapel:
Initially only the production requirements. Will do the same for other requirements

PR #1096, opened 2018-03-20
lemur create_config: Non-ASCII character lemur/common/defaults.py

cjwaian:
Running the `lemur create_config` fails with the error message:
```
SyntaxError: Non-ASCII character '\xc3' in file /var/www/lemur/lemur/common/defaults.py on line 13, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

The transgressing line is actually a comment:
```
# Strip all character accents (ä => a): decompose Unicode characters and then drop combining chars.
```

Possible fix:
From PEP-0263 "...make the Python source code encoding both visible and changeable on a per-source file basis by using a special comment at the top of the file to declare the encoding."
```
# - *- coding: utf- 8 - *-
```
or
```
# coding: utf-8
```

I understand there's several ways to deal with this all subject to Lemur's design and code style.

### DIFF
--- a/lemur/common/defaults.py
+++ b/lemur/common/defaults.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 import re
 import unicodedata
 


kevgliss:
For some reason that encoding is failing quite a few tests, it's hard to say why because travis isn't printing stack traces. Perhaps it's easiest to simply remove the offending char?

cjwaian:
Hm. I've had no issues with my build but I agree that's probably the easiest solution. And, I don't think the example is necessary to perceive the intended meaning from the comment.

I'll re-submit the PR striking the offending character, if you care.

PR #1104, opened 2018-03-20
lemur create_config: Remove non-ASCII character 

cjwaian:
Running the `lemur create_config` fails with the error message:
```
SyntaxError: Non-ASCII character '\xc3' in file /var/www/lemur/lemur/common/defaults.py on line 13, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

The transgressing line is actually a comment:
```
# Strip all character accents (ä => a): decompose Unicode characters and then drop combining chars.
```

[Solution per PEP-0263](https://github.com/Netflix/lemur/pull/1096) failed encoding tests.

**Fix:** Removed special character.


kevgliss:
Whelp, it looks like it wasn't your change failing the test. Will take a look at what's going on with Travis, in the meantime I'll merge this in thanks for the PR!

PR #1125, opened 2018-03-22
add validation for port field in linuxdst plugin

### LINKED ISSUE #968: Destination Plugin: Linux Host Dst Port doesn't accept a value
Placing this here as something that is broken and needs to be looked into. I was able to reproduce it in multiple environments.

https://github.com/Netflix/lemur/blob/e43268f58504bdecc98e591dc0c7f3d8731ec897/lemur/plugins/lemur_linuxdst/plugin.py#L22-L28

image

dmitryzykov:
This pull request should fix #968 by adding validation for port field in linuxdst plugin


dmitryzykov:
but for some reason it still using previous (default?) validation pattern in WebUI for this field:
`ng-pattern="/^[0-9]{12,12}$/"`
 
screen shot 2018-03-22 at 5 46 57 pm

Should I add this validation pattern somewhere else?

kevgliss:
If you take a look at: https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/destinations/destination/destination.tpl.html#L64

It looks like `int` has that as an embedded value. Perhaps we should make that the default if there is no value for `item.validation`.

dmitryzykov:
@kevgliss, I've updated `destination.tpl.html` in 5c099b4 to use validation value from a plugin or use the default value if validation is not defined. 

Also, I've changed default expression for `int` validation
from `/^[0-9]{12,12}$/` to `^[0-9]+$'`
because old one always required 12 digits number (maybe legacy from AWS account validation)
the new expression just allow any positive number

Now it works fine for linuxdst plugin and fix #968
screen shot 2018-03-23 at 4 56 10 pm

dmitryzykov:
I'm going to create a new PR with a new plugin where all these fixes are included. 
I close this issue for now.

PR #1164, opened 2018-04-02
Fix filtering on boolean columns, broken with SQLAlchemy 1.2 upgrade

intgr:
SQLAlchemy 1.2 does not allow comparing string values to boolean
columns. This caused errors like:

    sqlalchemy.exc.StatementError: (builtins.TypeError) Not a boolean value: 'true'

For more details see http://docs.sqlalchemy.org/en/latest/changelog/migration_12.html#boolean-datatype-now-enforces-strict-true-false-none-values

But in all honesty, the filtering logic is quite ugly overall and deserves to be rewritten.
I'm not too proud of my "fix" either.

### DIFF
--- a/lemur/authorities/service.py
+++ b/lemur/authorities/service.py
@@ -9,6 +9,7 @@
 
 """
 from lemur import database
+from lemur.common.utils import truthiness
 from lemur.extensions import metrics
 from lemur.authorities.models import Authority
 from lemur.roles import service as role_service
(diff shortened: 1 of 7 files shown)


castrapel:
Thanks for the PR Marti. It looks like a test is failing, can you take a look?

intgr:
@castrapel Doh, fixed now. I had reverted one line to check that my unit test was testing the right thing and forgot to restore it. :)

castrapel:
Thanks! Looks good to me.

PR #1170, opened 2018-04-03
SFTP destination plugin

dmitryzykov:
Introducing SFTP destination plugin.

Allow the uploading of certificates to SFTP.
NGINX and Apache export formats are supported.

Password and RSA private key are supported.
_Passwords are not encrypted and stored as a plain text._

screen shot 2018-04-03 at 1 47 10 pm

Detailed logging when Lemur debug mode is enabled. 
Example from log file
```
FTP destination plugin started
Connecting to root@1.2.3.4:22
Using RSA private key
Creating /etc/nginx/certs
/etc/nginx/certs already exist, resuming
Creating /etc/nginx/certs/test.com
Uploading test.com.key to /etc/nginx/certs/test.com
Uploading test.com.pem to /etc/nginx/certs/test.com
```

This plugin makes the linuxdst plugin deprecated because SFTP plugin solves the same tasks, but it actually works, unlike linuxdst.  

This PR contains `update angular destination template to use validation value from plugin` from #1125 which fixes #968
Also fixes #1132 and fixes #969

PR #1171, opened 2018-04-04
Remove linuxdst plugin

dmitryzykov:
This PR intends to remove linuxdst plugin.

Since #1170 has been merged linuxdst isn't needed anymore. 
Linuxdst hasn't worked for a long time and is disabled by default. 
So that, I think, we can just remove it, without deprecation.


dmitryzykov:
Ok. I'll keep this PR open waiting a proper time.

Issue #1075, opened 2018-03-05
Add ECC key support to Lemur


kevgliss:
We should be able to support ECC private key generation.

Add new key generation param here:
https://github.com/Netflix/lemur/blob/2578970f7d06a7bfadf4a6c43602f784f87d7a91/lemur/common/utils.py#L77

Docs for ECC generation:
https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.ECDSA

Support curves:
```
_CURVE_TYPES = {
    "prime192v1": SECP192R1,
    "prime256v1": SECP256R1,

    "secp192r1": SECP192R1,
    "secp224r1": SECP224R1,
    "secp256r1": SECP256R1,
    "secp384r1": SECP384R1,
    "secp521r1": SECP521R1,
    "secp256k1": SECP256K1,

    "sect163k1": SECT163K1,
    "sect233k1": SECT233K1,
    "sect283k1": SECT283K1,
    "sect409k1": SECT409K1,
    "sect571k1": SECT571K1,

    "sect163r2": SECT163R2,
    "sect233r1": SECT233R1,
    "sect283r1": SECT283R1,
    "sect409r1": SECT409R1,
    "sect571r1": SECT571R1,
}
```
Must support at least `prime256v1`

Allow intermediate mapping based on key (ECC use different intermediates than RSA based certificates):
https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_digicert/plugin.py#L494

cc @andrewachen

andrewachen:
We should also add support for X25519 while we're at it.

willbengtson:
Will add X25519 at another time

castrapel:
Closing this out. x25519 is a key-exchange-only algorithm. Ed25519 is a signing algorithm. Unlike classic RSA, the signing and the key exchange are two different operations. So this won’t work.

PR #1191, opened 2018-04-10
ecc: add the support for ECC

willbengtson:
update generate_private_key to support ECC.  Move key types to constant.  Update UI for the new key types


willbengtson:
Closes #1075

PR #1206, opened 2018-04-13
Upgrade package requests

titouanc:
https://github.com/shazow/urllib3/pull/1246 has been merged

The older requests version currently in master conflicts with the one required by InfluxDB, which I use in https://github.com/titouanc/lemur-influxdb

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -14,7 +14,7 @@ certifi==2018.1.18        # via requests
 cffi==1.11.5              # via cryptography
 chardet==3.0.4            # via requests
 click==6.7                # via flask
-cookies==2.2.1            # via moto
+cookies==2.2.1            # via moto, responses
 coverage==4.5.1
 cryptography==2.2.2       # via moto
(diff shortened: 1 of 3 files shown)


kevgliss:
So it looks like it was merged, do you know if it's been released? According to: https://pypi.python.org/pypi/urllib3/1.22 they haven't released since last year 👎 

Unfortunately we have some production systems relying on mTLS support and without the fix being released we would be in a bad state.

kevgliss:
Additionally would you like for your plugin to be included on http://lemur.readthedocs.io/en/latest/administration.html#rd-party-plugins ? If so feel free to submit a PR.

titouanc:
You're right, I only paid attention to the PR itself as said in the comment. Sorry :disappointed: 

I will downgrade the influxdb requirement in my plugin instead, and open another PR here to reference the plugin in 3rd parties. Closing this one for now.

kevgliss:
Yeah, it really sucks. :-(

PR #1213, opened 2018-04-18
Show revocation status in certificates list

titouanc:
* Add a columns in the certificates list view that shows the current revocation status
* Do not show the "revoke" entry in the certificate actions menu if it is already revoked
lemur-revoke-list


castrapel:
This is a great addition but clutters the certificates view for organizations that don't use the revocation functionality often. We'd be open to adding this if you also added options on which columns to show.

titouanc:
Mmmh, this is somewhat next-level for me, as I don't have any clue on how to express this in Angular _(tbh, this is the first time I'm using Angular)_. @castrapel Maybe you have some example of such a feature (or similar) within Lemur ?

kevgliss:
Lemur uses ng-table, I believe this is an example to hiding/showing columns with that particular library. 

http://ng-table.com/#/columns/demo-visibility

We don't currently do anything similar within Lemur itself today.

intgr:
I agree that this table is already struggling with horizontal space but there's some vertical space that you can use below existing fields (highlighted). Would this be acceptable?

screen shot 2018-07-03 at 15 32 34

I also have a WIP branch that marks expired certificates with `status='expired'` which would make this more useful even for people who don't care about revocation.

castrapel:
Sorry for the delay on getting back to you on this one. I think we're okay if options are introduced but don't want to clutter the vertical view and maintain good search/filtering. I am going to close this out for now, but please feel free to revise and submit PR again if you think of a good suitable compromise with UI display options or something alternative to this.

titouanc:
@castrapel Thx for the explanation. I got stuck with Angular (it's the first time I had to dive in this framework) and eventually did not reach anything working, hence the absence of activity on this PR :)

In the end, I'm able to see relevant information via the API, so it's not that bad.

PR #1223, opened 2018-04-24
Check for default rotation policy before updating db

### LINKED ISSUE #1046: Running init more than once will create multiple 'default' rotation policies
When you run `init`, it checks for default roles, default users, etc. and skips creating them if those are found.  However, default rotation policy ('default': 30) is created and inserted into the database every time.  This causes an issue when you try to create a certificate without a rotation policy, then the default rotation policy is used (named: "default").  This commonly occurs if you use lemur-docker to test, and then only restart the web container and keep the existing lemur-postgres container with the existing data.

The fix is to either go into the DB and remove the duplicate.  We should check if 'default' already exists (name should probably be set to unique as well).

seils:
Fixes #1046 

This PR adds a check for the default rotation policy before attempting to add it through `lemur init`.


kevgliss:
lgtm thanks again!

PR #1250, opened 2018-05-23
Re-raise destination upload exceptions

jchuong:
This change re-raises exceptions caught when destination upload fails
for a certificate.  This allows the user to know that their certificate
was not uploaded to their selected destination instead of failing
silently.

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -369,10 +369,13 @@ def update_destinations(target, value, initiator):
             destination_plugin.upload(target.name, target.body, target.private_key, target.chain, value.options)
             status = SUCCESS_METRIC_STATUS
     except Exception as e:
+        current_app.logger.exception(e)
         sentry.captureException()
-
-    metrics.send('destination_upload', 'counter', 1,
-                 metric_tags={'status': status, 'certificate': target.name, 'destination': value.label})
+        # Re-raise exception, this bubbles up to the user
+        raise
+    finally:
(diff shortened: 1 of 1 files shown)


kevgliss:
Does this prevent the certificate from saving? Initially, I didn't want to implement it this way because of the disconnect between certificate create (success) and development (failure).  

This used to have ramifications for users buying certificates on a per-unit basis. With some of the newer subscription models perhaps it's no longer a big issue to simple re-create a certificate when a destination fails to upload?

jchuong:
You're right, this probably won't work as I expected.

The cert will be issued by CA but Lemur won't save it, which would be a problem for per-unit cost.  Is there a way to bubble up an issue to the UI otherwise?

kevgliss:
I tried a few different paths, but never came up with anything that worked very well. 

I think the only decent solution would be to give Lemur the ability to create background "tasks" via RQ or something similar. Then the certificate could be issued and saved, but a task could fail. 

Unfortunately, that is quite a large change (including database tables and UI/APIs for inspecting tasks) that I haven't had time to implement.

jchuong:
Hmm, well that is probably out of scope here so I will just close this.  I expect most people will hit this kind of case for misconfigured plugins (unless they are pushing certs to external destination that has gone down) so this is low priority. 

Thanks!

kevgliss:
Yeah, we typically see this when an API limit is reached. i.e. AWS IAM says we have too many certificates deployed. It's not really something the user can fix, but is confusing none the less.

PR #1251, opened 2018-05-23
Use serial_number instead of serial

jchuong:
This fixes deprecation warning coming from cryptography package about using cert.serial instead of serial_number.


kevgliss:
lgtm thanks!

PR #1254, opened 2018-05-29
Fix import metrics in notifications/messaging.py

jchuong:
`from lemur import metrics` is incorrect for notifications/messaging.py
because that is importing the `metrics` module rather than the
instanciated `lemur.extensions.metrics` object.  This will cause errors
if you import notifications/messaging.py elsewhere, since it can cause
circular dependencies.


kevgliss:
This did end up passing when forcing travis to re-build. So, lgtm thanks!

Issue #1249, opened 2018-05-21
create authority fails with nonetype


JohnTheodore:
I'm deploying lemur in my environment now to test it out that is works.

We're using oauth2 with a oauth2 identity provider written in go. So oauth2 was broken for me until I changed the lowercase b to a capital B here: https://github.com/Netflix/lemur/compare/master...JohnTheodore:master

Here if the RFC and the built in library/function for golang: RFC: https://tools.ietf.org/html/rfc7617#section-2 and golang link: https://golang.org/src/net/http/request.go?s=27803:27869#L857

I think it's supposed to be capital Basic.

So once I fixed that, oauth2 works now.

I'm having an additional problem now. So my lemur.conf.py variables that matter more are here: https://gist.github.com/JohnTheodore/bfbd98105c7aa66b44d895018b9a4e35

After I can login as the first time (with my oauth2 user having the 'admin' role attached), I believe the next step I need to take is to create an authority. When I try to create an authority I get this error in the lemur journal: https://gist.github.com/JohnTheodore/19e8803f9c50a65be1b861aa67ce5222

There are dozens of fields in the form for creating an authority. I'm not sure which part of is is causing this error: `TypeError: 'NoneType' object cannot be interpreted as an integer`

Note, once I have things working and can cut a certificate, I'll make a small PR with any fixes or docs that can make things more clear.

JohnTheodore:
Here are my kwargs from lemur/certificates/service.py(237)

https://gist.github.com/JohnTheodore/baae0d3cfd24881d5f3b88ac7af96f89

I can see 'body' is empty. When I make the authority am I also supposed to make a new cert as well?

kevgliss:
For authorities, these are typically configured via the plugin itself:
http://lemur.readthedocs.io/en/latest/administration.html#plugin-specific-options

You can typically configure which root/intermediate you need from your CA. When you create an authority for that plugin it will use these values. Some authorities create their values dynamically. Symantec/Verisign does not.

JohnTheodore:
I set all those options in my lemur.conf.py: https://gist.github.com/JohnTheodore/bfbd98105c7aa66b44d895018b9a4e35

Though there are no authorities listed when I sign in.

kevgliss:
I believe `VERISIGN_INTERMEDIATE` and `VERISIGN_ROOT` both need to be set, and be pem formatted strings. You can easily do `open('path-to-pem', 'r').readlines()` if that file if you prefer.

[...]

PR #1255, opened 2018-05-30
a few small fixes (oauth2, api example, js dependency)

JohnTheodore:
https://github.com/Netflix/lemur/issues/1249


lemur/auth/views.py
```diff
@@ -379,12 +379,10 @@ def post(self):
             verify_cert=verify_cert
         )
 
-        jwks_url = current_app.config.get('PING_JWKS_URL')
+        jwks_url = current_app.config.get('OAUTH2_JWKS_URL')
```
castrapel: Hey! Sorry for sitting on this for a while. This is a breaking change for us, and potentially for other Lemur users. Would you mind checking for both of these configuration values and issuing a warning if they are using an older value? Still allow the older value to be used for now if it is the only one configured, and we will deprecate it later.

castrapel:
We're cleaning out old PRs and it looks like this one hasn't been updated in a while. Closing this out. Please feel free to re-open if still needed.

PR #1305, opened 2018-06-09
Elasticsearch logging configuration

castrapel:


lemur/elasticsearch_logger.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+
+from elasticsearch import Elasticsearch
+from logging import getLogger, Handler, ERROR
+from pytz import timezone
+
+
+class ESHandler(Handler):
+    def __init__(self, es, index):
```
mikegrima: Make a `kwarg` for setting the log level -- default it to `ERROR`.

kevgliss:
I'd agree with @reilings that a more centralized approach to logging to ES is probably better here.

lemur/elasticsearch_logger.py
```diff
@@ -0,0 +1,35 @@
...
+        getLogger("elasticsearch").setLevel(level)
+        getLogger("elasticsearch.trace").setLevel(level)
+        getLogger("urllib3").setLevel(level)
+
+    def emit(self, record):
+        record.eventTime = datetime.datetime.now(timezone('US/Pacific')).isoformat()
+        log_entry = self.format(record)
+
+        # Push action to ES
+        return self.es.index(index=self.index, doc_type="python_log", body=log_entry)
```
mikegrima: Wrap this in `try/except` to prevent breakage on logging failure.

castrapel:
I agree with the centralized logging approach, however baseami doesn't have anything available for this purpose. We could create a paved path for something like logstash, but that would take a while longer.

lemur/elasticsearch_logger.py
kevgliss: I think the concern is that, the logic isn't necessary to have in Lemur at all. Write to a file (which is default/standard) and then let some other process figure out buffering etc., For instance we'd wouldn't want to block/delay a request simply because ES is taking a while to respond.

PR #1388, opened 2018-06-27
Cache parsed certificate instead of re-parsing for each field

intgr:
Use @cached_property decorator to cache the results of parse_certificate().

This significantly cuts down on the number of times certs need to be
parsed for a list view. Seemed like such a waste.


intgr:
Added unit test.

PR #1394, opened 2018-06-27
Expose certificate dateCreated via API

intgr:


kevgliss:
Can you update the docs for this change?

intgr:
Doh, of course, good point. I went through all docstrings for views that use `CertificateOutputSchema`. Anything else?

kevgliss:
No, that looks good to me. Thanks!

PR #1395, opened 2018-06-27
Clean up module imports

intgr:
Example:
* import lemur.common.utils -> from lemur.common import utils
* import sqlalchemy.types as types -> from sqlalchemy import types


intgr:
This can be merged now also. Discussed with kevgliss on Gitter.

castrapel:
lgtm thanks.

PR #1426, opened 2018-07-10
Sinful Use of `$`

jpartain89:
Using the `$` sign within any block of text already marked as a code block is a grievous sin due to the fact that it makes it 100% pointless for you to have USED THE CODE BLOCK IN THE FIRST PLACE!

The `$` becomes included in the text we're trying to highlight for us to be able to actually use in our own projects. 

Why post the info if you don't want us to use it. 

Thank you.


castrapel:
Thanks! lgtm

PR #1429, opened 2018-07-11
Missing file lemur/dns_providers/__init__.py

Brett-Wood:
I was receiving a ModuleNotFoundError: No module named 'lemur.dns_providers error.  I looked into it and there was not an __init__.py file within the lemur/dns_providers directory, so I added one locally.  After that everything worked as intended.


castrapel:
Looks good to me, thanks!

PR #1509, opened 2018-08-03
Fill in missing cert rotation_policy; don't ignore validation errors when re-issuing certs

intgr:
CertificateInputSchema requires the rotation_policy field, but
certificates created before the field existed have set to NULL. Thus
saving such certificates failed and probably caused other errors.

Made cert re-issuing (get_certificate_primitives) more strict so such
errors are harder to miss in the future.

### DIFF
--- a/lemur/tests/factories.py
+++ b/lemur/tests/factories.py
@@ -31,6 +31,16 @@ class BaseFactory(SQLAlchemyModelFactory):
         sqlalchemy_session = db.session
 
 
+class RotationPolicyFactory(BaseFactory):
+    """Rotation Factory."""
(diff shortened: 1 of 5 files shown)


lemur/tests/factories.py
```diff
@@ -43,6 +53,7 @@ class CertificateFactory(BaseFactory):
     description = FuzzyText(length=128)
     active = True
     date_created = FuzzyDate(date(2016, 1, 1), date(2020, 1, 1))
+    rotation_policy = SubFactory(RotationPolicyFactory)
```
intgr: It would be nicer to re-use the same 'default' policy every time, but I don't know how to make that happen.

PR #1515, opened 2018-08-03
WIP: Disallow duplicate certificates in certificate import

intgr:


intgr:
There are multiple possible approaches to this, I'm not sure anymore this is the best one.

If we want to go all-in about preventing duplicates, it would be better to add a column storing the certificate fingerprint and a unique index over it.

On the other hand, if there are some valid use cases for actually keeping duplicates in the database, maybe it's a better idea to make this validation optional; like a request parameter `"allow_duplicate": true`.

It's possible that some source plugins are broken by this if they don't properly check for pre-existing certs.

intgr:
For future reference @kevgliss had this to say about duplicate certificates (in Gitter):

> Aug 15 19:01
> @intgr Depends on what you mean by duplicate. We do have use-cases where the same certificate may be referenced by a different name. Today the base way to handle that is to simple re-import it under a different name.

So a unique database constraint is not on the table right now. I will evaluate my options.

castrapel:
Thanks for the PR @intgr, it sounds like this one is still on hold. We will close it in the interim but please feel free to submit another PR when ready.

PR #1575, opened 2018-08-17
allow null validity periods

castrapel:


kevgliss:
I assume this is because some issuers like LE don't allow you to specify validity dates? What would the user experience be like if the issuer does require dates? Are the errors clear enough for the user to realize that it's required?

PR #1576, opened 2018-08-17
force owner into lowercase. Properly decode bytecode private key

castrapel:
Force owner into lowercase whenever a cert, authority, or owner of another entity is created or updated.

### DIFF
--- a/lemur/common/schema.py
+++ b/lemur/common/schema.py
@@ -60,6 +60,8 @@ class LemurSchema(Schema):
 class LemurInputSchema(LemurSchema):
     @pre_load(pass_many=True)
     def preprocess(self, data, many):
+        if isinstance(data, dict) and data.get('owner'):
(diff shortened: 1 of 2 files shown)


lemur/common/schema.py
```diff
@@ -60,6 +60,8 @@ def wrap_with_envelope(self, data, many):
 class LemurInputSchema(LemurSchema):
     @pre_load(pass_many=True)
     def preprocess(self, data, many):
+        if isinstance(data, dict) and data.get('owner'):
+            data['owner'] = data['owner'].lower()
```
intgr: Surely this code belongs in the schemas classes that actually have the `owner` field, not the base class?

intgr:
Lowercasing the owner email breaks our Lemur integrations, which expect to get the same value back that they submitted.

What's the use case for this?

Issue #1079, opened 2018-03-06
lemur certificate check_revoked - failed to verify -  bunch of tracebacks 


arthurlogilab:
```
Failed to verify
Traceback (most recent call last):
  File "/var/www/lemur/lemur/certificates/verify.py", line 107, in verify
    return ocsp_verify(cert_path, issuer_chain_path)
  File "/var/www/lemur/lemur/certificates/verify.py", line 46, in ocsp_verify
    raise Exception("Did not receive a valid response")
Exception: Did not receive a valid response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/www/lemur/lemur/certificates/verify.py", line 110, in verify
    return crl_verify(cert_path)
  File "/var/www/lemur/lemur/certificates/verify.py", line 62, in crl_verify
    distribution_points = cert.extensions.get_extension_for_oid(x509.OID_CRL_DISTRIBUTION_POINTS).value
  File "/var/www/lemur/lib/python3.4/site-packages/cryptography/x509/extensions.py", line 103, in get_extension_for_oid
    raise ExtensionNotFound("No {0} extension was found".format(oid), oid)
cryptography.x509.extensions.ExtensionNotFound: No <ObjectIdentifier(oid=2.5.29.31, name=cRLDistributionPoints)> extension was found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/www/lemur/lemur/certificates/cli.py", line 360, in check_revoked
    status = verify_string(cert.body, cert.chain)
  File "/var/www/lemur/lemur/certificates/verify.py", line 129, in verify_string
    status = verify(cert_tmp, issuer_tmp)
  File "/var/www/lemur/lemur/certificates/verify.py", line 112, in verify
    raise Exception("Failed to verify")
Exception: Failed to verify
```
I'm not entirely sure if we're missing some configuration, or a python dependency or what.

kevgliss:
It looks like your certificate does not have a CRL in it to check, so it can't be determined if it's been revoked. Is it perhaps an internal certificate?

arthurlogilab:
Indeed. It doesn't have a CRL, maybe there could be an option to not get a traceback or is this a bug ?

kevgliss:
We could catch the exception and issue a warning instead. I'd be happy to merge a PR covering that use case.

PR #1654, opened 2018-08-31
Improved verify

explody:
This PR addresses a few issues in the CLI verification process, and #1079 

* _Stacked exceptions were yielding lots of tracebacks_
   Rather than using exceptions for non-error failure states, I've updated the code to use return values (False = revoked, True = good, None = unknown).  Exceptions were left in place where they are for unexpected errors (parse errors, http errors, etc.)

* _Verifying a large number of certificates was causing slowdowns as the code was pulling the same  CRL over and over for certs with the same CRL._
   I implemented an extremely simple cache so that an individual CRL is only downloaded once.  This works fine for the CLI where the code is loaded fresh each time but it *could* be a problem if the verify methods are used by the running web app, as the cache has no update/expire feature.  Currently, verification is only in use by the CLI code.

* _Better logging_
   Added debug logging output in case anyone needs to troubleshoot the process.  Also, parse the cert in verify() and pass it to crl_verify() and ocsp_cerify() so they both have access to cert data for logging purposes.


castrapel:
This looks great! It looks like there are some lint errors. Can you run "make lint" and "make test" (or just "make test-python") locally to ensure flake8 and the unit tests are passing?

PR #1655, opened 2018-08-31
add missing requirements files for setup

gesquive:
When packaging lemur, installs fail because [`setup.py`](https://github.com/Netflix/lemur/blob/1b77dfa47a310a3c72f9e869e7cd6d84f5658304/setup.py#L43) directly reads all of the `requirements*.txt` files that are not included as part of the package.


castrapel:
lgtm, thanks!

PR #1661, opened 2018-09-04
cfssl plugin error message typo

gesquive:
error message while creating cert is incorrect


castrapel:
Looks good to me! Will merge. Thanks

PR #1779, opened 2018-09-27
Improved verify

explody:
Redoing this completely since I broke the last one.  

This PR addresses a few issues in the CLI verification process, and #1079 

* _Stacked exceptions were yielding lots of tracebacks_
   Rather than using exceptions for non-error failure states, I've updated the code to use return values (False = revoked, True = good, None = unknown).  Exceptions were left in place where they are for unexpected errors (parse errors, http errors, etc.)

* _Verifying a large number of certificates was causing slowdowns as the code was pulling the same  CRL over and over for certs with the same CRL._
   I implemented an extremely simple cache so that an individual CRL is only downloaded once.  This works fine for the CLI where the code is loaded fresh each time but it *could* be a problem if the verify methods are used by the running web app, as the cache has no update/expire feature.  Currently, verification is only in use by the CLI code.

* _Better logging_
   Added debug logging output in case anyone needs to troubleshoot the process.  Also, parse the cert in verify() and pass it to crl_verify() and ocsp_cerify() so they both have access to cert data for logging purposes.

Tests, flake8 and docstrings have been updated since the previous PR and master is merged in.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -363,7 +363,10 @@ def check_revoked():
             else:
                 status = verify_string(cert.body, "")
 
-            cert.status = 'valid' if status else 'revoked'
+            if status is None:
+                cert.status = 'unknown'
+            else:
+                cert.status = 'valid' if status else 'revoked'
 
         except Exception as e:
             sentry.captureException()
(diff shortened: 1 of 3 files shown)


explody:
Build appears to be failing in test-cli, I don't _think_ it's related to anything in this PR.  test-python locally passes fine.

castrapel:
This appears to be an issue with lint. Try running flake8. The specific error in Travis-CI is:

--> Linting Python files
PYFLAKES_NODOCTEST=1 flake8 lemur
lemur/tests/test_verify.py:16:53: E711 comparison to None should be 'if cond is None:'
make: *** [lint-python] Error 1

Issue #1830, opened 2018-10-12
Probable issue(s) in source syncing


explody:
We weren't sure exactly which parts of this are problems vs intentional so it's mostly questions to start, but we're up front willing to provide PRs if there's agreement that some of this needs changing.

Full disclosure, we're partially responsible for some of this, via prior bug fixes and PRs, before noticing the possible problems.

I'll start here

https://github.com/Netflix/lemur/blob/b83c5eca49756401f150159a6da4b70a7a8d23c7/lemur/sources/service.py#L119-L122

This conditional looks first for a certificate matching the certificate name (presumably the cert CN).  But, given that Lemur naturally mangles the name (e.g. "CN-issuer-notBefore-notAfter-[possible hex serial if there's a name conflict]), this check will *never* match _unless_ the source itself is pre-mangling the names to the Lemur format (we did this, it was a mistake and we undid it), which is incredibly unlikely.

Proposed fix
> Either pre-mangle the certificate name so we'd be searching for a name that may exist, or scrap the check-by-name entirely.

After the name check, comes a more concerning issue

https://github.com/Netflix/lemur/blob/b83c5eca49756401f150159a6da4b70a7a8d23c7/lemur/sources/service.py#L124-L129

The get_by_serial queries are done as if serial number were unique, which even between separate Sources/Authorities which it _definitely is not_ (not unique in the DB either, whereas 'name' is).  Any given source(s) could start the serial at 1 (openssl does) and as written, `exists` could end up with certs from completely different sources, just based on overlapping serial numbers.

Under what use case would checking for existing cert legitimately return _multiple certs_?

Proposed fix
> When checking for preexisting certs, they should be checked with a combo of issuer _and_ serial number, which would be a reasonably unique combo. Or, perhaps a fingerprint field for certs so they can be universally uniquely identified.

This leads to these next snips, where the possibility of multiple `exists` is handled, but as the question above, it's confusing _why_.

Here, potentially for _multiple_ certs in  `exists`, which may be from other sources,  the `external_id` and `authority_id` are updated to match the _one_ current cert being processed in the loop.

This would overwrite fields on these certs, without even checking if they're from the same source.

Proposed fix
> IMHO, a single cert from a source should not have multiple entries in Lemur (should it?).  The "exists" checks should be sufficiently constrained that so there are not multiples - or at minimum, really confirm that the entries are all the *same* cert if there's a use case I'm missing.

[...]

PR #1865, opened 2018-10-17
Added get_by_attributes to the certificates service, for fetching cer…

explody:
…ts based on arbitrary attributes. Also associated test and extra tests for other service methods

This indirectly addresses #1830 by adding a method to the certificates service wherein one or more certs can be fetched based on any given attributes, passed as a hash of { attr_name: attr_value }.

Additionally, the source `sync_certificates()` method has been updated to look for a "search" key in the hash of each certificate returned to it from a plugin's `get_certificates()` method.  This allows any  plugin to effectively define the criteria that will be used to determine if a sync'd certificate exists yet or not.

If the `search` key exists in the certificate hash, `sync_certificates()` will use that data to call `get_by_attributes` to attempt to find the existing cert(s) that match the criteria.  If `search` is _not_ in the certificate hash, the method continues as before, so behavior for plugins expecting `get_by_name` and so on should remain unchanged.

I also snuck in additional tests for some methods in the certificates service that were not already covered.


hosseinsh:
Hi @explody, thanks for the PR. Sorry it took a bit longer.
Just merged.

Issue #1646, opened 2018-08-30
Certificates appear to use ORGANIZATION_NAME as 'issuer'


explody:
Possibly related to #1645 in that, if source import accepted these missing fields, it may be implicitly fixed.  Regardless, the 'issuer' column is set to, and remains, the org name even after update

While looking at how certificate data is initially wrong upon import from a source, we noticed that the 'issuer' field in the DB is set to a slugified string value of the cert's x509 ORGANIZATION_NAME CN.

In the UI, prior to the next sync job running, and while the cert data is incomplete, this renders the "Issuer" column as that org name slug.

After a sync job runs, `authority_id` is updated and the UI now displays the issuer name (i.e. a sub-CA CN) rather than the org name.    We presume this is related to `authority_id` because this is what the DB looks like at the two different stages:

#### After import from source, before sync runs, UI displaying the org name as issuer
```
lemur_prod=# select name,issuer,external_id,authority_id from certificates where name like 'C%';
                       name                        |       issuer       | external_id | authority_id
---------------------------------------------------+--------------------+-------------+--------------
 CertCN-OrgName-20180823-20190823                  | OrgName            |             |
 CertCN-OrgName-20190828-20190823                  | OrgName            |             |
 CertCN-OrgName-20180828-20190828                  | OrgName            |             |
 CertCN-OrgName-20180829-20190829                  | OrgName            |             |
```

#### After the sync job runs, fields are updated,  and with `authority_id` set, the UI displays the correct issuer name
```
lemur_prod=# select name,issuer,external_id,authority_id from certificates where name like 'C%';
                       name                        |       issuer       |     external_id     | authority_id
---------------------------------------------------+--------------------+---------------------+--------------
  CertCN-OrgName-20180823-20190823                 | OrgName            | XXXXXXXXXXXXXXXXX   |            4
  CertCN-OrgName-20180828-20190828                 | OrgName            | XXXXXXXXXXXXXXXXX   |            4
  CertCN-OrgName-20180829-20190829                 | OrgName            | XXXXXXXXXXXXXXXXX   |            4
  CertCN-OrgName-20180828-20190828                 | OrgName            | XXXXXXXXXXXXXXXXX   |            4
```
This seems to come from here, where the org name is preferred over the issuer CN:

https://github.com/Netflix/lemur/blob/acd473683e9e010e11966b05843477a71d7f01c5/lemur/common/defaults.py#L235

[...]

PR #1869, opened 2018-10-17
Improved source (large tests update)

explody:
I'm submitted this as a *possible* PR, but it involves several notable changes that should really be scrutinized to ensure they fit with your strategy.

**The Easy**

A bunch of Flake8/PEP8 tweaks.  TBH, they really weren't necessary but my editor kept yelling at me.

**The Less Easy**

PR #1865 is included in this PR.

Updated certificates.service.get_by_serial to take an optional argument of `issuer` such that a plugin calling it could constrain its search. As I mentioned in #1830, serial isn't guaranteed unique but serial+issuer is _almost_ guaranteed unique in the context of one source plugin.  This is somewhat redundant given `get_by_attributes` in #1865 and could possibly be replaced by that method.

Added method `valid_certificate_hostname` to certificates.validators.  Previously, `common_name()` only checked for the existence of spaces in a CN to determine if it might be a `sensitive_domain` but in testing, I ran into scenarios where that was not a guarantee. So, `valid_certificate_hostname` does a fairly standard regexp to validate that the given string is actually a valid _hostname or FQDN_, plus has support for wildcards.

Tossed in a few more tests for methods that were previously not covered, though several depend on other changes here.

**The Complicated**

We ran into problems with the test data as-in, in that all the test cert records were importing the *same* two cert bodies over and over, yielding repeating serials and CNs, yet random names, which made addressing #1830 really difficult, as there was no way to reliably query for one unique cert.

We needed a way to have more realistic certificate data - i.e. a bunch of valid but random cert bodies that would emulate certs coming from a source, and/or with control over other characteristics for deeper testing.

So, I added a "factory" (it's not really a factory_boy factory, but still...) for arbitrarily generating ceritificates from the already embedded CA cert in vectors.py.  It works like this

There's a wrapper class `SignedCertificateFactory`, with just class methods, that acts as a registry of certs, so repeating calls for the same name will not constantly generate new certs, but will return one with that name if it's already been generated.

Then there's `class SignedCertificate(object):` which is the object representing a cert itself.  Given a set of parameters, it will create a CSR and sign it using the embedded CA cert. Aside from its instance attrs, it only has two methods, `cert_pem()` and `key_pem()` which return the cert and key, PEM encoded.

I used this in places such as:

* Modified the CertificateFactory to generate body/key content from the SignedCertificate factory, so that the stack of test certs are not all identical, yielding a more realistic test scenario
*  The test source plugin, to generate *new* certs that had not been recorded yet, so as to trigger certain code to get better coverage.
* In `test_import` to import a new, not previously imported cert
* In `test_find_duplicates` to generate a new cert independent of the CertificateFactory, then import it more than once to guarantee duplicates.

Let me know what you think


castrapel:
Thanks for your commit, @expody. I'm wondering if it would make sense to add a certificate hash column in the certificates table with just a sha1 of the cert, and to compare that in addition to get_by_serial? The issue I see is that the certificate issuer for legacy issuers (as far as Lemur is concerned) may differ from the actual Certificate issuer name (as far as what is in the x509) because of https://github.com/Netflix/lemur/pull/1691 . For example, certificate issuer might be LetsEncryptAuthorityX3 but the authority name in Lemur might just be LetsEncrypt. , so get_by_serial(serial, "LetsEncryptAuthorityX3") would not pull the correct cert.

explody:
Yeah, I definitely considered that, which in large part was the approach in #1865, to let plugins define which attribute(s) constituted uniqueness for a cert - it left open the option to add a fingerprint without immediately requiring it.

It actually also brings up https://github.com/Netflix/lemur/issues/1646#issuecomment-430639525

My first thought was to add a migration that updates the Issuer name, for sake of consistency, which would indirectly address the `get_by_serial()` issue, but I'm worried that it would affect the AWS plugin if they are embedding the OrgName in the ARN.  It's somewhat separate, but related.

But regardless, I think a fingerprint is a really good idea - it transcends serial, issuer, etc and would also make get_duplicates a lot simpler/lighter.

Perhaps it could be an automatic function in the certificate model to fill an attribute with the SHA1 of `Certificate.body`?  Also, given that with the AWS plugin there are going to be repeating cert bodies, it should not be database-unique, right? For that matter, why not add both sha1 and sha256 fingerprints automatically?  I suspect that recording them in columns at save() time would be sufficiently lightweight to not impact performance - and they'd never change as long as the body didn't.

Thoughts?

castrapel:
I agree. The purpose of two fingerprints vs 1 is to prevent collisions? 

I think at the very least implementing fingerprinting now (and indexing those columns), and retroactively filling in those fingerprints for all the certs would be a good idea. If that code is checked in the source sync, we can test on our end for breaking scenarios before implementing. This type of change would also warrant a version bump.

In an ideal world, certificates in Lemur would be unique but I can guarantee that they're not for many users. As such, we may (In a different PR) want to prevent or warn users before they import duplicate certificates, and prevent lemur from importing duplicate certificates (I've seen this issue crop up periodically).

castrapel:
Hi @explody , we really appreciate this PR as it highlighted issues of syncing certificates incorrectly due to non-unique serial numbers. We've recently addressed the source syncing issue in another PR (We still rely on serial number to find possible certificate matches, but then we take an MD5 of the cert body and compare it to the MD5 of those matches). https://github.com/Netflix/lemur/pull/2360/files . Your PR has a lot of useful elements that we would like to use including the tests and lint fixes. I'm going to close this for now, but would you mind please submitting any relevant fixes in a future PR? There are fixes in your PR that we've missed as well, and we'd like to ask that any changes also be submitted in a future PR. It's easier for us to review smaller PRs due to the complexity of our environment and our Lemur usage. (We can better test for breaking changes that way).

Thanks so much!

PR #1909, opened 2018-10-24
Add CSR to certificiates

jchuong:
Add csr column to certificates field, as pending certificates have
exposed the CSR already.  This is required as generating CSR from
existing certificate is will not include SANs due to OpenSSL bug:
https://github.com/openssl/openssl/issues/6481

### DIFF
--- a/lemur/tests/test_certificates.py
+++ b/lemur/tests/test_certificates.py
@@ -48,7 +48,7 @@ def test_get_certificate_primitives(certificate):
 
     with freeze_time(datetime.date(year=2016, month=10, day=30)):
         primitives = get_certificate_primitives(certificate)
-        assert len(primitives) == 25
+        assert len(primitives) == 26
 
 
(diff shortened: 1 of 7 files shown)


jchuong:
Related to #987, since we can have `cryptography` generate a self-signed cert + private key and CSR, then create cert with a third-party CA, import and replace the original self-signed cert with the new cert.  The CSR exporting plugin I wrote is flawed because OpenSSL drops extensions, so you would need to read the SANs, and re-add them to the CSR.

lemur/tests/test_certificates.py
```diff
@@ -48,7 +48,7 @@ def test_get_certificate_primitives(certificate):
 
     with freeze_time(datetime.date(year=2016, month=10, day=30)):
         primitives = get_certificate_primitives(certificate)
-        assert len(primitives) == 25
+        assert len(primitives) == 26
```
jchuong: I wasn't entirely sure if CSR would be considered a primitive here.

Issue #1790, opened 2018-10-02
Kubernetes plugin question


aaronmell:
I added a certificate and a Kubernetes endpoint. The certificate doesn't show up in Kubernetes and I can't seem to find any logs why.  Where can I find information about why it failed to place the certs in kubernetes? I'm running the app as docker containers and attempting to load the cert into EKS

kevgliss:
Are you using the UI or API? Is there anything in `lemur.log`?

kevgliss:
cc @mik373

wfhartford:
I am having the same experience. Lemur is running in a Kubernetes cluster, trying to deploy to the same cluster. The SFTP destination is working, but I get no feedback at all from the Kuberentes destination. There are no messages about it in the pod logs or the lemur.log file in the container. I'm interacting with lemur through the UI.

wfhartford:
I took a closer look at the Kubernetes plugin code and there are some pretty serious issues with it, mostly related to incompatible types. I'm not a python expert, but my guess is that the plugin worked in python 2 but fails in python 3. I've made a few changes that make the plugin work. They're pretty ugly at this stage, but I'll clean up my fixes and submit a PR.

There are a few functional changes I would like to make as well. Would a separate PR which makes the following changes be accepted?
* Allow configuration of Kubernetes certificate and token as file paths, better supporting lemur running inside Kubernetes.
* Default the Kubernetes namespace to the namespace Lemur is running in if it is running in Kubernetes.
* Configurable deployment of private keys so that the plugin only deploys certificates by default.

wfhartford:
@aaronmell The problem with the Kubernetes plugin should be fixed by #2219 and further improved by #2229. If you still have any interest in Lemur & Kubernetes, I would love any feedback you have.

PR #2219, opened 2018-12-10
 Fix Kubernetes Destination Plugin 

wfhartford:
The Kubernetes plugin was broken. There were two major issues:
* The server certificate was entered in a string input making it impossible (as far as I know) to enter a valid PEM certificate.
* The base64 encoding calls were passing strings where bytes were expected.

The fix to the first issue depends on #2218 and a change in the options structure. I've also included some improved input validation and logging.

### DIFF
--- a/lemur/plugins/lemur_kubernetes/plugin.py
+++ b/lemur/plugins/lemur_kubernetes/plugin.py
@@ -35,14 +35,8 @@ def ensure_resource(k8s_api, k8s_base_uri, namespace, kind, name, data):
 
     if 200 <= create_resp.status_code <= 299:
         return None
-
-    else:
-        json = create_resp.json()
-        if 'reason' in json:
-            if json['reason'] != 'AlreadyExists':
-                return create_resp.content
-        else:
-            return create_resp.content
+    elif create_resp.json().get('reason', '') != 'AlreadyExists':
+        return create_resp.content
 
     url = _resolve_uri(k8s_base_uri, namespace, kind, name)
     current_app.logger.debug("K8S PUT request URL: %s", url)
@@ -53,7 +47,7 @@ def ensure_resource(k8s_api, k8s_base_uri, namespace, kind, name, data):
     if not 200 <= update_resp.status_code <= 299:
         return update_resp.content
 
-    return None
+    return
 
 
 def _resolve_ns(k8s_base_uri, namespace, api_ver=DEFAULT_API_VERSION,):
@@ -152,8 +146,8 @@ class KubernetesDestinationPlugin(DestinationPlugin):
                 }
             })
         except Exception as e:
-            current_app.logger.exception("Exception in upload")
-            raise e
+            current_app.logger.exception("Exception in upload: {}".format(e), exc_info=True)
+            raise
 
         if err is not None:
             current_app.logger.debug("Error deploying resource: %s", err)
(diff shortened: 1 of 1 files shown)


wfhartford:
This should address  #1790. As I commented there, I'll soon submit another PR with a couple minor improvements.

Fair Warning: I'm not much of a python programmer. I've probably done things that aren't very pythonic. Feel free to let me know.

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -26,21 +28,32 @@ def ensure_resource(k8s_api, k8s_base_uri, namespace, kind, name, data):
...
 
     create_resp = k8s_api.post(url, json=data)
+    current_app.logger.debug("K8S POST response: %s", create_resp)
 
     if 200 <= create_resp.status_code <= 299:
         return None
 
-    elif create_resp.json()['reason'] != 'AlreadyExists':
-        return create_resp.content
+    else:
```
castrapel: This whole blurb  from L39-L45 is a bit wordy. We can make it safer by just changing the original lines to the following:

`elif create_resp.json().get('reason', '') != 'AlreadyExists':
        return create_resp.content`

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -26,21 +28,32 @@ def ensure_resource(k8s_api, k8s_base_uri, namespace, kind, name, data):
...
 
-    update_resp = k8s_api.put(_resolve_uri(k8s_base_uri, namespace, kind, name), json=data)
+    update_resp = k8s_api.put(url, json=data)
+    current_app.logger.debug("K8S PUT response: %s", update_resp)
 
     if not 200 <= update_resp.status_code <= 299:
         return update_resp.content
 
-    return
+    return None
```
castrapel: `return` and `return None` are essentially the same thing. It's more Pythonic to just use return in most cases. even without any `return`, None will be returned, but explicitly having just `return` is fine.

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -106,33 +125,38 @@ def __init__(self, *args, **kwargs):
...
+                },
+                'data': {
+                    'combined.pem': base64encode('%s\n%s' % (body, private_key)),
+                    'ca.crt': base64encode(cert_chain),
+                    'service.key': base64encode(private_key),
+                    'service.crt': base64encode(body),
+                }
+            })
+        except Exception as e:
+            current_app.logger.exception("Exception in upload")
```
castrapel: if you add a parameter  here, the log can contain traceback info specifying specific details about the exception

current_app.logger.exception("Exception in upload: {}".format(e), exc_info=True)

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -106,33 +125,38 @@ def __init__(self, *args, **kwargs):
...
+            raise e
```
castrapel: You can just say `raise` instead of `raise e`

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: Excellent, I knew there would be a better way.

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: Thanks

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: Cool, thanks

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: Excelent

castrapel:
LGTM! Thanks for the help here.

PR #2229, opened 2018-12-12
 Improve the Kubernetes Destination plugin

wfhartford:
This pull request depends on my two previous PRs: #2218 and #2219

The plugin now supports loading details from local files rather than requiring them to be entered through the UI. This is especially relaent when Lemur is deployed on Kubernetes as the certificate, token, and current namespace will be injected into the pod. The location these details are injected are the defaults if no configuration details are supplied.

The plugin now supports deploying the secret in three different formats:
* Full - matches the formate used by the plugin prior to these changes.
* TLS - creates a secret of type `kubernetes.io/tls` and includes the certificate chain and private key, this format is used by many Kubernetes features.
* Certificate - creates a secret containing only the certificate chain, suitable for use as trust authority where private keys should _NOT_ be deployed.

The deployed secret can now have a name set through the configuration options; the setting allows the insertion of the placeholder '{common_name}' which will be replaced by the certificate's common name value.

Debug level logging has been added.

### DIFF
--- a/lemur/plugins/lemur_kubernetes/plugin.py
+++ b/lemur/plugins/lemur_kubernetes/plugin.py
@@ -73,6 +73,7 @@ def _resolve_uri(k8s_base_uri, namespace, kind, name=None, api_ver=DEFAULT_API_V
 def base64encode(string):
     return base64.b64encode(string.encode()).decode()
 
+
 def build_secret(secret_format, secret_name, body, private_key, cert_chain):
     secret = {
         'apiVersion': 'v1',
@@ -100,6 +101,7 @@ def build_secret(secret_format, secret_name, body, private_key, cert_chain):
             'tls.crt': base64encode(cert_chain),
         }
     return secret
+
 
 class KubernetesDestinationPlugin(DestinationPlugin):
     title = 'Kubernetes'
--- a/lemur/plugins/lemur_kubernetes/plugin.py
+++ b/lemur/plugins/lemur_kubernetes/plugin.py
@@ -25,7 +25,6 @@ DEFAULT_API_VERSION = 'v1'
 
 
 def ensure_resource(k8s_api, k8s_base_uri, namespace, kind, name, data):
-
     # _resolve_uri(k8s_base_uri, namespace, kind, name, api_ver=DEFAULT_API_VERSION)
     url = _resolve_uri(k8s_base_uri, namespace, kind)
     current_app.logger.debug("K8S POST request URL: %s", url)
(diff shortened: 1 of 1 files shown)


wfhartford:
As mentioned in my previous PRs, I'm not a seasoned python developer, and I would appreciate any feedback. This collection of PRs represents most of the python I've worked with.

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -70,92 +118,179 @@ class KubernetesDestinationPlugin(DestinationPlugin):
...
-            'data': {
-                'combined.pem': base64.b64encode(body + private_key),
-                'ca.crt': base64.b64encode(cert_chain),
-                'service.key': base64.b64encode(private_key),
-                'service.crt': base64.b64encode(body),
-            }
-        })
+        try:
+            k8_base_uri = self.get_option('kubernetesURL', options)
+            secret_format = self.get_option('secretFormat', options)
```
castrapel: nit: Most of these can be single-spaced

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -70,92 +118,179 @@ class KubernetesDestinationPlugin(DestinationPlugin):
...
+                k8s_api,
+                k8s_base_uri=k8_base_uri,
+                namespace=self.k8s_namespace(options),
+                kind="secret",
+                name=secret_name,
+                data=secret
+            )
+
+        except Exception as e:
+            current_app.logger.exception("Exception in upload")
```
castrapel: Same feedback here from my review of your other PR: Add "exc_info=True" and just "raise" instead of "raise e"

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -70,92 +118,179 @@ class KubernetesDestinationPlugin(DestinationPlugin):
...
+            raise e
 
         if err is not None:
+            current_app.logger.debug("Error deploying resource: %s", err)
```
castrapel: current_app.logger.error instead of current_app.logger.debug

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -70,92 +118,179 @@ class KubernetesDestinationPlugin(DestinationPlugin):
...
+    def k8s_bearer(self, options):
+        bearer = self.get_option('kubernetesAuthToken', options)
+        if not bearer:
+            bearer_file = self.get_option('kubernetesAuthTokenFile', options)
+            with open(bearer_file, "r") as file:
+                bearer = file.readline()
+            if bearer:
+                current_app.logger.debug("Using token read from %s", bearer_file)
+            else:
+                raise Exception("Unable to locate token in options or from %s", bearer_file)
```
castrapel: Should we log if this exception is raised?

lemur/plugins/lemur_kubernetes/plugin.py
```diff
@@ -70,92 +118,179 @@ class KubernetesDestinationPlugin(DestinationPlugin):
...
+    def k8s_namespace(self, options):
+        namespace = self.get_option('kubernetesNamespace', options)
+        if not namespace:
+            namespace_file = self.get_option('kubernetesNamespaceFile', options)
+            with open(namespace_file, "r") as file:
+                namespace = file.readline()
+            if namespace:
+                current_app.logger.debug("Using namespace %s from %s", namespace, namespace_file)
+            else:
+                raise Exception("Unable to locate namespace in options or from %s", namespace_file)
```
castrapel: Should we log if this exception is raised?

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: That exception gets logged in the exception handler in upload. Would you like me to add a log entry here?

lemur/plugins/lemur_kubernetes/plugin.py
wfhartford: That exception gets logged in the exception handler in upload. Would you like me to add a log entry here?

PR #2261, opened 2018-12-21
Properly handle Unicode in issuer name sanitization

intgr:
If the point of sanitization is to get rid of all non-alphanumeric
characters then Unicode characters should probably be forbidden too.

We can re-use the same sanitization function as used for cert 'name'.


castrapel:
Looks good to me, thanks!

PR #1514, opened 2018-08-03
WIP: Check that uploaded chain and private key match certificate

intgr:
This is done in the upload validator as well as Certificate constructor,
to make sure that no code path can persist invalid certificates.

PR #2269, opened 2018-12-26
Check that stored private keys match certificates

intgr:
I split this out from #1514, to reduce its scope and make the review/merge easier.

Verification is done in two places:
* Certificate import validator -- throws validation errors.
* Certificate model constructor -- to ensure integrity of Lemur's data even when issuer plugins or other code paths have bugs.


intgr:
OK, this should now finally be ready for merging. :)

intgr:
Added several tests for `CertificateUploadInputSchema` logic

hosseinsh:
Thanks Marti for the PR! Just Merged

PR #2289, opened 2019-01-03
Implement delete cert API call

rmoesbergen:
Implement certificate delete call in the API. It only marks the certificate as 'deleted' in the database and only succeeds when the certificate is expired.

Please note that builds are currently failing due to an invalid 'kombu' dep on version 4.2.2 which for some reason no longer exists. I ran tests with 4.2.1 and they were successful.


rmoesbergen:
I'll re-create this PR against a clean tree, it's become a bit of a mess.

PR #2311, opened 2019-01-08
Fix: Update kombu dep to 4.2.2.post1

ExaneServerTeam:
kombu 4.2.2 is no more available on pypi and thus
breaks the current install.
It is now required to depend on 4.2.2.post1


castrapel:
Hi  @ExaneServerTeam , thanks for the PR! It looks like we're both trying to fix this issue at the same time. I've submitted a PR that changes the way we look at / update the requirements (It is here: https://github.com/Netflix/lemur/pull/2312). I'll close this for now because of that. We really appreciate the help.

ExaneServerTeam:
Sure thanks for the heads up !

PR #2262, opened 2018-12-21
Use special issuer values <selfsigned> and <unknown> in special cases

intgr:
This way it's easy to find/distinguish selfsigned certificates stored in Lemur.

PR #2336, opened 2019-01-11
Check that stored certificate chain matches certificate

intgr:
Similar to how the private key is checked in #2269.

The second and final part of what was originally #1514.

### DIFF
--- a/lemur/common/validators.py
+++ b/lemur/common/validators.py
@@ -1,27 +1,14 @@
 import re
 
 from cryptography import x509
+from cryptography.exceptions import UnsupportedAlgorithm, InvalidSignature
 from cryptography.hazmat.backends import default_backend
 from cryptography.x509 import NameOID
 from flask import current_app
 from marshmallow.exceptions import ValidationError
 
 from lemur.auth.permissions import SensitiveDomainPermission
-from lemur.common.utils import parse_certificate, is_weekend
-
-
-def public_certificate(body):
-    """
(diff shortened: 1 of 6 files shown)


lemur/common/validators.py
```diff
@@ -138,3 +125,35 @@ def verify_private_key_match(key, cert, error_class=ValidationError):
     """
     if key.public_key().public_numbers() != cert.public_key().public_numbers():
         raise error_class("Private key does not match certificate.")
+
+
+def verify_cert_chain(certs, error_class=ValidationError):
+    """
+    Verifies that the certificates in the chain are correct.
```
intgr: This is exactly the same function as used in #2262. It doesn't matter which one gets merged first.

intgr:
Ping?

lemur/tests/factories.py
```diff
@@ -140,6 +140,7 @@ class CACertificateFactory(CertificateFactory):
 class InvalidCertificateFactory(CertificateFactory):
     body = INVALID_CERT_STR
     private_key = ''
+    chain = ''
```
intgr: Needed to override this, otherwise it would cause `AssertionError` in `Certificate.check_integrity`

hosseinsh:
This looks good to us, there is just a conflict to resolve.

intgr:
Cool. Rebased & resolved conflicts.

hosseinsh:
Thanks Marti for the PR. 
Just merged it.

PR #2360, opened 2019-01-14
Compare certificate hashes 

castrapel:
Compare certificate hashes to determine if Lemur already has a synced certificate. Serial number alone is not a reliable method to determine if we have a matching certificate.


intgr:
It seems a less hacky version would be to add a fingerprint column.

PR #2388, opened 2019-01-21
Implemente certificates 'delete' API call

rmoesbergen:
This PR implements the 'delete' operation of the 'certificates' API endpoint. It will mark a certificate as deleted in the database and create an audit entry. This is only succeeds when the certificate is expired.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -6,6 +6,7 @@
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
 import base64
+import arrow
 from builtins import str
 
 from flask import Blueprint, make_response, jsonify, g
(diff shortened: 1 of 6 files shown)


rmoesbergen:
Hi Hossein, thanks for merging this PR.
To answer your question: I work for a hosting company and we provide SSL certificates for our customers as and added service. This means that we have thousands of certificates to manage, that expire every year, or even 3 months when using let's encrypt. This leaves a lot of expired certificates which to us do not have any value, but do make searching / working with the certificates more difficult. This is why a delete option is very useful to us.

Currently, setting the deleted flag does nothing, the certificates still show up in the UI and API lists. I plan to submit a PR to change that (the API would, by default, not return deleted certificates). Would you consider that acceptable?

hosseinsh:
Thanks Ronald for the clarification.

Currently we are not planning to give the users the possibility of "deleting" certs, as once issued they will be around and we would like to keep track of them, as in the methodology of the Certificate Transparency.

I agree with you, for house-keeping it is a good idea to have expired certs marked as deleted to exclude them from search and keep the UI light. However, we have to find an elegant way in the UI to expand the search to include/exclude expired certs (marked as deleted).

rmoesbergen:
Okay, I will try and see if I can make that UI adjustment. I'm thinking of a 'show deleted certs' slider in the 'filter' bar that defaults to 'off'.

hosseinsh:
that would be awesome.

rmoesbergen:
Created as #2466

PR #2396, opened 2019-01-23
Expose full certificate RFC 4514 Distinguished Name string

intgr:
Using rfc4514_string() method added in cryptography version 2.5.


castrapel:
Thanks for the PR! Just merged it in.

PR #2466, opened 2019-02-05
Implement 'Show deleted certificates' toggle button

rmoesbergen:
This PR implements a 'Show deleted certificates' toggle button in the filter bar of the 'Certificates' overview. It defaults to off, so certificates marked as deleted in the database are not shown by default anymore. This is a continuation of PR #2289, as discussed with @hosseinsh.

The API is not affected and will still return all certificates, including deleted ones. I did however add the possibility to filter on the deleted flag by passing ?filter=deleted;false. The UI does not use this however, because filter only understands one key;value pair, and it would mess up the regular UI table filtering.

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -101,7 +101,7 @@ class Certificate(db.Model):
     issuer = Column(String(128))
     serial = Column(String(128))
     cn = Column(String(128))
-    deleted = Column(Boolean, index=True)
+    deleted = Column(Boolean, index=True, default=False)
     dns_provider_id = Column(Integer(), ForeignKey('dns_providers.id', ondelete='CASCADE'), nullable=True)
 
     not_before = Column(ArrowType)
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -325,6 +325,8 @@ def render(args):
             query = query.filter(Certificate.notify == truthiness(terms[1]))
         elif 'active' in filt:
             query = query.filter(Certificate.active == truthiness(terms[1]))
+        elif 'deleted' in filt:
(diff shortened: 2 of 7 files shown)


hosseinsh:
Hi Ronald,
thanks for the PR and the quick turn-around about this. Is there a way we can make this a configurable option, to have the toggle visible or not?
Since we don't have the delete cert possibility, we won't be able to show it and I assume there might be others who also don't have this use-case.
Hence, it would be great if this option can be configurable. 
Thanks!

rmoesbergen:
Hi Hossein,

I have tried to implement this, but I can't find a clean way to get a configuration variable into the frontend without all kinds of nasty hacks. Also, from what I can see, there's no other place that uses configuration settings in the frontend. So I'm a bit stuck here...

However, I would like to propose a different solution: What if we drop the front-end toggle button, and introduce a setting called ALLOW_CERT_DELETION which defaults to False. If set to True, the backend will honor the 'deleted' flag and return only certs not marked as deleted, if set to False (default), it will behave just like it does now. Our use-case has no need for the toggle button (we never want to see deleted certs), your use-case has no need for it either (you always want to see deleted certs). This would make it a backend-only solution which is much simpler and more consistent between API en UI results.

hosseinsh:
Hi Ronald,
that sounds like a good solution.
Though I like to have a toggle button for "show expired certificates" in the future. This would be different than the delete option and we would need to expand the database with a binary expired column to track expired certs. Ideally, the column should derive its state from not_after < current_date. But for the delete option, I would go with the  ALLOW_CERT_DELETION option. 
Thanks!

rmoesbergen:
I will create a new PR for this, closing this one.

PR #2468, opened 2019-02-05
Celery job for synching destinations and sources

hosseinsh:
Creating a new celery task to sync sources with destinations. This is as a measure to make sure important new destinations are also present as sources.

Only certain type of destinations qualify for us as sources.


lemur/common/celery.py
```diff
@@ -255,3 +258,37 @@ def sync_source(source):
...
+    """
+    current_app.logger.debug("Syncing source and destination")
+
+    # a set of all accounts numbers available as sources
+    src_accounts = set()
+    sources = validate_sources("all")
+    for src in sources:
+        src_accounts.add(IPlugin.get_option('accountNumber', src.options))
+
+    for dst in destinations_service.get_all():
```
castrapel: Do we want to add a restriction somewhere  here to only check aws destinations? your call

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -266,6 +225,49 @@ def clean(self, certificate, options, **kwargs):
         iam.delete_cert(certificate.name, account_number=account_number)
 
 
+class AWSDestinationPlugin(DestinationPlugin):
```
castrapel: Why copy and paste in this whole section from further up to down here? It makes it difficult to spot the differences

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -266,6 +225,49 @@ def clean(self, certificate, options, **kwargs):
...
+        },
+        {
+            'name': 'path',
+            'type': 'str',
+            'default': '/',
+            'helpMessage': 'Path to upload certificate.'
+        }
+    ]
+
+    # 'elb': {
```
castrapel: I know you didn't add this but feel free to remove this commented section

lemur/common/celery.py
hosseinsh: yeap, we do this via `destination_plugin.sync_as_source`, since we wanted to avoid hard-coding " aws destinations"

lemur/plugins/lemur_aws/plugin.py
hosseinsh: good point, put there is not changes, except
sync_as_source_name = AWSSourcePlugin.slug

which is the reason why moving it down.

lemur/plugins/lemur_aws/plugin.py
hosseinsh: ok

PR #2469, opened 2019-02-05
Source and Destination registration

hosseinsh:
While adding a new destination, we also add it as source, in case relevant for our source table


lemur/destinations/service.py
```diff
@@ -28,6 +30,13 @@ def create(label, plugin_name, options, description=None):
             del option['value']['plugin_object']
 
     destination = Destination(label=label, options=options, plugin_name=plugin_name, description=description)
+    current_app.logger.info("Destination: %s created", label)
+
+    # add the destination as source, to avoid new destinations that are not in source, as long as an AWS destination
+    if plugin_name == 'aws-destination':
```
kevgliss: Instead of hardcoding this, is there a way for use to allow a plugin to self-declare that it meets these requirements? Perhaps some metadata we can inspect?

kevgliss: Today, we define `type` of either source/destination. Do we need a new type of `source+destination` we could key this logic off of?

https://github.com/Netflix/lemur/blob/master/lemur/plugins/bases/source.py#L13

hosseinsh: Good point Kevin, now a Destination-plugin can define if it qualifies as source. Also adopted this approach for the celery job that syncs destinations and sources based on this qualification. 
Thanks.

PR #2529, opened 2019-02-14
Implement ALLOW_CERT_DELETION setting

rmoesbergen:
Implement a ALLOW_CERT_DELETION option (boolean, default False). When enabled, the certificate delete API call will work and the UI will no longer display deleted certificates. When disabled (the default), the delete API call will not work (405 method not allowed) and the UI will show all certificates, regardless of the 'deleted' flag.

Followup for PR #2466


lemur/certificates/views.py
```diff
@@ -678,17 +677,21 @@ def delete(self, certificate_id, data=None):
...
+           :statuscode 405: certificate deletion is disabled
 
         """
+        if not current_app.config.get('ALLOW_CERT_DELETION', False):
+            return dict(message="Certificate deletion is disabled"), 405
+
         cert = service.get(certificate_id)
 
-        if not cert:
+        if not cert or cert.deleted:
```
castrapel: Can you make a separate conditional? Something like:

`if cert.deleted:
    return dict(message="Certificate is already deleted"), 412`

rmoesbergen: Done!

rmoesbergen:
The requested changes are done, but I can't seem to get the 'changes requested' status to go away..

hosseinsh:
Thanks Ronald for the PR, and addressing the requested changes.
Just merged it.

PR #2534, opened 2019-02-15
Allow uploading csr along with certificates

jplana:
This small patch allows uploading a CSR along with the PEM, KEY, and CHAIN for those certificates not managed directly by Lemur. With this, you can maintain the same tracking level as the certificates managed by Lemur, and reuse the CSR.


castrapel:
Looks like one of the unit tests is failing:

```______________________ test_certificate_upload_schema_ok _______________________
  File "/home/travis/build/Netflix/lemur/lemur/tests/test_certificates.py", line 463, in test_certificate_upload_schema_ok
    assert not errors
AssertionError: assert not {'csr': ['Domain ACommonName does not match whitelisted domain patterns. Contact an administrator to issue the certificate.']}
```

Can you try to address this? Running `make test-python` locally should run the unit test suite and allow you to replicate the error / test for fixes.

jplana:
I've added a new CSR to the vectors files, matching with the actual certificate being uploaded, and updated the branch to master.

castrapel:
Tested  uploading cert without and with CSR. Looks great! Thanks for the PR!

PR #2549, opened 2019-02-19
updated requirements

hosseinsh:


hosseinsh:
why is coverage/coveralls taking for ever?

PR #2563, opened 2019-02-21
Fix setting alias to "blah" when exporting without explicit alias

rmoesbergen:
When no alias is entered when exporting a certificate, the alias is set to 'blah'.
This fix sets it to the common name instead.


castrapel:
Looks good to me! Just merged it in. Thanks.

PR #2617, opened 2019-03-01
Hashi Vault Destination Plugin

alwaysjolley:
Adding plugin to upload certificates to a Hashi Vault server, supports different formats and multiple certificates in the same object as key/value pairs.


castrapel:
Syntax LGTM, plugin itself is untested as we don't use Vault. I've merged this in, please add PRs to support this if needed.,

PR #2700, opened 2019-03-15
Parse DNSNames from CSR into Lemur Certificate

jramosf:
When user inputs a CSR for creating a new certificate, if it contains any DNSName in the SAN they will be parsed and automatically added into the "DOMAINS" section.


castrapel:
LGTM, thanks for the commit

PR #2747, opened 2019-03-26
Parse SubjectAlternativeNames from CSR into Lemur Certificate

jramosf:
@castrapel I have adapted it for working with any.

Since cryptography.x509 classes do correspond to Lemur's API, the generalization is quite straightforward


hosseinsh:
Javier, it would be really great, if you could also add automated testing for parsing SANs. This would make the integration easier.
Thanks

jramosf:
> Javier, it would be really great, if you could also add automated testing for parsing SANs. This would make the integration easier.
> Thanks

Hi @hosseinsh, I've added `test_certificate_input_schema_parse_csr` to cover the new functionality.

Since we don't supply complete dicts from the tests to API and Python dicts don't have autovivification built-in, I've changed `schemas.py` accordingly. I've detected this edge case from running the new test.

Regards,

lemur/certificates/schemas.py
```diff
@@ -112,10 +112,20 @@ def load_data(self, data):
...
+                data['extensions'] = {
+                    'subAltNames': {
+                        'names': []
+                    }
+                }
+            elif not data['extensions'].get('subAltNames'):
+                data['extensions']['subAltNames'] = {
+                    'names': []
+                }
+            elif not data['extensions']['subAltNames']['names']:
```
castrapel: `elif not data['extensions']['subAltNames'].get('names'):`

jramosf: I've added 58dd424de8781e852fbb4fac102273746120e5e2 to prevent this case, thanks!

castrapel:
This looks great! Thanks for the PR, I'll merge it. Tested by going through certificate creation flow, including CSRs, and setting breakpoints at modified code.

PR #2751, opened 2019-04-05
Rewrite Java Keystore/Truststore support based on pyjks library

intgr:
As discussed with @kevgliss in Gitter. This way it's not necessary to install the JDK to have JKS export support in Lemur.


hosseinsh:
Thanks Marti for the PR.
Just tested the export flow locally and merged the PR.

PR #2759, opened 2019-04-22
Adding SAN filter to Vault Destination Plugin

alwaysjolley:
Adds a regex filter to prevent non-matching SAN names from being uploaded to the Vault Destination.
- This filter is completely optional, defaults to pass
- If and SAN fails the regex filter the cert will not upload to Vault and the plugin will error
- Logs record the certificate did not pass the SAN filter
- Allows for more granular control of SAN certificates allowed per destination.


hosseinsh:
Thanks for the PR. Merging this.

PR #2773, opened 2019-04-30
Better error management in jwks token validation

jplana:
Instead of failing horribly when validating the jwks token, use the actual errors already controlled by the exception handlers.

### DIFF
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -337,9 +337,9 @@ class Ping(Resource):
         )
 
         jwks_url = current_app.config.get('PING_JWKS_URL')
-        result = validate_id_token(id_token, args['clientId'], jwks_url)
-        if result:
-            return result
+        error_code = validate_id_token(id_token, args['clientId'], jwks_url)
+        if error_code:
+            return error_code
         user, profile = retrieve_user(user_api_url, access_token)
         roles = create_user_roles(profile)
         update_user(user, profile, roles)
@@ -387,9 +387,9 @@ class OAuth2(Resource):
         )
 
         jwks_url = current_app.config.get('PING_JWKS_URL')
-        result = validate_id_token(id_token, args['clientId'], jwks_url)
-        if result:
-            return result
+        error_code = validate_id_token(id_token, args['clientId'], jwks_url)
+        if error_code:
+            return error_code
 
         user, profile = retrieve_user(user_api_url, access_token)
         roles = create_user_roles(profile)
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -331,8 +331,9 @@ class Ping(Resource):
         )
 
         jwks_url = current_app.config.get('PING_JWKS_URL')
-        validate_id_token(id_token, args['clientId'], jwks_url)
-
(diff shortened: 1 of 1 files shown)


lemur/auth/views.py
```diff
@@ -331,8 +331,9 @@ def post(self):
         )
 
         jwks_url = current_app.config.get('PING_JWKS_URL')
-        validate_id_token(id_token, args['clientId'], jwks_url)
-
+        result = validate_id_token(id_token, args['clientId'], jwks_url)
```
hosseinsh: from what I see validate_id_token() either returns the payload from jwt.decode or (dict(message=#message), 401). would this mean that now we always 'return' after this if-case?

jplana: As fas as I understand the code from `validate_id_token()` it will only return something on errors, but never the content of the token:

`jwt.decode(id_token, secret.decode('utf-8'), algorithms=[algo], audience=client_id)` 

The only explicit returns from that function seem to be errors

hosseinsh: hmm, looking at jwt.decode() it appears that it does return the payload.

```    def decode(self,
               jwt,  # type: str
               key='',   # type: str
               verify=True,  # type: bool
               algorithms=None,  # type: List[str]
               options=None,  # type: Dict
               **kwargs):

        if verify and not algorithms:
            warnings.warn(
                'It is strongly recommended that you pass in a ' +
                'value for the "algorithms" argument when calling decode(). ' +
                'This argument will be mandatory in a future version.',
                DeprecationWarning
            )

        payload, _, _, _ = self._load(jwt)

        if options is None:
            options = {'verify_signature': verify}
        else:
            options.setdefault('verify_signature', verify)

        decoded = super(PyJWT, self).decode(
            jwt, key=key, algorithms=algorithms, options=options, **kwargs
        )

        try:
            payload = json.loads(decoded.decode('utf-8'))
        except ValueError as e:
            raise DecodeError('Invalid payload string: %s' % e)
        if not isinstance(payload, Mapping):
            raise DecodeError('Invalid payload string: must be a json object')

        if verify:
            merged_options = merge_dict(self.options, options)
            self._validate_claims(payload, merged_options, **kwargs)

        return payload
```

jplana: Yes, `jwt.decode()` does return the content of the jwt token, but  `validate_id_token()` doesn't. This is (more or less) what `validate_id_token()` is doing:

```
In [1]: def decode():
   ...:     return "Something"
   ...:

In [2]: def validate():
   ...:     decode()
   ...:

In [3]: validate()

In [4]: decode()
Out[4]: 'Something'
``` 

Not that I'm very fond of the pattern in `validate_id_token()` (not returning any value) but It works (and is quite common), and in this case seems to be the intention, just return the errors.

hosseinsh: I see your point. Thanks for the clarification.

hosseinsh: for better readability, would you mind changing result into error_code or something similar to avoid any misunderstandings?

jplana: Sure

hosseinsh:
Thanks Jose. Just merged this PR.

PR #2774, opened 2019-04-30
Don't fail if googleGroups is not found in user profile

jplana:
This will aid when integration new Identity providers. Whether using `googleGroups` or any other more standard attribute probably is irrelevant (as most Identity providers will allow mapping any name here) but instead of failing it would be great to have a more detailed log trace.


hosseinsh:
Thanks Jose for the PR. Just merged it.

PR #2775, opened 2019-04-30
Enable gulp server to proxy backend

jplana:
While this code is not perfect, it was already there, but commented out. It took me a while to understand what was going on, and found it very useful to integrate/test Indentity Providers when developing. This should be harmless in production as probably nobody is serving statics via gulp there.


castrapel:
Tested this locally and confirmed that it looks good. Thanks for the PR!

PR #2779, opened 2019-05-06
Fix Cloudflare DNS

dciancu:
Fixes:
```
  File "/www/lemur/lemur/plugins/lemur_acme/cloudflare.py", line 71, in delete_txt_record
    zone_id, record_id = change_id
ValueError: not enough values to unpack (expected 2, got 1)
```

### DIFF
--- a/lemur/plugins/lemur_acme/cloudflare.py
+++ b/lemur/plugins/lemur_acme/cloudflare.py
@@ -66,11 +66,12 @@ def create_txt_record(host, value, account_number):
     return zone_id, r['id']
 
 
-def delete_txt_record(change_id, account_number, host, value):
+def delete_txt_record(change_ids, account_number, host, value):
     cf = cf_api_call()
-    zone_id, record_id = change_id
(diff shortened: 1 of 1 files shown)


lemur/plugins/lemur_acme/cloudflare.py
```diff
@@ -68,9 +68,10 @@ def create_txt_record(host, value, account_number):
 
 def delete_txt_record(change_id, account_number, host, value):
     cf = cf_api_call()
-    zone_id, record_id = change_id
-    current_app.logger.debug("Removing record with id {0}".format(record_id))
-    try:
-        cf.zones.dns_records.delete(zone_id, record_id)
-    except Exception as e:
-        current_app.logger.error('/zones.dns_records.post: %s' % e)
+    for change_id in change_ids:
```
hosseinsh: change_ids is not defined before.

dciancu: Thanks for the heads-up, solved it.

PR #2782, opened 2019-05-07
Converting userinfo authorization to a config var

castrapel:
@jplana - Unfortunately this [PR](https://github.com/Netflix/lemur/pull/2772) broke with our SSO provider. It's difficult for us to test the flow locally. I've converted this to use a configuration variable. Could you let me know your thoughts?

### DIFF
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -112,11 +112,17 @@ def retrieve_user(user_api_url, access_token):
     """
     user_params = dict(access_token=access_token, schema='profile')
 
+    headers = {}
+
+    if current_app.config.get('PING_INCLUDE_BEARER_TOKEN'):
+        headers = {'Authorization': f'Bearer {access_token}'}
+
     # retrieve information about the current user.
     r = requests.get(
         user_api_url,
         params=user_params,
-        headers={'Authorization': 'Bearer {}'.format(access_token)})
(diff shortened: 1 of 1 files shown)


jplana:
@castrapel Sorry about that! Didn't expect it, as `userinfo`endpoint is standarized in https://openid.net/specs/openid-connect-core-1_0.html#UserInfo and should accept the `Authorization: Bearer` header.

Researching a little I found that PING even document the userinfo endpoint here: https://www.pingidentity.com/content/developer/en/resources/openid-connect-developers-guide.html#userinfo_endpoint

With the following example:

```GET https://pf.company.com:9031/idp/userinfo.openid HTTP/1.1
Authorization: Bearer <access token>```

And found other examples (like here: https://www.pingidentity.com/content/ping/en/company/blog/posts/2013/userinfo-endpoint-and-scope-in-openid-connect.html)


Have you checked removing the token from the query parameters while keeping the header? (just curious). I agree that this is very hard to test locally, if you don't have any other measures in place (like limiting where to initiate authentication so you cannot initiate from your laptop) you could try with #2775 (this is how I tested our integration).

In any case, while I haven't checked the specific solution in this PR It looks like it would work, I'll check it and let you know. 

BTW, thanks for the update, I might have missed this change otherwise :-)

PR #2787, opened 2019-05-09
Allow searching for certificates by name via API

castrapel:


lemur/static/app/angular/certificates/view/view.tpl.html
```diff
@@ -47,7 +47,7 @@ <h2 class="featurette-heading">Certificates
             </td>
             <td data-title="''" style="text-align: center; vertical-align: middle;">
                 <div class="btn-group pull-right" role="group" aria-label="...">
-                  <a class="btn btn-sm btn-primary" ui-sref="certificate({name: certificate.name})">Permalink</a>
+                  <a class="btn btn-sm btn-primary" ui-sref="certificate_name({name: certificate.name})">Permalink</a>
```
hosseinsh: shall we keep this out of this change, to not break the Permalinks for now?

PR #2803, opened 2019-05-31
Deduplicate notifications

Wiston999:
This PR tries to reduce the number of notifications sent when owner and notification plugin recipients are the same.

### DIFF
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -155,7 +155,7 @@ def send_expiration_notifications(exclude):
 
             if (
                 notification_recipient
-                and owner != notification_recipient
+                and owner not in notification_recipient # notification_recipient is a list
                 and security_email != notification_recipient
             ):
                 if send_notification(
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -155,7 +155,7 @@ def send_expiration_notifications(exclude):
 
             if (
                 notification_recipient
-                and owner not in notification_recipient # notification_recipient is a list
+                and owner not in notification_recipient  # notification_recipient is a list
                 and security_email != notification_recipient
             ):
                 if send_notification(
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -145,6 +145,8 @@ def send_expiration_notifications(exclude):
             )
             if notification_recipient:
                 notification_recipient = notification_recipient.split(",")
+            else:
+                notification_recipient = []
 
             if send_notification(
                 "expiration", notification_data, [owner], notification


hosseinsh:
Thanks Victor for the PR. This is certainly of value. Would you be able to resolve the failing integration tests? that would be great.

lemur/notifications/messaging.py
```diff
@@ -155,7 +155,7 @@ def send_expiration_notifications(exclude):
 
             if (
                 notification_recipient
-                and owner != notification_recipient
+                and owner not in notification_recipient  # notification_recipient is a list
                 and security_email != notification_recipient
```
hosseinsh: is `notification_recipient` always a list? if this is the case, we should also change the next line `and security_email != notification_recipient`. Otherwise, if `notification_recipient` is not always a list, and can also be just a string, we might run to some undesired side-effects.

Wiston999: From what I see in previous lines, `notification_recipient` is built as: 
```
            notification_recipient = get_plugin_option(
                "recipients", notification.options
            )
            if notification_recipient:
                notification_recipient = notification_recipient.split(",")
```
When `get_plugin_option` doesn't find a value, it just doesn't return any value, so in that case `notification_recipient` will be `None`. I'll add logic to ensure that `notification_recipient` is always a list.

About `security_email`, it comes from `LEMUR_SECURITY_TEAM_EMAIL` configuration variable, which documetation at https://lemur.readthedocs.io/en/latest/administration.html shows an example being it a list, although there is no explicit check about it being a list. If they are both lists, `==` won't work as desired, and one of the solutions exposed [here](https://stackoverflow.com/questions/1388818/how-can-i-compare-two-lists-in-python-and-return-matches) should be used.

hosseinsh: good point Victor, `security_email` is a list, and the reason we have the equality check there is to prevent duplicate emails. A simple equality check for two lists won't work here, and if we were to do it properly, then we might as well just remove duplicate emails from the `notification_recipient`

```
>>> notification_recipient = ["a", "b", "c"]
>>> security_email = ["b", "s"]
>>> [i for i in notification_recipient if i not in security_email]
['a', 'c']
```

btw, if `notification_recipient` takes none, we don't need the empty list logic, since `notification_recipient` is part of the **and** in the if-statement. 

I have put my suggestions as a PR:
https://github.com/Netflix/lemur/pull/2806

any thoughts?

Wiston999: You are right, `notification_recipient` is checked for thruthy value and so the empty list logic is not needed, I'll remove it.

About your suggested PR, I think that the implementation is cleaner than mine. Do you want to close my PR and keep working on yours?

hosseinsh: sure, we can do that. Thanks for getting this conversation started. Will do some local testing, and merge the other one.

PR #2804, opened 2019-05-31
Common name auto addition as domain name

kush-bavishi:
Changes to allow automatically add the common name to the list of Domain names. How it works - When the user types a common name and moves focus from the text box, the common name gets added to the list of Domain Names. Any further edits to the common name will be appropriately reflected in the list on Domain names automatically as well.

image
Text box is in focus and the list of domain names is empty.

image
Text box loses focus, common name gets added to list

image
Subsequent alternate names get added to the list after the common name

image
Common name modified, changes reflect in the list of domain names


hosseinsh:
Thanks Kush for the PR. merging it.

PR #2807, opened 2019-06-11
Faster permalink

kush-bavishi:
Permalink button on the certificates page provides a permanent link to the specified certificate. These changes modify the behavior of Permalink. It now points to a new API which provides a faster response for better performance.


hosseinsh:
thanks Kush for this PR. The new endpoint `/api/1/certificates/name/<certname>` from https://github.com/Netflix/lemur/pull/2787 is indeed faster, since instead of a search it does a lookup based on the certname.

PR #2808, opened 2019-06-18
manage.py: Restore shebang line

intgr:
It's been a while, been busy with other stuff. This is a trivial fix that I've carried in my own branches for now:

> This is an executable file but cannot be executed without the interpreter.
> 
> The shebang line was lost in commit 8cbc6b8325c08bb3a72932c7e67c6476f7d29edb


hosseinsh:
Thanks, nice to have you back. Merging it

PR #2809, opened 2019-06-19
fixing regex to be more flexable

alwaysjolley:
updating the regex for vault plugin to be more flexible.


hosseinsh:
@alwaysjolley thanks for the PR. Merged it.

PR #2810, opened 2019-06-20
Expired Toggle

kush-bavishi:
This change updates the default behavior of Lemur to hide certs that have expired more than a month ago. (The current default behavior is to include certs that have expired up to 6 months ago). This will help lower the load on the server. There is also an additional button on the Certificates page now, that allows the user to view all certs, including all expired certs. The user can switch between showing and hiding expired certs.
image

image

Respective API changes have also been made to replicate this behavior. An additional parameter "showExpired=1" is required to retrieve all (including expired) certs, while viewing certificates.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -329,12 +329,14 @@ def render(args):
     """
     query = database.session_query(Certificate)
 
-    time_range = args.pop("time_range")
-    if not time_range:
-        six_month_old = arrow.now()\
-            .shift(months=current_app.config.get("HIDE_EXPIRED_CERTS_AFTER_MONTHS", -6))\
+    show_expired = args.pop("showExpired")
+    if show_expired != 1:
+        one_month_old = arrow.now()\
+            .shift(months=current_app.config.get("HIDE_EXPIRED_CERTS_AFTER_MONTHS", -1))\
             .format("YYYY-MM-DD")
-        query = query.filter(Certificate.not_after > six_month_old)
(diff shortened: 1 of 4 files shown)


hosseinsh:
Thanks Kush for the PR. Merging it now.

intgr:
This kind of API changes are very bad, totally took me by surprise. And I think it's bad API design as well, to filter results by default and you have to explicitly supply an argument to disable that filtering.

Why wasn't this implemented in a backwards-compatible manner -- with the default being to return all certificates?

The UI could provide query parameters to override the defaults anyway.

lemur/certificates/service.py
```diff
@@ -329,12 +329,14 @@ def render(args):
     """
     query = database.session_query(Certificate)
 
-    time_range = args.pop("time_range")
-    if not time_range:
-        six_month_old = arrow.now()\
-            .shift(months=current_app.config.get("HIDE_EXPIRED_CERTS_AFTER_MONTHS", -6))\
+    show_expired = args.pop("showExpired")
+    if show_expired != 1:
```
intgr: This is contrary to how integers are interpreted as boolean everywhere else.

Usually a value `0` means False, everything else means True.

But this is implemented so that `2` is also effectively False.

hosseinsh:
@intgr sorry for the inconvenience this has caused. 
I agree this could have been implemented in a cleaner way.

intgr:
And I'm sorry to be so blunt, I was irritated after a long debugging session. :)

PR #2814, opened 2019-06-22
Expose new certificate field hasPrivateKey

intgr:
We can also now disable the 'private key' tab when cert doesn't have a private key.


castrapel:
Thank you for the PR @intgr ! Merging now.

PR #2815, opened 2019-06-24
Cleanup no chain in Vault destination plugin

alwaysjolley:
fix for deploying with an empty chain.

was resulting in a string of 'None', now will result in an empty string.


castrapel:
Thank you for the PR @alwaysjolley ! Looks good to me, merging now.

PR #2817, opened 2019-06-26
Perma link resolving private key load

kush-bavishi:
These code changes fix the issue where the Permalink changes were breaking the PrivateKey functionality. Previously, the change for Permalink to call a new API was causing PrivateKey feature to point to a non existent API on the Permalink page. The Permalink feature is handled differently now, so as to not affect the PrivateKeyLoad feature. A parameter called "perma" is added to the URL for permalink and "perma=1" is added to the GET request to do a lookup for a certificate by its name instead of searching for it.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -353,6 +353,11 @@ def render(args):
         if terms[1].startswith('"') and terms[1].endswith('"'):
             term = terms[1][1:-1]
 
+        perma = args.pop("perma")
+        if perma == 1:
+            if "name" in terms:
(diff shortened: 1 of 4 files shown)


hosseinsh:
The "Show expired" and "Hide Expired" button are confusing in the Permalink page, and also that Filter search won't work while in that stage is a bummer. Currently, when one starts a new search while on the Permalink page, it gets redirected to Perma path, and the search key is considered as a certname.

zarfide:
Fixed via alternative PR: https://github.com/Netflix/lemur/pull/2873/files

PR #2825, opened 2019-07-10
Add ADCS Plugin to Docs

arnydo:
Add info regarding the ADCS plugin created by "https://github.com/sirferl/lemur".
"lemur_adcs" plugin is part of Lemur by default (at least included in the main repo) so I added to main plugins section within Docs. Please let me know if it should be listed under 3rd-party instead.


hosseinsh:
Thank @arnydo for the PR. I just merged it.

PR #2837, opened 2019-07-30
moving all cron jobs to become celery jobs

hosseinsh:
This helps with better logging and alerting

### DIFF
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -268,11 +268,14 @@ def clean_all_sources():
     be ran periodically. This function triggers one celery task per source.
     """
     function = f"{__name__}.{sys._getframe().f_code.co_name}"
+    log_data = {
+        "function": function,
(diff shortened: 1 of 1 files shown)


lemur/common/celery.py
```diff
@@ -366,3 +371,83 @@ def sync_source_destination():
...
+@celery.task()
+def certificate_reissue():
+    """
+    This celery task reissues certificates which are pending reissue
+    :return:
+    """
+    function = f"{__name__}.{sys._getframe().f_code.co_name}"
+    current_app.logger.debug(f"{function}: reissuing certificates")
+    cli_certificate.reissue(None, True)
+    current_app.logger.debug(f"{function}: reissuance completed")
```
castrapel: Can we log in dictionary format? This will prepare for the eventual JSONification of the logs. Example:

{"function": function,
"message": "reissuing certificates"
}

PR #2839, opened 2019-07-30
Ultradns plugin

kush-bavishi:
This change adds an ultraDNS plugin to Lemur, which will allow Lemur to issue Lets Encrypt certs for domains managed by ultraDNS.
3 new files are added - 
1. ultradns.py - This is the main code for the ultraDNS plugin. It handles ultradns authentication, creation and deletion of text records and checks for dns propagation
2. ultradns_record.py - Class for handling ultraDNS records
3. ultradns_zone.py - Class for handling ultraDNS zones


lemur/plugins/lemur_acme/ultradns.py
```diff
@@ -0,0 +1,323 @@
...
+def wait_for_dns_change(change_id, account_number=None):
+    # Waits and checks if the DNS changes have propagated or not.
+    # First check the domains authoritative server. Once this succeeds,
+    # we ask a public DNS server (Google <8.8.8.8> in our case).
+    fqdn, token = change_id
+    number_of_attempts = 20
+    nameserver = get_authoritative_nameserver(fqdn)
+    for attempts in range(0, number_of_attempts):
+        status = _has_dns_propagated(fqdn, token, nameserver)
+        current_app.logger.debug("Record status on ultraDNS authoritative server for fqdn: {}: {}".format(fqdn, status))
```
castrapel: Can you log in dictionary format for all the places you are logging? This will (eventually) be converted to JSON so we can parse out fields from the log better.
Example:
function = f"{__name__}.{sys._getframe().f_code.co_name}"
log_data = {
    "function": function,
    "fqdn": fqdn,
    "status": status,
    "message": "Record status on ultraDNS authoratative server"
}
current_app.logger.debug(log_data)

lemur/plugins/lemur_acme/ultradns.py
```diff
@@ -0,0 +1,323 @@
...
+            if z.count(".") > zone_name.count("."):
+                zone_name = z
+    if not zone_name:
+        metrics.send("ultradns_no_zone_name", "counter", 1)
+        raise Exception("No UltraDNS zone found for domain: {}".format(domain))
+    return zone_name
+
+
+def create_txt_record(domain, token, account_number):
+    # Create a TXT record for the given domain.
```
castrapel: Convert to docstring? Recommend for all of your functions: https://www.python.org/dev/peps/pep-0257/

lemur/plugins/lemur_acme/ultradns.py
```diff
@@ -0,0 +1,323 @@
...
+    zone_name = get_zone_name(domain, account_number)
+    zone_parts = len(zone_name.split("."))
+    node_name = ".".join(domain.split(".")[:-zone_parts])
+    path = "/v2/zones/{}/rrsets/16/{}".format(zone_name, node_name)
+
+    try:
+        rrsets = _get(path)
+        record = Record(rrsets)
+    except Exception as e:
+        metrics.send("delete_txt_record_geterror", "counter", 1)
```
castrapel: recommend `metrics.send(f"{function}.geterror,...`

Making the metrics based on the function name will allow us to ensure consistency.
Prefer dot notation to make it easier to distinguish function namespace against the metric name, underscores can be confusing because the function name also contains underscores

lemur/plugins/lemur_acme/ultradns_record.py
```diff
@@ -0,0 +1,26 @@
+class Record:
```
castrapel: Could this class exist in ultradns.py? Just to keep it all in one file, since the other plugins operate this way

lemur/plugins/lemur_acme/ultradns_zone.py
```diff
@@ -0,0 +1,33 @@
+class Zone:
```
castrapel: Ditto for this: possible to put in ultradns.py ?

lemur/plugins/lemur_acme/ultradns.py
```diff
@@ -96,12 +96,14 @@ def _has_dns_propagated(name, token, domain):
             for txt_record in rdata.strings:
                 txt_records.append(txt_record.decode("utf-8"))
     except dns.exception.DNSException:
-        metrics.send("has_dns_propagated_fail", "counter", 1)
+        function = sys._getframe().f_code.co_name
+        metrics.send("{}.fail".format(function), "counter", 1)
```
hosseinsh: to be consistent, it would be great if we could use f-strings for visibility:
`metrics.send(f"{function}.fail", "counter", 1)`

https://cito.github.io/blog/f-strings/

hosseinsh:
Thanks Kush for your contribution. Merging this in a few minutes.

PR #2840, opened 2019-08-06
[WIP] Adding wip example of how to add full text search to Lemur

kevgliss:
This is a basic example how how we might add full text search to Lemur (using only postgres).


hosseinsh:
closing this WIP, until we look into a better search bar in the near future, as part of the UI re-design.
@charhate for visibility

PR #2842, opened 2019-08-08
improving check revoked by only considering authorities which do supp…

hosseinsh:
…ort revocation and also only including not expired certs

This is an important update, specially for large DBs, and also for cases where an Authority does not implement revocation due to short lived certificates. 

To be included in the config file
```
SUPPORTED_REVOCATION_AUTHORITY_PLUGINS = []
```


lemur/certificates/service.py
```diff
@@ -102,6 +102,27 @@ def get_all_certs():
     return Certificate.query.all()
 
 
+def get_all_valid_certs(authority_plugin_name):
+    """
+    Retrieves all valid (not expired) certificates within Lemur, for the given authority plugin names
+    ignored if no authority_plugin_name provided.
+
+    Note that depending on the DB size retrieving all certificates might an expensive operation
```
csine-nflx, line 110: might "be" an expensive...

csine-nflx:
The addition of get_all_valid_certs() makes sense to me.  Good improvement.

However, I am not a fan of having to hard-code which CAs support revocation services. It would be better to check the certs/CAs for revocation information in the AIA field.  This is the industry standard way to know if a cert supports revocation.  If it contains revocation information, then we check the revocation information.  The configurable bit should be what we do if the revocation service is not available or does not respond on a per-CA basis, not a per-plugin basis.

hosseinsh:
good points, we shall consider checking AIA field in the next revision. However, having the possibility of adding/removing Issuer plugins for this check seems a valuable value, considering the volumes of certs issued and needed to be checked frequently for revocation

PR #2848, opened 2019-08-20
Show number of found items in pager

jramosf:
This commit does not involve any additional query as the data is already in API calls' responses


hosseinsh:
Thanks @jramosf for the PR. Just merged it.

Issue #2861, opened 2019-10-10
SFTP plugin - issue updating replacement certs fails due to overwrite permissions


pmelse:
steps to reproduce:

1. import cert to lemur
2. successfully deploy that cert with the sftp plugin.
3. create a cert that replaces the old cert.
4. on deployment, you will see:

> Oct 10 11:17:57 lemur lemur[778]: /etc/ssl/lemur/*.private.somedir.ca already exist, resuming
Oct 10 11:17:57 lemur lemur[778]: Uploading *.private.feds.ca.key to /etc/ssl/lemur/*.private.somedir.ca
Oct 10 11:17:57 lemur lemur[778]: [2019-10-10 11:17:57,285] ERROR in plugin: ERROR in <class 'PermissionError'>: [Errno 13] Permission denied
Oct 10 11:17:57 lemur lemur[778]: ERROR in <class 'PermissionError'>: [Errno 13] Permission denied

> the extent files are still 0400 as deployed the first round from lemur.

I'm thinking we need a check in https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_sftp/plugin.py around line 168 to check to see if the files already exist in the subdirectory for the cert, and if so, delete them.

pmelse:
this is fixed by #2876

PR #2862, opened 2019-10-10
bug fix for overwriting certificates with sftp plugin

pmelse:
this is my fix for issue #2861 

I wanted to preserve file permissions unless absolutely necessary to change them, and also throw an error to logs that this occurs.


hosseinsh:
Thank you @pmelse. 
Looks likeTravis is complaining, would be great if you could resolve this issue.

pmelse:
ha, this is what I get for using an html editor. (i'll fix the trailing whitespace, and resubmit)

pmelse:
fixed the trailing whitespace, will resubmit

PR #2864, opened 2019-10-15
creating a dict of certs with mismatched names compared to the Lemur DB

hosseinsh:


lemur/sources/service.py
```diff
@@ -152,6 +167,8 @@ def sync_certificates(source, user):
             cert = parse_certificate(certificate["body"])
             matching_serials = certificate_service.get_by_serial(serial(cert))
             exists = find_matching_certificates_by_hash(cert, matching_serials)
+            # create a mapping for mis-matched certificates due to different names
+            certs_with_mismatched_name[certificate["name"]] = [x for x in exists if x]
```
zarfide: I'm not familiar with this python syntax; do we need special handling for the case where it's still not found even by hash?

lemur/sources/service.py
```diff
@@ -89,6 +89,20 @@ def sync_endpoints(source):
 
         endpoint["certificate"] = certificate_service.get_by_name(certificate_name)
 
+        if not endpoint["certificate"] and certificate_name in certs_with_mismatched_name:
+            endpoint["certificate"] = certs_with_mismatched_name[certificate_name][0]
```
zarfide: Worth incrementing a metric of found by hash?

lemur/sources/service.py
```diff
@@ -65,7 +65,7 @@ def sync_update_destination(certificate, source):
             certificate.destinations.append(dest)
 
 
-def sync_endpoints(source):
+def sync_endpoints(source, certs_with_mismatched_name):
```
zarfide: Any way we can test this?

lemur/sources/service.py
hosseinsh: `find_matching_certificates_by_hash ` returns a list, and in this case`[x for x in exists if x]` would work. However, I agree, it is more safe to  make sure exists is a list

lemur/sources/service.py
hosseinsh: added more metrics

PR #2869, opened 2019-10-18
Cert sync endpoint find by hash

hosseinsh:
endpoints such as load balancers have the name of the cert, however this name might vary with Lemur's local DB. To be able to link all certs to Lemur's local DB, we now also fetch the actual cert for each endpoint, for better search, for instance via serial number and hash digest. Discovery of new certs is still view source sync.

### DIFF
--- a/lemur/plugins/lemur_aws/plugin.py
+++ b/lemur/plugins/lemur_aws/plugin.py
@@ -280,11 +280,12 @@ class AWSSourcePlugin(SourcePlugin):
             certificate_name = certificate_name.split('/')[1]
         try:
             cert = iam.get_certificate(certificate_name, account_number=account_number)
-            return dict(
-                body=cert["CertificateBody"],
-                chain=cert.get("CertificateChain"),
-                name=cert["ServerCertificateMetadata"]["ServerCertificateName"],
-            )
+            if cert:
+                return dict(
+                    body=cert["CertificateBody"],
+                    chain=cert.get("CertificateChain"),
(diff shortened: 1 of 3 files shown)


lemur/plugins/lemur_aws/plugin.py
```diff
@@ -271,6 +273,28 @@ def clean(self, certificate, options, **kwargs):
         account_number = self.get_option("accountNumber", options)
         iam.delete_cert(certificate.name, account_number=account_number)
 
+    def get_certificate_by_name(self, certificate_name, options):
+        account_number = self.get_option("accountNumber", options)
+        # certificate name may contain path, in which case we remove it
+        if "/" in certificate_name:
+            certificate_name = certificate_name.split('/')[1]
```
zarfide: Should we be getting the last element as opposed to the second?

lemur/sources/service.py
```diff
@@ -89,6 +89,39 @@ def sync_endpoints(source):
...
+            if certificate_attached_to_endpoint:
+                lemur_matching_cert, updated_by_hash_tmp = find_cert(certificate_attached_to_endpoint)
+                updated_by_hash += updated_by_hash_tmp
+
+                if lemur_matching_cert:
+                    endpoint["certificate"] = lemur_matching_cert[0]
+
+                if len(lemur_matching_cert) > 1:
+                    current_app.logger.error(
+                        "Too Many Certificates Found. Name: {0} Endpoint: {1}".format(
```
zarfide: Should error indicate just how many as well?

PR #2875, opened 2019-12-20
Eliminate SQL subqueries when showing certificates list

ilyalabun:
When loading list of certificates in Web UI  endpoint `/api/1/certificates` returns full Certificate entry with all associations loaded.It produces about 7-8 SQL requests for **each** returned certificate, i.e. several hundreds requests for large pages (50-100 certificates). Given that certificates table needs only few Certificate fields it makes sense to return only data we need. It eliminates tons of SQL queries and reduces traffic a bit :)

So this PR introduces boolean query parameter `short` for `/api/v1/certificates` which tells backend to return only those certificate fields we need to render certificate table.

### DIFF
--- a/lemur/certificates/schemas.py
+++ b/lemur/certificates/schemas.py
@@ -6,6 +6,8 @@
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
 from flask import current_app
+from flask_restful import inputs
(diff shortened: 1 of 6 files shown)


hosseinsh:
Thank you @ilyalabun for the PR, very practical change, improving the UI flow.

PR #2876, opened 2019-12-27
updated permission fix for sftp plugin if certificate is set read-only

pmelse:


hosseinsh:
thanks for the PR  @pmelse!
Just merged it.

Issue #2879, opened 2020-01-07
Unable to create certificates using the CFSSL Issuer Plugin


rohanmehto2:
There is an error in creating a certificate using CFSSL as Issuer.

Logs from the lemur server:

```
[2020-01-07 17:46:32,386] ERROR in service: Exception minting certificate
Traceback (most recent call last):
  File "/www/lemur/lemur/certificates/service.py", line 273, in create
    cert_body, private_key, cert_chain, external_id, csr = mint(**kwargs)
  File "/www/lemur/lemur/certificates/service.py", line 223, in mint
    cert_body, cert_chain, external_id = issuer.create_certificate(csr, kwargs)
  File "/www/lemur/lemur/plugins/lemur_cfssl/plugin.py", line 58, in create_certificate
    key = bytes.fromhex(hex_key)
TypeError: fromhex() argument must be str, not None
Exception minting certificate
Traceback (most recent call last):
  File "/www/lemur/lemur/certificates/service.py", line 273, in create
    cert_body, private_key, cert_chain, external_id, csr = mint(**kwargs)
  File "/www/lemur/lemur/certificates/service.py", line 223, in mint
    cert_body, cert_chain, external_id = issuer.create_certificate(csr, kwargs)
  File "/www/lemur/lemur/plugins/lemur_cfssl/plugin.py", line 58, in create_certificate
    key = bytes.fromhex(hex_key)
TypeError: fromhex() argument must be str, not None
Exception minting certificate
Traceback (most recent call last):
  File "/www/lemur/lemur/certificates/service.py", line 273, in create
    cert_body, private_key, cert_chain, external_id, csr = mint(**kwargs)
  File "/www/lemur/lemur/certificates/service.py", line 223, in mint
    cert_body, cert_chain, external_id = issuer.create_certificate(csr, kwargs)
  File "/www/lemur/lemur/plugins/lemur_cfssl/plugin.py", line 58, in create_certificate
    key = bytes.fromhex(hex_key)
TypeError: fromhex() argument must be str, not None
[2020-01-07 17:46:32,387] ERROR in schema: fromhex() argument must be str, not None
Traceback (most recent call last):
  File "/www/lemur/lemur/common/schema.py", line 160, in decorated_function
    resp = f(*args, **kwargs)
  File "/www/lemur/lemur/certificates/views.py", line 482, in post
    cert = service.create(**data)
  File "/www/lemur/lemur/certificates/service.py", line 273, in create
[...]

PR #2881, opened 2020-01-09
fix for #2879 cfssl-key TypeError

rohanmehto2:
Mention changed [here](https://github.com/Netflix/lemur/issues/2879#issuecomment-572049821).

If CFSSL-KEY is not provided in the config the code halts which shouldn't happen since the key is optional.


hosseinsh:
Thanks for the  PR @rohanmehto2 
Merged it.

Issue #2882, opened 2020-01-10
Kubernetes authentication option for Vault plugins


rohanmehto2:
Feature request:
It would be great if the vault plugins supported other authentication methods too, instead of only token.
I'll be willing to add this for kubernetes and perhaps for other auth methods too.
If this seems ok then I just need some pointers regarding how to structure forms in the UI.

PR #2883, opened 2020-01-11
implemented #2882 k8s auth for vault

rohanmehto2:
[Functionality](https://github.com/Netflix/lemur/issues/2882#issue-548144411) wise its works perfectly but I was not able to find a better way to implement it for the UI. It doesn't seem possible to have conditional sub options in the user interface form according to the current implementation.
Please let me know if there is a better way to implement this for the user interface form. I'll make the necessary changes if required. Because if there is a way to have conditional sub options in the form then I can even add more authentication methods for vault.


hosseinsh:
Thanks @rohanmehto2  for the  PR. I think since the functionality is given, let's for now move on with the merging. We can revise this, if there are better ideas for the user interface form

PR #2891, opened 2020-01-23
fix fatal error in schema validator

rajatsharma94:
Currently the request crashes and gives out a 500 instead of handling the missing field gracefully and providing an invalid schema error with a 400.


hosseinsh:
Thank you @rajatsharma94 for the PR. 
Just merged it.

PR #2900, opened 2020-01-27
freeze dependencies for more reliable builds

### LINKED ISSUE #2899: Check in lockfile
Please follow Node.js standards and include the `package-lock.json` file in version control.

This file helps DevOps staff to ensure that builds are reliably reproduced the same way each time. This also reduces the likelihood of grandchild dependencies from introducing breaking changes.

mcandre:
Closes #2899


hosseinsh:
Thanks for raising this PR. have crafted a new one based on our  latest lock in production
https://github.com/Netflix/lemur/pull/2985

PR #2910, opened 2020-01-31
Fixes and enhancements of docker image

e11it:
Hi!

- fix nginx ssl conf
- add example docker-compose.yml
- change cmd order in Docker for fast rebuild
- fix entrypoint
- allow to pass more cfg options via Env


docker/pgsql-env
```diff
@@ -0,0 +1,4 @@
+POSTGRES_USER=lemur
+POSTGRES_PASSWORD=12345
```
hosseinsh: shall we leave this rather empty?

e11it: This is required due to checks in entripoint script.

https://github.com/Netflix/lemur/blob/30ab0bcdbef1b042a081945cb33824157bb70afe/docker/entrypoint#L3

hosseinsh: Maybe something along the lines, would avoid us from having a default PASSWORD in code.
```
import random
import string
secret_key = ''.join(random.choice(string.ascii_letters) for x in range(20))
```

e11it: This file used only in docker-compose to pass environment.
```
env_file:
       - pgsql-env
```
Files `lemur-env` and `pgsql-env` excluded from docker image. They are used only in docker-compose.
`docker-compose.yml` was added as example to help users to start using lemur.

This is the same as in docker-compose for tests:
https://github.com/Netflix/lemur/blob/30ab0bcdbef1b042a081945cb33824157bb70afe/docker-compose.yml#L12
https://github.com/Netflix/lemur/blob/30ab0bcdbef1b042a081945cb33824157bb70afe/docker-compose.yml#L20

hosseinsh:
Thanks for the PR.

hosseinsh:
merged it

PR #2917, opened 2020-02-15
New variable VERISIGN_PRODUCT_(authority.name)

sirferl:
If there is a config variable with VERISIGN_PRODUCT_<upper(authority.name)> take the value as Cert product-type else default to "Server" to be compatible with former versions.
This enables the creation of different Verisign cert-products eg. EV or Standard Certs using several cert authorities.

### DIFF
--- a/lemur/plugins/lemur_verisign/plugin.py
+++ b/lemur/plugins/lemur_verisign/plugin.py
@@ -98,14 +98,12 @@ def process_options(options):
     :param options:
     :return: dict or valid verisign options
     """
-
     # if there is a config variable with VERISIGN_PRODUCT_<upper(authority.name)> take the value as Cert product-type
     # else default to "Server", to be compatoible with former versions
     authority = options.get("authority").name.upper()
     product_type = current_app.config.get("VERISIGN_PRODUCT_{0}".format(authority))
     if product_type is None:
         product_type = "Server"
-
     data = {
         "challenge": get_psuedo_random_string(),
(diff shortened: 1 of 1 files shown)


sirferl:
@hosseinsh : Hi, I've  been a bit productive. Hope this can be added swiftly, it would help us with a situation. Thank you !

lemur/plugins/lemur_verisign/plugin.py
```diff
@@ -98,10 +98,16 @@ def process_options(options):
     :param options:
     :return: dict or valid verisign options
     """
+    # if there is a config variable with VERISIGN_PRODUCT_<upper(authority.name)> take the value as Cert product-type
+    # else default to "Server", to be compatoible with former versions
+    authority = options.get("authority").name.upper()
+    product_type = current_app.config.get("VERISIGN_PRODUCT_{0}".format(authority))
```
hosseinsh: One minor suggestion, to leverage  the default  option, and avoid the following if-statement.
``product_type = current_app.config.get("VERISIGN_PRODUCT_{0}".format(authority), "Server")``

hosseinsh:
hi @sirferl, thanks for the PR. I had only a minor suggestion. Otherwise it looks all good to me.

sirferl:
@hosseinsh: Thank you for showing me the right way to do it. I added youe suggestion.

hosseinsh:
Thanks. Just merged it.

Issue #2921, opened 2020-02-27
San values are duplicated on on reissuing certificates


e11it:
Hi!

San values are duplicated on reissue certificates.

Steps to reproduce:
- create short live cettificate with cryptography plugin
- reissue certs with command: `lemur -c /home/lemur/.lemur/lemur.conf.py certificate reissue -c`

Example:

I have created cert(output from web)
```
Distinguished Name CN=rotate-exp,....
Certificate Authority KafkaUser (kafkausers)
Serial Hex: C9FBA6E304B8471AB8974AFCD7725A2D
Decimal: 268481478530508147274752666851746142765
Valid From 2020-02-26T21:00:00+00:00 (17 hours ago)
Valid To 2020-02-28T21:00:00+00:00 (in a day)
SAN x
Key Length 2048
Signing Algorithm sha256
```

Reissue certificate.
Output:
```
[!] Running in COMMIT mode.
[+] Starting certificate re-issuance.
[+] rotate-exp-kafkausers-20200226-20200228 is eligible for re-issuance
[+] Re-issuing certificate with the following details:
	[+] Common Name: rotate-exp
	[+] Subject Alternate Names: rotate-exp,rotate-exp
	[+] Authority: KafkaUser
	[+] Validity Start: 2020-02-27T13:54:28.756101+00:00
	[+] Validity End: 2020-02-29T13:54:28.756101+00:00

[2020-02-27 13:54:28,949] DEBUG in service: Adding Critical Extension: sub_alt_names {'names': <SubjectAlternativeName(<GeneralNames([<DNSName(value='rotate-exp')>, <DNSName(value='rotate-exp')>])>)>}
Adding Critical Extension: sub_alt_names {'names': <SubjectAlternativeName(<GeneralNames([<DNSName(value='rotate-exp')>, <DNSName(value='rotate-exp')>])>)>}
[...]

PR #2922, opened 2020-02-27
Docker fixes. Vault dest path and obj_name support vars from cert

e11it:
Hi!

* docker/Dockerfile-src 
  - To build directly from repo (useful for fork)
* docker/entrypoint
  -  use lemur instead of manage.py 
  - add reissue cron job
  - fix cron: file name must match username(was broken)
* docker/src/lemur.conf.py
  -  use `str` instead of `repr` because `reps` adds quotes
  - add more variables
* docker/supervisor.conf 
  - add env to cron too
  -  use lemur instead of manage.py
* lemur/plugins/lemur_vault_dest/plugin.py
  -  options `path` and `obj_path` can contains vars: `{CN|OU|O|L|S|C}` from exported cert.
  - fix bug with undefined auth_key


e11it:
+ fix for #2921

e11it:
Hi! 
I think it will be better to allow to use variables in vaultPath and don't add new option vaultSuffix.

I will come back with new implementation.

e11it:
Bug:

/opt/lemur/lemur/plugins/lemur_vault_dest/plugin.py
```
auth_key = self.get_option("tokenFile/vaultRole", options)
```
will return None due to slash.

Rename value to `"tokenFileOrVaultRole`.

e11it:
@hosseinsh , hi!
It's ready to merge.

lemur/certificates/schemas.py
```diff
@@ -146,7 +146,9 @@ def load_data(self, data):
                 data["extensions"]["subAltNames"] = {"names": []}
             elif not data["extensions"]["subAltNames"].get("names"):
                 data["extensions"]["subAltNames"]["names"] = []
-            data["extensions"]["subAltNames"]["names"] += csr_sans
+
+            if not data["extensions"]["subAltNames"]["names"]:
```
hosseinsh: is there a reason for this additional check?
do you think it is important to prevent overwriting ?
`data["extensions"]["subAltNames"]["names"]`

e11it: This fix bug that I described here: https://github.com/Netflix/lemur/issues/2921.
When we always append `csr_sans` it leads to duplicates sans when cert is rotated and number of san records increase in a progression.

hosseinsh:
Hi @e11it, sorry this is taking for ever.
Due to the current COVID-19 situation, there might be longer delays with PR reviews. As much as I would love to move forward with reviews, I am running short on time. 
Hope to be able to do some testing for this PR, before merging it. Hopefully sometime in the coming days.

lemur/certificates/schemas.py
hosseinsh: sorry for this taking so long. I would suggest to change 
```
if not data["extensions"]["subAltNames"]["names"]:
   data["extensions"]["subAltNames"]["names"] += csr_sans
```
to
```
   data["extensions"]["subAltNames"]["names"] = csr_sans
```

Lemur does take the CSR provided, for the request to the CA, and hence, the SANS provided in the CSR are considered. If the SANS provided via the UI do not match the CSR this would result into issues during the rotation, so I believe we should take the CSR as the ground truth

PR #2931, opened 2020-03-15
New clean cert cli

hosseinsh:
adding to new cli commands for cleaning certificates from source:

a) either about to expire in X days and not attached to an endpoint
a) or issued since X days but still not attached to an endpoint

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -117,6 +117,41 @@ def get_all_pending_cleaning(source):
     )
 
 
+def get_all_pending_cleaning_about_to_expire_certs(source, days_to_expire):
+    """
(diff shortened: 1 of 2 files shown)


lemur/certificates/service.py
```diff
@@ -117,6 +117,41 @@ def get_all_pending_cleaning(source):
...
 
 
+def get_all_pending_cleaning_about_to_expire_certs(source, days_to_expire):
+    """
+    Retrieves all certificates that are available for cleaning: not attached to endpoint,
+    and within X days from expiration.
+
+    :param days_to_expire:
+    :param source:
+    :return:
```
csine-nflx: Please update the documentation for each param and return for clarity.

hosseinsh: thanks for the feedback

PR #2936, opened 2020-03-27
Bugfix:  compare offset-naive and offset-aware datetimes

hosseinsh:
expired is now called for new certs, where the not_after field might be in datetime format, and not comparable to utc

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -8,8 +8,6 @@
 from datetime import timedelta
 
 import arrow
-import pytz
-import datetime
 from cryptography import x509
 from cryptography.hazmat.primitives.asymmetric import rsa
 from flask import current_app
@@ -323,13 +321,9 @@ class Certificate(db.Model):
 
     @hybrid_property
     def expired(self):
-        if isinstance(self.not_after, datetime.datetime):
-            # can't compare offset-naive and offset-aware datetimes
-            if self.not_after.replace(tzinfo=pytz.UTC) <= arrow.utcnow():
-                return True
-        else:
-            if self.not_after <= arrow.utcnow():
-                return True
+        # can't compare offset-naive and offset-aware datetimes
+        if arrow.Arrow.fromdatetime(self.not_after) <= arrow.utcnow():
+            return True
 
     @expired.expression
     def expired(cls):
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -8,6 +8,8 @@
 from datetime import timedelta
 
 import arrow
+import pytz
+import datetime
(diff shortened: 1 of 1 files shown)


lemur/certificates/models.py
```diff
@@ -321,8 +323,13 @@ def public_key(self):
 
     @hybrid_property
     def expired(self):
-        if self.not_after <= arrow.utcnow():
-            return True
+        if isinstance(self.not_after, datetime.datetime):
+            # can't compare offset-naive and offset-aware datetimes
+            if self.not_after.replace(tzinfo=pytz.UTC) <= arrow.utcnow():
+                return True
+        else:
```
csine-nflx: not_after is of type datetime.  There should be no situation where this else clause is encountered, right?  If there is, then we need to fix how we are storing data in this column.  With that in mind, maybe we can simplify to the following?  See the Warning at https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp.

def expired(self):
  if datetime.fromtimestamp(self.not_after, tz=timezone.utc) <= arrow.utcnow():
    return True:

hosseinsh: thanks for your thorough feedback. I like your alternative solution. 
since it can handle both datetime and arrow

```
>>> arrow.Arrow.fromdatetime(arrow.utcnow()) <= arrow.utcnow()
True
>>> arrow.Arrow.fromdatetime(datetime.datetime.now()) <= arrow.utcnow()
True
```

to be honest, I am not sure about not_after field being always datetime, then this bug would have surfaced way earlier.


I see
```
    not_after = fields.DateTime()
    validity_end = ArrowDateTime(attribute="not_after")
```
https://github.com/Netflix/lemur/blob/21c651a56610ef2d099290b526b2a1cced93944f/lemur/certificates/schemas.py#L261-L262

ans also 
```
    not_before = Column(ArrowType)
    not_after = Column(ArrowType)
```
https://github.com/Netflix/lemur/blob/37e58574066921f50485398d195994b1b805c216/lemur/certificates/models.py#L138-L139


and I also see a [cast](https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.cast) here:
https://github.com/Netflix/lemur/blob/37e58574066921f50485398d195994b1b805c216/lemur/reporting/service.py#L47

csine-nflx: I believe these mean that not_after is ALWAYS of type datetime, and validity_end is of type Arrow, but validity_end is a calculated field (not stored in the DB, calculated at runtime).  The models.py code doesn't quite make sense to me yet.  gonna dig into that when I get some time.

PR #2950, opened 2020-04-24
fixing the private DNS zone issue.

hosseinsh:
Private hosted zones will never be visible to third-parties like LetsEncrypt, and Lemur should not consider them as authoritative zones.
This fix, make sure  they are not added to the  dns_provider table.


hosseinsh:
The acme plugin finds the dns providers based on the zones in the dns_provider table, and then it calls create_txt_record, where the zone_id for that account is retrieved.


hence in fine_zone_id we should also ignore PrivateZone
since the private zone might be on the same account, as the authoritative account

PR #2951, opened 2020-04-28
Autorotate certs attached to endpoints

castrapel:
Celery task to enable autorotate for all certificates attached to endpoints without it enabled

### DIFF
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -17,10 +17,8 @@ from celery.signals import task_failure, task_received, task_revoked, task_succe
 from datetime import datetime, timezone, timedelta
 from flask import current_app
 
-from lemur import database
 from lemur.authorities.service import get as get_authority
 from lemur.certificates import cli as cli_certificate
-from lemur.certificates.service import get_all_certs_attached_to_endpoint_without_rotate
 from lemur.common.redis import RedisHandler
 from lemur.destinations import service as destinations_service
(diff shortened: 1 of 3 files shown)


lemur/common/celery.py
```diff
@@ -812,3 +814,25 @@ def notify_expirations():
...
+    task_id = None
+    if celery.current_task:
+        task_id = celery.current_task.request.id
+
+    log_data = {
+        "function": function,
+        "task_id": task_id,
+    }
+
+    eligible_certs = get_all_certs_attached_to_endpoint_without_rotate()
```
hosseinsh, line 831: I was wondering if should outsource this part to a certificates/cli.py command?
this would keep the celery part more lean, allow for other types of scheduled tasks to benefit from this  functionality.

lemur/common/celery.py
```diff
@@ -812,3 +812,24 @@ def notify_expirations():
...
+        task_id = celery.current_task.request.id
+
+    log_data = {
+        "function": f"{__name__}.{sys._getframe().f_code.co_name}",
+        "task_id": task_id,
+        "message": "Enabling autorotate to eligible certificates",
+    }
+    current_app.logger.debug(log_data)
+
+    cli_certificate.automatically_enable_autorotate()
```
hosseinsh, line 835: shall we add 

```    
metrics.send(f"{function}.success", "counter", 1)
return log_data
```

PR #2953, opened 2020-05-07
Certificate rotation region by region

hosseinsh:
At times it might be necessary to rotate certs region by region. This PR enables this feature by allowing the celery task to receive a region param. 
A new rotate method considers the region and rotates only matching patterns in the dnsname of the endpoint, for instance, rotate only if "us-east-1" in the dnsname of the hostname. 

one can according put different schedules for rotation region-by-region, and then one default one to rotate any missing regions.

example scheudule:
```
CELERYBEAT_SCHEDULE = {
    'certificate_rotate': {
        'task': 'lemur.common.celery.certificate_rotate',
        'options': {
            'expires': 180
        },
        'schedule': crontab(minute="*"),
        'kwargs': {'region': 'us-east-1'}
    }
}
```

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -648,18 +648,33 @@ def check_revoked():
 @manager.command
 def automatically_enable_autorotate():
     """
-    This function automatically enables autorotation for unexpired certificates that are
+    This function automatically enables auto-rotation for unexpired certificates that are
     attached to an endpoint but do not have autorotate enabled.
+
+    WARNING: This will overwrite the Auto-rotate toggle!
     """
     log_data = {
         "function": f"{__name__}.{sys._getframe().f_code.co_name}",
     }
 
+    permitted_authorities = current_app.config.get("ENABLE_AUTO_ROTATE_AUTHORITY", [])
+
     eligible_certs = get_all_certs_attached_to_endpoint_without_autorotate()
     for cert in eligible_certs:
+
+        if cert.authority_id not in permitted_authorities:
+            continue
+
         log_data["certificate"] = cert.name
         log_data["certificate_id"] = cert.id
         log_data["message"] = "Enabling auto-rotate for certificate"
         current_app.logger.info(log_data)
+        # TODO:  add the cert destination to the logging
+        metrics.send("automatically_enable_autorotate",
+                     "counter", 1,
+                     metric_tags={"certificate": cert.name,
+                                  "certificate_id": cert.id,
+                                  "authority_id": cert.authority_id,
+                                  "authority_name": Authority.get(cert.authority_id).name})
         cert.rotation = True
         database.update(cert)
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -285,6 +285,17 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
     )
 
 
+def request_rotation_region(region, endpoint, new_cert, message, commit, log_data):
+    if region in endpoint.dnsname:
+        log_data["message"] = "Rotating endpoint in region"
+        #request_rotation(endpoint, new_cert, message, commit)
+    else:
+        log_data["message"] = "Skipping rotation, region mismatch"
+
+    print(log_data)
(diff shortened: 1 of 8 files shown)


lemur/certificates/cli.py
```diff
@@ -285,6 +285,171 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
+                    log_data = f"[+] Rotating {endpoint.dnsname} in {region}"
+                    print(log_data)
+                    current_app.logger.info(log_data)
+                    request_rotation(endpoint, new_cert, message, commit)
+                else:
+                    log_data = f"[+] Skipping rotation of {endpoint.dnsname} since not in {region}"
+                    print(log_data)
+                    current_app.logger.info(log_data)
+        else:
+            log_data = f"[+] Rotating all endpoints that have new certificates available in {region}"
```
csine-nflx, line 377: It is not obvious from the docstring that this method will rotate all certificates in a region, which could have some negative consequences.  You could just update the docstring, or alternatively, break rotate_region() into two separate methods, something like rotate_region() to rotate an entire region and an updated version of rotate_certificate() that takes a region argument to rotate a single certificate.

csine-nflx: Looking at rotate(), the fact that it will rotate all certificates that are pending is not documented either.  May be worth just updating the docstrings and calling it a day.

lemur/certificates/cli.py
```diff
@@ -285,6 +285,171 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
+    try:
+        old_cert = validate_certificate(old_certificate_name)
+        new_cert = validate_certificate(new_certificate_name)
+        endpoint = validate_endpoint(endpoint_name)
+
+        if endpoint and new_cert:
+            if region in endpoint.dnsname:
+                log_data = f"[+] Rotating endpoint in region: {endpoint.dnsname} to certificate {new_cert.name}"
+                print(log_data)
+                current_app.logger.info(log_data)
```
castrapel, line 355: Can we log this as a dictionary, so it'll get converted to json? endpoint_name, region, commit, and the certificate information should all be in their own fields.

lemur/certificates/cli.py
```diff
@@ -285,6 +285,171 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
+    print("[+] Starting endpoint rotation.")
+    status = FAILURE_METRIC_STATUS
+
+    try:
+        old_cert = validate_certificate(old_certificate_name)
+        new_cert = validate_certificate(new_certificate_name)
+        endpoint = validate_endpoint(endpoint_name)
+
+        if endpoint and new_cert:
+            if region in endpoint.dnsname:
```
castrapel, line 352: We'll need to keep in mind that `region in endpoint.dnsname` is very generic to AWS load balancers. Cloudfront distributions do not have region in the dns name, and other endpoint types may not either.

lemur/certificates/cli.py
```diff
@@ -285,6 +285,171 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
+                log_data = f"[+] Rotating endpoint in region: {endpoint.dnsname} to certificate {new_cert.name}"
+                print(log_data)
+                current_app.logger.info(log_data)
+                request_rotation(endpoint, new_cert, message, commit)
+            else:
+                log_data = f"[+] Skipping rotation of {endpoint.dnsname} since not in {region}"
+                print(log_data)
+                current_app.logger.info(log_data)
```
castrapel, line 360: Ditto here for log_data

lemur/certificates/cli.py
hosseinsh: correct, to cover for these cases, the recommended path would be run the rotate_region explicitly for different regions, but then run also a generic one to capture  all the other ones.

lemur/certificates/cli.py
hosseinsh: excellent idea, have now adjusted the logging and will do the same for the old rotation job.

PR #2975, opened 2020-05-12
WIP Plugin destination for Fastly CDN

alwaysjolley:
Adding destination for uploading and maintaining certificates in Fastly CDN service


alwaysjolley:
I am not sure how we can work around this but the use of hashlib.sha1 is to validate the unique identifier from Fastly to prevent duplicate uploads to the destination. We are not using sha1 for any form of security and it should not be flagged as such.

kevgliss:
You can add a bandit ignore line (#nosec), see: https://github.com/PyCQA/bandit#exclusions

hosseinsh:
looks like lint still not happy 

```
lemur/plugins/lemur_fastly/plugin.py:27:1: E302 expected 2 blank lines, found 1
lemur/plugins/lemur_fastly/plugin.py:40:5: E303 too many blank lines (2)
lemur/plugins/lemur_fastly/plugin.py:44:5: E303 too many blank lines (2)
lemur/plugins/lemur_fastly/plugin.py:99:45: E261 at least two spaces before inline comment
```

hosseinsh:
also long term, having testing would be of value

hosseinsh:
hi @alwaysjolley, are you still planing to work on this PR?

PR #2982, opened 2020-05-18
Fix intermediate CA creation on cryptography plugin

raul-te:
Creating an intermediate CA using the `cryptography` plugin fails with the message `"Private would be ignored, authority key used instead"`. This pull request fixes that issue and adds a test to cover that use-case.

### DIFF
--- a/lemur/plugins/lemur_cryptography/plugin.py
+++ b/lemur/plugins/lemur_cryptography/plugin.py
@@ -24,7 +24,12 @@ from lemur.certificates.service import create_csr
 def build_certificate_authority(options):
     options["certificate_authority"] = True
     csr, private_key = create_csr(**options)
-    cert_pem, chain_cert_pem = issue_certificate(csr, options, private_key)
+
+    if options.get("parent"):
+        # Intermediate Cert Issuance
+        cert_pem, chain_cert_pem = issue_certificate(csr, options, None)
+    else:
+        cert_pem, chain_cert_pem = issue_certificate(csr, options, private_key)
 
     return cert_pem, private_key, chain_cert_pem
 
--- a/lemur/plugins/lemur_cryptography/tests/test_cryptography.py
+++ b/lemur/plugins/lemur_cryptography/tests/test_cryptography.py
@@ -25,6 +25,31 @@ def test_build_certificate_authority():
     assert chain_cert_pem == ""
 
 
+def test_build_intermediate_certificate_authority(authority):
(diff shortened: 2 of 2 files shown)


hosseinsh:
@raul-te thanks for the PR, would you be able to update this PR? thanks

raul-te:
@hosseinsh done. Thanks for reviewing.

lemur/plugins/lemur_cryptography/plugin.py
```diff
@@ -41,9 +41,10 @@ def issue_certificate(csr, options, private_key=None):
     if options.get("authority"):
         # Issue certificate signed by an existing lemur_certificates authority
         issuer_subject = options["authority"].authority_certificate.subject
-        assert (
-            private_key is None
-        ), "Private would be ignored, authority key used instead"
+        if "parent" not in options:
+            assert (
+                private_key is None
+            ), "Private would be ignored, authority key used instead"
```
hosseinsh, line 47: @raul-te, I am wondering if we shouldn't move this logic rather  to `def build_certificate_authority(options):`, since `private_key` would be anyways overwritten in case an authority is provided, and might make it more difficult to debug. Any thoughts?

for example
```
def build_certificate_authority(options):
    options["certificate_authority"] = True
    csr, private_key = create_csr(**options)
    
    if options.get("parent"):
        # Intermediate Cert Issuance
        cert_pem, chain_cert_pem = issue_certificate(csr, options, None)
    else:
        cert_pem, chain_cert_pem = issue_certificate(csr, options, private_key)

    return cert_pem, private_key, chain_cert_pem
```

hosseinsh:
@raul-te sure thing. Added my review. 
Also there might be a permission missing, not allowing the admins to update your PR.

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

lemur/plugins/lemur_cryptography/plugin.py
raul-te: Yes, that makes sense. I've implemented your suggestion.

hosseinsh: awesome, thank you.

raul-te:
@hosseinsh Thanks. I'm not sure what's the problem is, but I don't see any options for giving permissions to the admins. Anyway, I've implemented your suggestion.

hosseinsh:
Thanks again for the PR @raul-te. merging the PR.

PR #3001, opened 2020-06-03
fixing domain validation

alwaysjolley:
- existing logic tests total length is < 253 and then removes extraneous trailing '.', reordered logic to correctly test length after removing trailing '.'
- existing regex testing for 253 when it should be 63 per label, fixed
- existing regex allowed for invalid characters '_' in domain, fixed
- existing regex did not account for short TLDs like co.uk, de, jp, io, fixed


alwaysjolley:
> Hi @alwaysjolley, thanks for the PR.
> The 253 applies to FQDNs and respectively DNSnames included as SANs.
> The 63 limit applies to CN.
> However this method is called for all domains include in the CSR.

The 253 FQDN is tested for the whole string on line 36, the regex is testing each label in the zone. The only requests using this function are submitting the "zone" not a FQDN, in fact there is only request to this function in the current code base https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_acme/powerdns.py#L159
other plugins have defined their own version of this function and submit the FQDN but either way, the current function does not work for the reasons described above.

hosseinsh:
good point, thanks for the clarification. 
You are right, the 63 char limit is per label vs. 253 limit for the FQDN.

hosseinsh:
Thanks for the PR again, just merged it.

PR #3040, opened 2020-07-01
Raise ValidationError if CSR contains invalid CN

jramosf:
If we supply a CSR that contains an empty field in the Subject, Lemur will crash with an error 500 as the ValueError exception is not captured. This change captures the exception and raises a ValidationError which in this case is a 400 sent back to client. Example to reproduce:

    Subject: C=ZZ, ST=Something, L=, O=My_Org, OU=My_Dept, CN=www.booking.com

The empty L= causes a ValueError which needs to be captured.


hosseinsh:
good catch @jramosf, thanks for the PR.

PR #3041, opened 2020-07-02
Fix unmatched field in Authorization

jramosf:
The field in the formatted string was not matching the args


hosseinsh:
Thanks for the bugfix @jramosf!

PR #3042, opened 2020-07-02
Remove f from non-f string

jramosf:
Fix from flake8 linter


hosseinsh:
thanks for the PR @jramosf.

PR #3074, opened 2020-08-04
Certificate cloning enhancements/fixes

charhate:
Support added for

- Honor selected algorithm during certificate cloning, currently it defaults to RSA2048
- Do not inherit **replacement** configs during cert clone
- Include ID of the cert being cloned in description
> Cloning from cert ID 80764

### DIFF
--- a/lemur/static/app/angular/certificates/certificate/certificate.js
+++ b/lemur/static/app/angular/certificates/certificate/certificate.js
@@ -215,11 +215,14 @@ angular.module('lemur')
   $scope.certificate = LemurRestangular.restangularizeElement(null, {}, 'certificates');
   CertificateApi.get(editId).then(function (certificate) {
     $scope.certificate = certificate;
+    // prepare the certificate for cloning
     $scope.certificate.name = ''; // we should prefer the generated name
     $scope.certificate.csr = null;  // should not clone CSR in case other settings are changed in clone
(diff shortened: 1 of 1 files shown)


lemur/static/app/angular/certificates/certificate/certificate.js
```diff
@@ -212,12 +212,18 @@ angular.module('lemur')
...
 .controller('CertificateCloneController', function ($scope, $uibModalInstance, CertificateApi, CertificateService, DestinationService, AuthorityService, AuthorityApi, PluginService, MomentService, WizardHandler, LemurRestangular, NotificationService, toaster, editId) {
+  $scope.certificate = LemurRestangular.restangularizeElement(null, {}, 'certificates');
   CertificateApi.get(editId).then(function (certificate) {
     $scope.certificate = certificate;
+    // prepare the certificate for cloning
     $scope.certificate.name = ''; // we should prefer the generated name
     $scope.certificate.csr = null;  // should not clone CSR in case other settings are changed in clone
     $scope.certificate.validityStart = null;
     $scope.certificate.validityEnd = null;
+    $scope.certificate.keyType = 'RSA2048'; // default algo to show during clone
```
charhate, line 223: After the change from line 215, UI does not show RSA2048 be default, thus added it here.
This will just pre-populate the value. A user can change it from UI as needed.

PR #3081, opened 2020-08-11
Cert validity should not exceed 397 days for publicly trusted issuers

charhate:
Considering the limit of 397 days of validity on certs issued by Public Issuers like DigiCert ([Reference](https://knowledge.digicert.com/generalinformation/2-year_Certificate_Availability?om_ext_cId=dc_email_7010z000000ynaJAAQ)), Lemur will limit the validity of such certs to 397 days. The validation is added to data validation logic and changed in plugin code as well.

- This change will be applicable to public issuers listed using the new config named **PUBLIC_CA_AUTHORITY_NAMES**
- The validity can also be configured using **PUBLIC_CA_MAX_VALIDITY_DAYS**

### DIFF
--- a/lemur/common/validators.py
+++ b/lemur/common/validators.py
@@ -152,6 +152,18 @@ def dates(data):
                         data["authority"].authority_certificate.not_after
                     )
                 )
+            # Allow no more than PUBLIC_CA_MAX_VALIDITY_DAYS (Default: 397) days of validity
+            # for certs issued by public CA
+            # The list of public issuers can be managed through a config named PUBLIC_CA
+            public_CA = current_app.config.get("PUBLIC_CA", [])
+            if data["authority"].name.lower() in [ca.lower() for ca in public_CA]:
+                max_validity_days = current_app.config.get("PUBLIC_CA_MAX_VALIDITY_DAYS", 397)
+                if (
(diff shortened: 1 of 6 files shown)


lemur/common/validators.py
```diff
@@ -152,6 +152,18 @@ def dates(data):
                         data["authority"].authority_certificate.not_after
                     )
                 )
+            # Allow no more than PUBLIC_CA_MAX_VALIDITY_DAYS (Default: 397) days of validity
+            # for certs issued by public CA
+            # The list of public issuers can be managed through a config named PUBLIC_CA
+            public_CA = current_app.config.get("PUBLIC_CA", [])
```
hosseinsh, line 158: maybe `PUBLIC_CA_AUTHORITY_NAMES` to indicate what this config is expected to contain, and that it is a list.

hosseinsh:
Some consistency in terms of config with `determine_validity_years(years):` might be desired
https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_digicert/plugin.py#L63-L75

lemur/common/validators.py
charhate: Done

PR #3091, opened 2020-08-18
empty OU and date changes on UI

charhate:
1. DigiCert will [deprecate the Organizational Unit](https://knowledge.digicert.com/alerts/ou-removal) field (simply ignoring the field from the requests). To honor this change and make it transparent to users, Lemur will not render this field as empty in UI and make it optional form input

2. Considering the limit of 397 days of validity of certs issued by Public Issuers ( #3081 ), the date picker from UI is now changed to show dates only within the allowed range - PUBLIC_CA_MAX_VALIDITY_DAYS (default 397). For any other CA which is not configured as Public, Lemur will allow selecting dates within DEFAULT_MAX_VALIDITY_DAYS.


lemur/certificates/models.py
```diff
@@ -311,6 +311,16 @@ def validity_remaining(self):
     def validity_range(self):
         return self.not_after - self.not_before
 
+    @property
+    def max_issuance_date(self):
+        public_CA = current_app.config.get("PUBLIC_CA_AUTHORITY_NAMES", [])
+        if self.name.lower() in [ca.lower() for ca in public_CA]:
+            default_validity_days = current_app.config.get("PUBLIC_CA_MAX_VALIDITY_DAYS", 397)
+        else:
+            default_validity_days = current_app.config.get("INTERNAL_CA_MAX_VALIDITY_DAYS", 365)   # 1 Year
```
charhate, line 320: Rename to DEFAULT_MAX_VALIDITY_DAYS, default value should be 3years

PR #3130, opened 2020-09-10
Improved csr support and EC key_type support

hosseinsh:
This PR makes sure we respect cn and key_type as defined in the CSR, and avoid conflict during cert minting and chain verification. 

The  PR also expands on key_type support to include ECC

### DIFF
--- a/lemur/common/utils.py
+++ b/lemur/common/utils.py
@@ -114,6 +114,41 @@ def get_authority_key(body):
     return authority_key.hex()
 
 
+def get_key_type_from_ec_curve(curve_name):
+    """
+    Give an EC curve name, return the matching key_type.
+
+    :param: curve_name
+    :return: key_type
+    """
+
+    _CURVE_TYPES = {
+        ec.SECP192R1().name: "ECCPRIME192V1",
+        ec.SECP256R1().name: "ECCPRIME256V1",
+        ec.SECP192R1().name: "ECCSECP192R1",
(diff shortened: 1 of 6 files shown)


lemur/common/utils.py
```diff
@@ -128,11 +161,11 @@ def generate_private_key(key_type):
...
     _CURVE_TYPES = {
-        "ECCPRIME192V1": ec.SECP192R1(),
-        "ECCPRIME256V1": ec.SECP256R1(),
-        "ECCSECP192R1": ec.SECP192R1(),
+        "ECCPRIME192V1": ec.SECP192R1(),  # duplicate
+        "ECCPRIME256V1": ec.SECP256R1(),  # duplicate
+        "ECCSECP192R1": ec.SECP192R1(),  # duplicate
         "ECCSECP224R1": ec.SECP224R1(),
-        "ECCSECP256R1": ec.SECP256R1(),
+        "ECCSECP256R1": ec.SECP256R1(),  # duplicate
```
charhate, line 168: Can the comment `# duplicate` be removed now that duplicates are removed?

hosseinsh: the duplicates are not removed yet. 
I think in a separate PR, we should churn all not support curves, based on what public CAs typically support, for instance: 

https://github.com/letsencrypt/boulder/blob/74c313968063af51c63580e90a2b3bf342bc343b/cmd/ceremony/ecdsa.go#L22-L28

charhate (approved):

PR #3135, opened 2020-09-15
Column addition key_type

charhate:
This PR contains only the schema change at this point.


lemur/dns_providers/models.py
```diff
@@ -12,7 +12,7 @@ class DnsProvider(db.Model):
     __tablename__ = "dns_providers"
     id = Column(Integer(), primary_key=True)
     name = Column(String(length=256), unique=True, nullable=True)
-    description = Column(Text(), nullable=True)
+    description = Column(String(length=1024), nullable=True)
```
charhate, line 15: This is change is done to match the definition of description column with actual schema definition from DB. This delta surfaced while creating a migration file. 
VARCHAR offers variable max size as compared to Text. This seems to be working well for Lemur, thus continuing with the same than doing a db change.

lemur/certificates/models.py
```diff
@@ -297,6 +296,7 @@ def location(self):
     def distinguished_name(self):
         return self.parsed_cert.subject.rfc4514_string()
 
+    """
```
hosseinsh, line 299: I would add a comment here, maybe along the lines of to be removed, as the functionality is provided via key_type column.

hosseinsh (approved):

PR #3144, opened 2020-09-19
Generate secrets on the fly

hosseinsh:
To set a good code hygiene example, removing any example secrets or non-used secrets for unit tests.
instead we now generate secrets on the fly, where needed.


charhate (approved):
Changes look good (looks like there is a minor lint error which should be a quick fix)

PR #3150, opened 2020-09-23
Backfill the key_type column

charhate:
DB Upgrade with alembic

Sample logs-
```
*** Starting new run(2020-09-22 18:39:42.869481) ***
Processing certificate with key type RSA 1024
Query: update certificates set key_type='RSA1024' where bits=1024 and not_after > CURRENT_DATE - 31 and key_type is null
--- 0.22719907760620117 seconds ---
Processing certificate with key type RSA 2048
Query: update certificates set key_type='RSA2048' where bits=2048 and not_after > CURRENT_DATE - 31 and key_type is null
--- 13.873061895370483 seconds ---
Processing certificate with key type RSA 4096
Query: update certificates set key_type='RSA4096' where bits=4096 and not_after > CURRENT_DATE - 31 and key_type is null
--- 0.23476719856262207 seconds ---
Processing certificate - ID: 80019 key_type: ECCPRIME256V1
Error in processing certificate - ID: 294 Error: Unable to load certificate. See https://cryptography.io/en/latest/faq/#why-can-t-i-import-my-pem-file for more details.
Processing certificate - ID: 80018 key_type: ECCPRIME256V1
--- 0.8531050682067871 seconds ---
--- Total 15.188664674758911 seconds ---
```

### DIFF
--- a/lemur/common/utils.py
+++ b/lemur/common/utils.py
@@ -71,6 +71,23 @@ def parse_private_key(private_key):
     )
 
 
+def get_key_type_from_certificate(body):
+    """
+
+    Helper function to determine key type by pasrding given PEM certificate
+
+    :param body: PEM string
+    :return: Key type string
+    """
+    parsed_cert = parse_certificate(body)
+    if isinstance(parsed_cert.public_key(), rsa.RSAPublicKey):
+        return "RSA{key_size}".format(
+            key_size=parsed_cert.public_key().key_size
+        )
+    elif isinstance(parsed_cert.public_key(), ec.EllipticCurvePublicKey):
+        return get_key_type_from_ec_curve(parsed_cert.public_key().curve.name)
+
+
 def split_pem(data):
     """
     Split a string of several PEM payloads to a list of strings.
--- a/lemur/migrations/versions/c301c59688d2_.py
+++ b/lemur/migrations/versions/c301c59688d2_.py
@@ -87,7 +87,7 @@ def update_key_type():
     conn = op.get_bind()
     start_time = time.time()
 
-    # Loop through all certificates are valid today or expired in last 30 days
+    # Loop through all certificates that are valid today or expired in the last 30 days.
     for cert_id, body in conn.execute(
             text(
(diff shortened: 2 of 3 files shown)


lemur/common/utils.py
```diff
@@ -71,6 +71,23 @@ def parse_private_key(private_key):
...
+    """
+    parsed_cert = parse_certificate(body)
+    if isinstance(parsed_cert.public_key(), rsa.RSAPublicKey):
+        return "RSA{key_size}".format(
+            key_size=parsed_cert.public_key().key_size
+        )
+    elif isinstance(parsed_cert.public_key(), ec.EllipticCurvePublicKey):
+        return get_key_type_from_ec_curve(parsed_cert.public_key().curve.name)
+
+
```
charhate, line 90: Added this method to utils as it will be used by lemur application code in future

hosseinsh: unit testing here would be awesome, for RSA/ECC certs.
https://github.com/Netflix/lemur/blob/master/lemur/tests/test_utils.py

lemur/migrations/versions/c301c59688d2_.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+    Helper methods performing updates for RSA and rest of the keys
+"""
+
+
+def update_key_type_rsa(bits):
+    log_file.write("Processing certificate with key type RSA %s\n" % bits)
+
+    stmt = text(
+        "update certificates set key_type='RSA{0}' where bits={0} and not_after > CURRENT_DATE - 31 and key_type is null".format(bits)
```
hosseinsh, line 75: better readability: 
```
f"update certificates set key_type='RSA{bits}' where bits={bits} and not_after > CURRENT_DATE - 31 and key_type is null"
```

hosseinsh (commented):
excellent PR. Thanks Sayali!

lemur/common/utils.py
charhate: Added [here](https://github.com/Netflix/lemur/pull/3150/files#diff-c09dceb332a94eea137a18f392af21aaR113-R116)

hosseinsh (approved):

PR #3151, opened 2020-09-23
ACME documentation improvement

### LINKED ISSUE #3149: Add documentation to reuse ACME accounts
Currently there is no documentation, on how to configure an ACME account, that is used for all certificates.

The current implementation creates a new account for each certificate request (as well as renewal), which can lead to a rate limit exhaustion.
In #3129 an additional issue was discovered that can be resolved using a preconfigured ACME account, but it's not entirely clear how to configure the `ACME_PRIVATE_KEY` and `ACME_REGR` properly.

It took me some time to figure out, how to properly set these two variables, so they work as intended, thus I propose to add a paragraph on how to configure it, and also how to convert a private key PEM to the JWK format required by the acme_client.

peschmae:
Adds documentation how to reuse an ACME account.

Additionally also adds the configuration parameters, to the configuration documentation.

Fixes #3149

### DIFF
--- a/docs/production/index.rst
+++ b/docs/production/index.rst
@@ -513,6 +513,8 @@ The following must be added to the config file to activate the pinning (the pinn
     """
 
 
+.. _AcmeAccountReuse:
+
 LetsEncrypt: Using a pre-existing ACME account
 -----------------------------------------------
 
--- a/docs/production/index.rst
+++ b/docs/production/index.rst
@@ -540,7 +540,7 @@ configuration.
     }
 
 
-Using `python-jwt` converting a existing private key in PEM format is quite easy::
+Using `python-jwt` converting an existing private key in PEM format is quite easy::
 
     import python_jwt as jwt, jwcrypto.jwk as jwk
 
@@ -554,4 +554,4 @@ Using `python-jwt` converting a existing private key in PEM format is quite easy
 
     {"body": {}, "uri": "https://acme-staging-v02.api.letsencrypt.org/acme/acct/<ACCOUNT_NUMBER>"}
 
-The uri can be retrieved from the ACME create account endpoint, when trying to create a new account, using the existing key.
+The URI can be retrieved from the ACME create account endpoint when creating a new account, using the existing key.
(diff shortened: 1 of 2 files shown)


docs/production/index.rst
```diff
@@ -503,3 +503,47 @@ The following must be added to the config file to activate the pinning (the pinn
...
+        "d": "llBlYhil3I.....",
+        "p": "-5LW2Lewogo.........",
+        "q": "zk6dHqHfHksd.........",
+        "dp": "qfe9fFIu3mu.......",
+        "dq": "cXFO-loeOyU.......",
+        "qi": "AfK1sh0_8sLTb..........."
+    }
+
+
+Using `python-jwt` converting a existing private key in PEM format is quite easy::
```
hosseinsh, line 535: `an existing`

docs/production/index.rst
```diff
@@ -503,3 +503,47 @@ The following must be added to the config file to activate the pinning (the pinn
...
+    ...
+    -----END RSA PRIVATE KEY-----""")
+
+    print(priv_key.export())
+
+`ACME_REGR` needs to be a valid JSON with a `body` and a `uri` attribute, similar to this::
+
+    {"body": {}, "uri": "https://acme-staging-v02.api.letsencrypt.org/acme/acct/<ACCOUNT_NUMBER>"}
+
+The uri can be retrieved from the ACME create account endpoint, when trying to create a new account, using the existing key.
```
hosseinsh, line 549:
```
The URI can be retrieved from the ACME create account endpoint when creating a new account, using the existing key.
``

hosseinsh (commented):
thanks for the PR. minor suggestions

peschmae:
Thanks for the feedback. I've integrated your suggestions

hosseinsh:
@peschmae thanks for the quick workaround. I don't seem to have the permissions, to update the branch to latest version, before being able to merge it.

peschmae:
@hosseinsh That's probably because I switched over from my private fork, to an organizational one. Need to check that tomorrow.

For now, I've rebased my branched to the master, and it should be up to date again

hosseinsh:
Thanks @peschmae for the PR!
Documentation is certainly an area, we need to improve.

PR #3153, opened 2020-09-23
Store ACME account details

### LINKED ISSUE #3152: ACME plugin should store account credentials per authority
Currently the ACME plugin, either reads the account details from `ACME_PRIVATE_KEY` and `ACME_REGR` configuration variables, or always creates a new account for each request.

This can lead to rate limit exhaustion if you try to create a lot of certificates at once, and also makes it impossible, to revoke a ACME certificate, since the revocation needs to be done using the same account as the certificate was issued for.

When creating a client, it tries to read the the private key and the registration resource from the authority
https://github.com/Netflix/lemur/blob/772894c414baca916c44ce1f91fb3fe6b33c88f1/lemur/plugins/lemur_acme/plugin.py#L237-L240

But since they aren't stored if a new account was created the first time around, this is futile.

After a new registration is done, it should store the credentials in the authority
https://github.com/Netflix/lemur/blob/772894c414baca916c44ce1f91fb3fe6b33c88f1/lemur/plugins/lemur_acme/plugin.py#L253-L265

peschmae:
As described in #3152 currently it's not possible to reuse a ACME account, and for each certificate a new account is created.
This can lead to rate limit exhaustion and also makes it impossible to revoke a certificate.

This pull request, implements a new option `store_account` for ACME authorities, which then stores the `acme_private_key`  as well as the `acme_regr` in the authority options.

Fixes #3152


lemur/plugins/lemur_acme/plugin.py
```diff
@@ -447,6 +471,13 @@ class ACMEIssuerPlugin(IssuerPlugin):
             "validation": "/^-----BEGIN CERTIFICATE-----/",
             "helpMessage": "Certificate to use",
         },
+        {
+            "name": "store_account",
+            "type": "bool",
+            "required": False,
+            "helpMessage": "Disable to create a new account for each ACME request",
+            "default": True,
```
hosseinsh, line 479: Shouldn't the default be `False`? 

```
            # if store_account is checked, add the private_key and registration resources to the options
            if options['store_account']:
```

peschmae: You're right. Having the default be `False` leaves the behaviour as is, otherwise we would change the default behaviour.

peschmae:
Rebased changes to master, and force pushed to avoid merge commits

hosseinsh:
@peschmae looks good to me. I don't seem to have permissions to update the branch, if you can do it would be great.

hosseinsh (approved):

hosseinsh:
Thanks @peschmae for the PR. just merged it.

PR #3166, opened 2020-09-29
Fix dns-providers type missing from schema

### LINKED ISSUE #3165: DNS Providers list doesn't  show type
In the DNS Providers list, there is a column for the provider type, but it's always empty.

Looking at the code, and the API requests, the issue seems to be with the dns_providers API call, which returns the list of all providers.

There should be a providerType value in the JSON, but it's not there. 

A quick glance at the `DnsProvidersNestedOutputSchema` shows that the value is called `providerType`, but in the database the field is called `provider_type` similar to `api_endpoint` which is called `api_endpoint` in the OutputSchema, so I guess, it's probably just mislabeled in the OutputSchema, and needs to be adjusted there, and maybe in the angular template.

peschmae:
Fixes the missing provider_type in the DNS providers list

Fixes #3165

### DIFF
--- a/lemur/dns_providers/schemas.py
+++ b/lemur/dns_providers/schemas.py
@@ -8,7 +8,7 @@ class DnsProvidersNestedOutputSchema(LemurOutputSchema):
     __envelope__ = False
     id = fields.Integer()
     name = fields.String()
-    providerType = fields.String()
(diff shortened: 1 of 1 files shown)


charhate (approved):
Thank you for the fix.

PR #3168, opened 2020-09-29
Remove bit length check from last query

charhate:
This change will include valid keys with bits set to null during the backfill.
-------------------------------------------------
Upgrade script initially modifies the key_type info of the keys with bit length 1024, 2048 and 4096. So remaining keys with _key_type=null_ are indeed of length <1024. In that sense this check is redundant and it is not considering keys with bits set to null

### DIFF
--- a/lemur/migrations/versions/c301c59688d2_.py
+++ b/lemur/migrations/versions/c301c59688d2_.py
@@ -90,7 +90,7 @@ def update_key_type():
     # Loop through all certificates that are valid today or expired in the last 30 days.
     for cert_id, body in conn.execute(
             text(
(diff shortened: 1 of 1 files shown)


hosseinsh:
how we handle the case when bit length is set to null and the cert is of type RSA?

https://github.com/Netflix/lemur/blob/59bfcec8085569a9286581c1bbe19b75a2a96f14/lemur/common/utils.py#L82-L88

charhate:
This method is looking at parsed certificate and not the bits column from DB, so I believe we it would work.
`key_size=parsed_cert.public_key().key_size `

hosseinsh (approved):

PR #3174, opened 2020-10-07
Authority create: Email added to subject DN

charhate:
Added Email to distinguished name options during CA creation.
The owner email will be copied over to Email filed by default. It can be modified to be empty before submitting the form.

### DIFF
--- a/lemur/authorities/schemas.py
+++ b/lemur/authorities/schemas.py
@@ -50,6 +50,8 @@ class AuthorityInputSchema(LemurInputSchema):
         missing=lambda: current_app.config.get("LEMUR_DEFAULT_COUNTRY")
     )
     state = fields.String(missing=lambda: current_app.config.get("LEMUR_DEFAULT_STATE"))
+    # Creating a String field instead of Email to allow empty value
+    email = fields.String()
 
     plugin = fields.Nested(PluginInputSchema)
 
--- a/lemur/authorities/schemas.py
+++ b/lemur/authorities/schemas.py
@@ -43,9 +43,7 @@ class AuthorityInputSchema(LemurInputSchema):
     organization = fields.String(
         missing=lambda: current_app.config.get("LEMUR_DEFAULT_ORGANIZATION")
     )
-    location = fields.String(
-        missing=lambda: current_app.config.get("LEMUR_DEFAULT_LOCATION")
-    )
+    location = fields.String()
     country = fields.String(
         missing=lambda: current_app.config.get("LEMUR_DEFAULT_COUNTRY")
     )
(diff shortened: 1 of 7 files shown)


hosseinsh (approved):
did some testing locally looks good.

Would it make sense to also allow Location to be optional within this PR?

Currently it complains with:
`You must enter a location`

charhate:
Sure, I can do that change if we want to go ahead with Location as optional field. I will push the changes to this PR

lemur/authorities/schemas.py
```diff
@@ -43,13 +43,13 @@ class AuthorityInputSchema(LemurInputSchema):
     organization = fields.String(
         missing=lambda: current_app.config.get("LEMUR_DEFAULT_ORGANIZATION")
     )
-    location = fields.String(
-        missing=lambda: current_app.config.get("LEMUR_DEFAULT_LOCATION")
-    )
+    location = fields.String()
```
intgr, line 46: What...

It was always possible to make `location` optional if you set `LEMUR_DEFAULT_LOCATION = ""` or `LEMUR_DEFAULT_LOCATION = None` in your settings.

But now you've removed the option of using `LEMUR_DEFAULT_LOCATION` setting entirely. In a PR titled "Email added to subject DN", which has no indication that it may change behaviof or the "location" field.

The **documentation and example config files** still mention `LEMUR_DEFAULT_LOCATION` in several places. It's still a **required setting** in `manage.py`.... But it no longer does anything?!

Sorry to be blunt, but the way Lemur merges pull requests that break backward compatibility in API behavior is reckless. It's starting to feel like any benefits of Lemur are greatly outweighed by the time required to test for and debug surprise regressions from every update.

Oh well, one more hack to add to my pile.

hosseinsh: Hey Marti, sorry that this change has caused trouble on your end, and thank you for brining this to our attention, and also for the feedback.

Our intention is to maintain backward compatibility, but at times, we oversee the implications of a simple change. 
Will be chatting with the team, and raise a PR to fix this issue.

charhate: Hi Marti, the erroneous change is now reverted in #3522. Thanks for reporting this and apologies for the inconvenience caused.
Lemur was still using the value of `LEMUR_DEFAULT_LOCATION` to populate defaults in UI. Thus it was not removed from documentation and config. There was an issue in API honoring the empty location value, if needed, and as part of fixing that this erroneous change got introduced. Nevertheless, this PR should have called it out explicitly and also the documentation. I hope the fix unblocks you.

Please let us know of any other issues that you would like to bring to our attention. We definitely do not intend Lemur users to put hack so as to make code work.

PR #3180, opened 2020-10-10
Certificate edit: update role and notification with owner change

charhate:
If the owner of an existing certificate changes, this code will now add the needed role for the new owner and remove the tole associated with old owner (there is only one certificate owner). This will allow the new owner to make change to the certificate.
Similarly it will add default notifications for new owner.
This PR also fixes the bug where removing a default notification for current owner was not possible. e.g. if someone wanted remove 30-day notification and keep only 15 and 2 day, the code will now honor the change.


lemur/certificates/service.py
```diff
@@ -256,6 +256,14 @@ def update(cert_id, **kwargs):
     return database.update(cert)
 
 
+def cleanup_owner_roles_notification(owner_name, kwargs):
+    kwargs["roles"] = [r for r in kwargs["roles"] if r.name != owner_name]
+    notification_prefix = "DEFAULT_{0}".format(
```
hosseinsh, line 261:
```
notification_prefix = f"DEFAULT_{owner_name.split("@")[0].upper()}"
```

lemur/certificates/service.py
```diff
@@ -268,16 +276,13 @@ def update_notify(cert, notify_flag):
...
-    if not owner_role:
-        owner_role = role_service.create(
-            kwargs["owner"],
-            description="Auto generated role based on owner: {0}".format(
-                kwargs["owner"]
-            ),
+    # create a role for the owner and assign it
+    owner_role = role_service.get_or_create(
+        kwargs["owner"],
+        description="Auto generated role based on owner: {0}".format(
```
hosseinsh, line 282:
```
        description = f"Auto generated role based on owner: {kwargs["owner"]}"
```

lemur/certificates/schemas.py
```diff
@@ -184,25 +185,52 @@ def load_data(self, data):
...
+
+        if data.get("owner"):
+            # Check if role already exists. This avoids adding duplicate role.
+            if data.get("roles") and any(r.get("name") == data["owner"] for r in data["roles"]):
+                return data
+
+            # Add required role
+            owner_role = roles_service.get_or_create(
+                data["owner"],
+                description="Auto generated role based on owner: {0}".format(data["owner"])
```
hosseinsh, line 197:
```
description = f"Auto generated role based on owner: {kwargs["owner"]}"
```

lemur/certificates/schemas.py
```diff
@@ -184,25 +185,52 @@ def load_data(self, data):
...
+
         return data
 
     @post_load
     def enforce_notifications(self, data):
         """
-        Ensures that when an owner changes, default notifications are added for the new owner.
-        Old owner notifications are retained unless explicitly removed.
+        Add default notification for current owner if none exist.
+        This ensures that the default notifications are added in the even of owner change.
```
hosseinsh, line 213: **event** of owner change.

lemur/certificates/schemas.py
```diff
@@ -184,25 +185,52 @@ def load_data(self, data):
...
+        Old owner notifications are retained unless explicitly removed later in the code path.
```
hosseinsh, line 214: is this still the case?

charhate: This method is loading schema and does not have knowledge of old vs. new owner. The removal of roles, if needed, needs to be handled explicitly. Thus slightly modified the comment to say
> unless explicitly removed later in the code path

Let know if its not clear though

hosseinsh (approved):

PR #3187, opened 2020-10-14
cab_compliant option in authority

charhate:
Using options field of Authority, store a boolean value for option `cab_compliant`.
For non-cab-compliant issuers, copy subject details during the cloning instead of using default values.

Additionally, the PR has some minor changes
1. code cleanup to remove OU from digicert plugin + unit test changes
2. set default algo in `CertificateService.getDefaults`


charhate:
I will rebase and raise separate PR

PR #3188, opened 2020-10-14
Cab compliant

charhate:
Using options field of Authority, store a boolean value for option `cab_compliant` in database.
`{"name": "cab_compliant", "type": "boolean", "value": true, "helpMessage": "Define if this issuer is CAB Compliant"}`
For non-cab-compliant issuers, copy subject details during the cloning instead of using default values.
This PR does not cover any frontend changes to set this option

Additionally, the PR has some minor changes
1. code cleanup to remove OU from digicert plugin + unit test changes
2. set default algo in `CertificateService.getDefaults`

### DIFF
--- a/lemur/authorities/models.py
+++ b/lemur/authorities/models.py
@@ -85,7 +85,7 @@ class Authority(db.Model):
     @property
     def is_cab_compliant(self):
         """
-        Parse the options to find whether authority is CAB Forum Compliant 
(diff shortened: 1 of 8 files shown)


hosseinsh (commented):
looks good to me. 
Is there a chance we could add testing for `handle_subject_details` to cover the three different scenarios.

hosseinsh (approved):

PR #3190, opened 2020-10-15
Use cab_compliant option instead of authority name list

charhate:
Removed PUBLIC_CA_AUTHORITY_NAMES and instead using option cab_compliant for authorities ( introduced in #3188 )

### DIFF
--- a/lemur/authorities/schemas.py
+++ b/lemur/authorities/schemas.py
@@ -111,8 +111,6 @@ class RootAuthorityCertificateOutputSchema(LemurOutputSchema):
     cn = fields.String()
     not_after = fields.DateTime()
     not_before = fields.DateTime()
-    max_issuance_days = fields.Integer()
-    default_validity_days = fields.Integer()
     owner = fields.Email()
     status = fields.Boolean()
     user = fields.Nested(UserNestedOutputSchema)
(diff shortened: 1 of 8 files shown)


lemur/certificates/models.py
```diff
@@ -317,20 +317,6 @@ def validity_remaining(self):
     def validity_range(self):
         return self.not_after - self.not_before
 
-    @property
```
hosseinsh, line 320: nothing better to see unused code being cleaned up

hosseinsh (approved):

lemur/authorities/schemas.py
```diff
@@ -138,8 +138,10 @@ class AuthorityNestedOutputSchema(LemurOutputSchema):
     owner = fields.Email()
     plugin = fields.Nested(PluginOutputSchema)
     active = fields.Boolean()
-    authority_certificate = fields.Nested(RootAuthorityCertificateOutputSchema, only=["max_issuance_days", "default_validity_days"])
+    authority_certificate = fields.Nested(RootAuthorityCertificateOutputSchema, only=["not_after", "not_before"])
```
charhate, line 141: "not_after", "not_before" were missing in the list, thus added now. Not related to current change

PR #3192, opened 2020-10-16
Expanding the S3 plugin with acme-http01-token support

hosseinsh:
The S3 Destination plugin can be employed for uploading acme-http01 to a defined S3 bucket. 
For this domain validation to work the domain to be validated needs to redirect to the  respective S3 bucket/path.

https://letsencrypt.org/docs/challenge-types/#http-01-challenge
```
http://<YOUR_DOMAIN>/.well-known/acme-challenge/<TOKEN>
```
The S3 bucket will be hosting `<TOKEN>`, where the file name is a uid, and the body of the file contains the expected challenge.

### DIFF
--- a/lemur/acme_providers/cli.py
+++ b/lemur/acme_providers/cli.py
@@ -128,7 +128,16 @@ def dnstest(domain, token):
     help="Bucket Name",
 )
 def upload_acme_token_s3(token, token_name, prefix, account_number, bucket_name):
-
+    """
+    This method serves for testing the upload_acme_token to S3, fetching the token to verify it, and then deleting it.
+    It mainly serves for testing purposes.
+    :param token:
+    :param token_name:
+    :param prefix:
+    :param account_number:
+    :param bucket_name:
+    :return:
+    """
     additional_options = [
         {
             "name": "bucket",
--- a/lemur/acme_providers/cli.py
+++ b/lemur/acme_providers/cli.py
@@ -10,7 +10,6 @@ from lemur.constants import SUCCESS_METRIC_STATUS
 from lemur.plugins import plugins
 from lemur.plugins.lemur_acme.plugin import AcmeHandler
 from lemur.plugins.lemur_aws import s3
-from lemur.utils import get_random_secret
 
 manager = Manager(
     usage="Handles all ACME related tasks"
--- a/lemur/acme_providers/cli.py
+++ b/lemur/acme_providers/cli.py
@@ -1,12 +1,16 @@
 import time
 import json
+import arrow
 
 from flask_script import Manager
 from flask import current_app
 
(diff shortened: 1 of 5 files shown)


hosseinsh:
@peschmae this is building on the acme-http01 work. Appreciate your input.

lemur/plugins/lemur_aws/s3.py
```diff
@@ -32,4 +35,37 @@ def put(bucket_name, region, prefix, data, encrypt, **kwargs):
...
+    """
+    try:
+        response = kwargs["client"].delete_object(Bucket=bucket_name, Key=prefix)
+        current_app.logger.debug(f"Delete data from S3."
+                                 f"Bucket: {bucket_name},"
+                                 f"Prefix: {prefix},"
+                                 f"Status_code: {response}")
+        return response['ResponseMetadata']['HTTPStatusCode'] < 300
+    except ClientError:
+        sentry.captureException()
```
charhate, line 57: Should it return false or some error after capturing exception. Considering we return boolean in try, the caller may expect some return?

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -406,3 +406,30 @@ def upload(self, name, body, private_key, chain, options, **kwargs):
...
+        filename = token_path.split("/")[-1]
+        if not prefix.endswith("/"):
+            prefix + "/"
+
+        s3.put(bucket_name=bucket_name,
+               region_name=region,
+               prefix=prefix + filename,
+               data=token,
+               encrypt=False,
+               account_number=account_number)
```
charhate, line 435: Would any success log add value?

peschmae:
I thought about adding a checkpoint, for the acme token upload. This might be something useful, but I'm not sure about success log.

It really depends on how you use it, and maybe it would be best to add both (a success log, as well as a metric)

lemur/plugins/lemur_aws/plugin.py
hosseinsh: certainly

lemur/acme_providers/cli.py
```diff
@@ -84,3 +87,105 @@ def dnstest(domain, token):
...
+    :param bucket_name:
+    :return:
+    """
+    additional_options = [
+        {
+            "name": "bucket",
+            "value": bucket_name,
+            "type": "str",
+            "required": True,
+            "validation": "[0-9a-z.-]{3,63}",
```
charhate, line 147: May be we can use `r""` regex notation to avoid getting warning in future for any escape characters
e.g. https://github.com/Netflix/lemur/blob/master/lemur/notifications/service.py#L46. Feel free to skip if that does not look necessary.

charhate (approved):

PR #3193, opened 2020-10-16
Miscellaneous notification fixes and tests

jtschladen:
This changes includes the following fixes:
- If multiple notifications were associated with a certificate, but _any_ of them was disabled, Lemur would stop iterating through the others, which could result in some notifications not being sent when they should have been.
- Rotation notifications and failed rotation notifications now work (only available via email). This involved a number of other changes, including some message template modifications.

Other changes:
- We now log when expiration notifications fail to be sent (previously the exception was swallowed, though metrics were sent).
- Added tests for messaging (in general) and email. Added a sample `LEMUR_EMAIL` in the config to support this.

### DIFF
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -101,6 +101,9 @@ def send_notification(event_type, data, targets, notification):
         notification.plugin.send(event_type, data, targets, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
+        current_app.logger.error(
+            "Unable to send notification to {}.".format(targets), exc_info=True
+        )
         sentry.captureException()
 
     metrics.send(
@@ -190,13 +193,13 @@ def send_rotation_notification(certificate, notification_plugin=None):
     status = FAILURE_METRIC_STATUS
     if not notification_plugin:
         notification_plugin = plugins.get(
-            current_app.config.get("LEMUR_DEFAULT_NOTIFICATION_PLUGIN")
+            current_app.config.get("LEMUR_DEFAULT_NOTIFICATION_PLUGIN", "email-notification")
         )
 
     data = certificate_notification_output_schema.dump(certificate).data
 
     try:
-        notification_plugin.send("rotation", data, [data["owner"]])
+        notification_plugin.send("rotation", data, [data["owner"]], [])
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
         current_app.logger.error(
@@ -290,7 +293,7 @@ def needs_notification(certificate):
 
     for notification in certificate.notifications:
         if not notification.active or not notification.options:
-            return
+            continue
 
         interval = get_plugin_option("interval", notification.options)
         unit = get_plugin_option("unit", notification.options)
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -101,8 +101,8 @@ def send_notification(event_type, data, targets, notification):
     log_data = {
         "function": function,
         "message": "Sending expiration notification for to targets {}".format(targets),
-        "notification_type": "rotation",
-        "targets": targets,
+        "notification_type": "expiration",
+        "certificate_targets": targets,
     }
     status = FAILURE_METRIC_STATUS
     try:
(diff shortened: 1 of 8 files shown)


lemur/notifications/messaging.py
```diff
@@ -101,6 +101,9 @@ def send_notification(event_type, data, targets, notification):
         notification.plugin.send(event_type, data, targets, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
+        current_app.logger.error(
```
hosseinsh, line 104: we have learned that structural data allows for better search and sorting, when looking at the logs, for instance via  kibana. 
This is the style we have been adopting
https://github.com/Netflix/lemur/blob/b96657994aae639c626a7501daaf3822643bde99/lemur/common/celery.py#L455-L472

lemur/notifications/messaging.py
```diff
@@ -290,7 +293,7 @@ def needs_notification(certificate):
 
     for notification in certificate.notifications:
         if not notification.active or not notification.options:
-            return
```
hosseinsh, line 293: that's is a good catch!

hosseinsh (commented):
just a minor improvement for the logging structure, to benefit from search/sort features.

lemur/notifications/messaging.py
jtschladen: I updated the log messages in this file to match that pattern. I'm not sure if there are any particular guidelines on which attributes we structure out, but I added ones that seemed useful.

lemur/notifications/messaging.py
```diff
@@ -96,11 +97,20 @@ def send_notification(event_type, data, targets, notification):
     :param notification:
     :return:
     """
+    function = f"{__name__}.{sys._getframe().f_code.co_name}"
+    log_data = {
+        "function": function,
+        "message": "Sending expiration notification for to targets {}".format(targets),
```
charhate, line 103: This is optional, instead of `format()` I have found `f"` notation to be more concise and readable for string formatting. Same as the one from line 100 ([here](https://github.com/Netflix/lemur/blob/master/lemur/certificates/schemas.py#L197) is another example). It was suggested to me by Hossein in my PR.

lemur/notifications/messaging.py
```diff
@@ -225,6 +243,14 @@ def send_pending_failure_notification(
     :param notification_plugin:
     :return:
     """
+    function = f"{__name__}.{sys._getframe().f_code.co_name}"
+    log_data = {
+        "function": function,
+        "message": "Sending pending failure notification for pending certificate {}".format(pending_cert.name),
+        "notification_type": "failed",
```
charhate, line 250: How about "pending failed" or "pending certificate failed" as notification_type, something more specific?

jtschladen: I used `failed` as it matches the name of the notification type sent to the plugin. I agree it's not the most descriptive name, but I'd also suggest it would be better to keep the name of the notification type consistent everywhere. Maybe we could look into renaming `failed` to something more specific later? I anticipate more notification refactor work coming soon(-ish).

lemur/notifications/messaging.py
```diff
@@ -96,11 +97,20 @@ def send_notification(event_type, data, targets, notification):
...
+        "notification_type": "expiration",
+        "certificate_targets": targets,
+    }
     status = FAILURE_METRIC_STATUS
     try:
         notification.plugin.send(event_type, data, targets, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
+        log_data["message"] = "Unable to send expiration notification to targets {}".format(targets)
+        current_app.logger.error(log_data, exc_info=True)
```
hosseinsh, line 113: this might be something worthwhile alerting, and adding metrics would facilitate that, similar to:
https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L552-L553

lemur/notifications/messaging.py
hosseinsh: yeah, this kinda makes for a good readability, and we have been adopting this approach in the last 2+ years, despite a lot of legacy code still being in the old format.

lemur/notifications/messaging.py
jtschladen: For the record: I spoke to Hossein, and we've confirmed there is already metric info for failed notifications a few lines below this, so no changes needed.

lemur/notifications/messaging.py
jtschladen: Yeah, it's a much better style. I updated all the strings in this file to use that.

charhate (approved):

hosseinsh (approved):

PR #3194, opened 2020-10-17
Show only few supported ECC algorithms on UI

charhate:
Not all the ECC curves are supported by CAs, thus trimming down the list of algorithms to few applicable ones. These are displayed on UI during certificate and authority creation. There is no backend change, so direct REST calls would work the same.


hosseinsh (approved):

charhate:
Something went wrong with re-base, I will raise another PR with same changes.

PR #3201, opened 2020-10-20
Support for SNS Expiration Notifications

jtschladen:
This change adds support for expiration notifications sent via AWS SNS. Topics are configured on a per-notification basis. Owner/security notifications will still be sent via email.

Here's a screenshot of the notification configuration:
Screen Shot 2020-10-20 at 12 05 10 PM

### DIFF
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -107,6 +107,7 @@ def send_plugin_notification(event_type, data, recipients, notification):
     }
     status = FAILURE_METRIC_STATUS
     try:
+        current_app.logger.debug(log_data)
         notification.plugin.send(event_type, data, recipients, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
@@ -203,6 +204,7 @@ def send_default_notification(notification_type, data, targets, notification_opt
     )
 
     try:
+        current_app.logger.debug(log_data)
         # we need the notification.options here because the email templates utilize the interval/unit info
         notification_plugin.send(notification_type, data, targets, notification_options)
         status = SUCCESS_METRIC_STATUS
@@ -288,8 +290,6 @@ def needs_notification(certificate):
             raise Exception(
                 f"Invalid base unit for expiration interval: {unit}"
             )
-        print(f"Does cert {certificate.name} need a notification {notification.label}? Actual: {days}, "
-              f"configured: {interval}")  # TODO REMOVE
         if days == interval:
             notifications.append(notification)
     return notifications
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -111,7 +111,7 @@ def send_plugin_notification(event_type, data, recipients, notification):
         notification.plugin.send(event_type, data, recipients, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
-        log_data["message"] = f"Unable to send expiration notification to recipients {recipients}"
+        log_data["message"] = f"Unable to send {event_type} notification to recipients {recipients}"
         current_app.logger.error(log_data, exc_info=True)
         sentry.captureException()
 
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -108,7 +108,7 @@ def send_plugin_notification(event_type, data, recipients, notification):
     }
     status = FAILURE_METRIC_STATUS
     try:
-        current_app.logger.debug(log_data)
+        current_app.logger.info(log_data)
         notification.plugin.send(event_type, data, recipients, notification.options)
         status = SUCCESS_METRIC_STATUS
     except Exception as e:
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -42,6 +42,7 @@ def get_certificates(exclude=None):
         .filter(Certificate.not_after <= max)
         .filter(Certificate.notify == True)
         .filter(Certificate.expired == False)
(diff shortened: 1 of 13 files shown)


lemur/plugins/lemur_aws/plugin.py
```diff
@@ -406,3 +407,54 @@ def upload(self, name, body, private_key, chain, options, **kwargs):
...
+
+        topic_arn = f"arn:aws:sns:{self.get_option('region', options)}:" \
+                    f"{self.get_option('accountNumber', options)}:" \
+                    f"{self.get_option('topicName', options)}"
+
+        current_app.logger.debug(f"Publishing {notification_type} notification to topic {topic_arn}")
+        try:
+            sns.publish(topic_arn, message, notification_type, region_name=self.get_option("region", options))
+        except Exception:
+            current_app.logger.exception(f"Error publishing {notification_type} notification to topic {topic_arn}")
```
charhate, line 460: How about capturing to sentry as well?
`sentry.captureException()`

lemur/notifications/messaging.py
```diff
@@ -30,7 +30,7 @@
 
 def get_certificates(exclude=None):
     """
-    Finds all certificates that are eligible for notifications.
+    Finds all certificates that are eligible for expiration notifications.
```
charhate, line 33: One thought about the query in this method, we should [add filter](https://github.com/Netflix/lemur/blob/master/lemur/notifications/messaging.py#L44) for `Certificate.revoked == True` as well.

lemur/plugins/lemur_aws/plugin.py
jtschladen: You make a good point, but I actually think I need to remove this try/catch instead. `messaging.py` is the only caller of this `send` method, and it has its own try/catch surrounding it that already alerts on failure. I'll remove it here - good catch, thanks.

lemur/notifications/messaging.py
jtschladen: Makes sense, no reason to send any notifications for revoked certs.

charhate (approved):

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -406,3 +407,51 @@ def upload(self, name, body, private_key, chain, options, **kwargs):
...
+        """
+        While we receive a `targets` parameter here, it is unused, as the SNS topic is pre-configured in the
+        plugin configuration, and can't reasonably be changed dynamically.
+        """
+
+        topic_arn = f"arn:aws:sns:{self.get_option('region', options)}:" \
+                    f"{self.get_option('accountNumber', options)}:" \
+                    f"{self.get_option('topicName', options)}"
+
+        current_app.logger.debug(f"Publishing {notification_type} notification to topic {topic_arn}")
```
hosseinsh, line 456: I wonder if this shouldn't be info

lemur/plugins/lemur_aws/sns.py
```diff
@@ -0,0 +1,54 @@
...
+    return "https://{hostname}/#/certificates/{name}".format(
+        hostname=current_app.config.get("LEMUR_HOSTNAME"), name=name
+    )
+
+
+def format_message(certificate, notification_type):
+    json_message = {
+        "notification_type": notification_type,
+        "certificate_name": certificate["name"],
+        "expires": arrow.get(certificate["validityEnd"]).format("dddd, MMMM D, YYYY"),
```
hosseinsh, line 50: Do you think a more machine-friendly format would be valuable?
like:
```
yyyy'-'MM'-'dd'T'HH':'mm':'ss | 2016-03-22T06:30:07
```

lemur/notifications/messaging.py
```diff
@@ -87,29 +89,30 @@ def get_eligible_certificates(exclude=None):
...
-        "message": f"Sending expiration notification for to targets {targets}",
+        "message": f"Sending expiration notification for to recipients {recipients}",
         "notification_type": "expiration",
-        "certificate_targets": targets,
+        "certificate_targets": recipients,
     }
     status = FAILURE_METRIC_STATUS
     try:
-        notification.plugin.send(event_type, data, targets, notification.options)
+        current_app.logger.debug(log_data)
```
hosseinsh, line 111: should this be info?

hosseinsh (commented):
an elaborate and well-tested PR. added some minor suggestions. 
It looks good to me. 
if I can find another 20min, I would take a more detailed look at `messaging.py`, otherwise, everything looks good to me.

lemur/plugins/lemur_aws/plugin.py
jtschladen: I could imagine this being a high volume log if there are many of these notifications configured, but it is also pretty useful to have logged. I'll go ahead and switch it (as well as the Slack equivalent) to info, and we can tune it back down later if needed.

lemur/plugins/lemur_aws/sns.py
jtschladen: Sure - I copied the message from the Slack plugin, but that does sound like a more useful format.

lemur/notifications/messaging.py
jtschladen: Same concern about volume (but even more than SNS) - I'm imagining there could be dozens/hundreds/maybe thousands of email notifications sent at once, which could be more logging than is actually useful. That said, if you don't think it's a valid concern, I can change it.

hosseinsh: I see your point. This might be way higher volume than the SNS part.

hosseinsh (approved):

PR #3202, opened 2020-10-20
Fix cert reissue when L/OU is not set

charhate:
get_certificate_primitives complains for None Location or Organizational Unit. These are optional fields that were added to [CertificateOutputSchema](https://github.com/Netflix/lemur/pull/3188/files#diff-20fb502199efa396152865a53b5a740e34917598249c46509a9a59c4fc8c3c50R335-R339)

### DIFF
--- a/lemur/certificates/schemas.py
+++ b/lemur/certificates/schemas.py
@@ -353,6 +353,12 @@ class CertificateOutputSchema(LemurOutputSchema):
                 data.pop("organization", None)
                 data.pop("organizational_unit", None)
 
+        # Removing optional fields if None, else it complains in de-serialization
+        if "location" in data and data["location"] is None:
+            data.pop("location")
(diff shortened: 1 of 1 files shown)


lemur/certificates/schemas.py
```diff
@@ -353,6 +353,12 @@ def handle_subject_details(self, data):
                 data.pop("organization", None)
                 data.pop("organizational_unit", None)
 
+        # Removing optional fields if None, else it complains in de-serialization
+        if "location" in data and data["location"] is None:
```
hosseinsh, line 357: should this be expanded to all `key: None` in `data`

charhate: The other fields are mandatory, but it could be safe to remove None and rely on Input schema to handle as needed.

hosseinsh (approved):

PR #3204, opened 2020-10-21
do not create db_upgrade.log during migrations

GnunuX:
I'm trying to launch Lemur with a local user without any right to change Lemur code.

During migrations, the file "db_upgrade.log" is created in current directory by lemur/migrations/versions/c301c59688d2_.py.

In my case I have the error:

PermissionError: [Errno 13] Permission denied: 'db_upgrade.log'

In my patch I'm using common logger as 1db4f82bc780_.py migrations' script.

### DIFF
--- a/lemur/migrations/versions/c301c59688d2_.py
+++ b/lemur/migrations/versions/c301c59688d2_.py
@@ -7,8 +7,9 @@ the rest of the keys, the certificate body is parsed to determine
 the exact key_type information.
 
 Each individual DB change is explicitly committed, and the respective
-log is added to a file named db_upgrade.log in the current working
-directory. Any error encountered while parsing a certificate will
+log is added to a file configured in LOG_UPGRADE_FILE or, by default,
+to a file named db_upgrade.log in the current working directory.
+Any error encountered while parsing a certificate will
 also be logged along with the certificate ID. If faced with any issue
 while running this upgrade, there is no harm in re-running the upgrade.
 Each run processes only rows for which key_type information is not yet
--- a/lemur/migrations/versions/c301c59688d2_.py
+++ b/lemur/migrations/versions/c301c59688d2_.py
@@ -31,12 +31,25 @@ down_revision = '434c29e40511'
 
 from alembic import op
 from sqlalchemy.sql import text
-from lemur.common import utils
 import time
 import datetime
+from flask import current_app
 
-import logging
-log = logging.getLogger(__name__)
+from logging import Formatter, FileHandler, getLogger
+
+from lemur.common import utils
+
+log = getLogger(__name__)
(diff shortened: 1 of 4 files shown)


charhate:
@GnunuX Thank you for raising the PR and catching the permission issue. It's a valid concern. The current changes look fine to me. Just wanted to know whether these logs would go to Lemur application logs?
The logs could be large in size depending on number of keys processed. The intention of adding the logs to a separate file was to collocate this information. It will be easier to read, search and look for error.
Any thoughts on alternative file path? Or how about having both options, if issue is encountered while creating new file, then use the logger.

GnunuX:
Here is a version with an alternative logger for upgrade.

lemur/migrations/versions/c301c59688d2_.py
```diff
@@ -31,15 +31,28 @@
...
 import datetime
+from flask import current_app
+
+from logging import Formatter, FileHandler, getLogger
 
-log_file = open('db_upgrade.log', 'a')
+from lemur.common import utils
+
+log = getLogger(__name__)
+handler = FileHandler(current_app.config.get("LOG_UPGRADE_FILE", "db_upgrade.log"))
```
charhate, line 43: Thanks for making this change, much appreciated. It might be a good idea to mention about this config in the comment from top of this file ([line 10](https://github.com/Netflix/lemur/blob/master/lemur/migrations/versions/c301c59688d2_.py#L10)).
`log is added to a file named db_upgrade.log (can be overridden using LOG_UPGRADE_FILE config)`

lemur/migrations/versions/c301c59688d2_.py
```diff
@@ -95,9 +107,9 @@ def update_key_type():
         try:
             cert_key_type = utils.get_key_type_from_certificate(body)
         except ValueError as e:
-            log_file.write("Error in processing certificate - ID: %s Error: %s \n" % (cert_id, str(e)))
+            log.error("Error in processing certificate - ID: %s Error: %s \n" % (cert_id, str(e)))
         else:
-            log_file.write("Processing certificate - ID: %s key_type: %s\n" % (cert_id, cert_key_type))
+            log.error("Processing certificate - ID: %s key_type: %s\n" % (cert_id, cert_key_type))
```
charhate, line 112: This one should be `log.info`

lemur/migrations/versions/c301c59688d2_.py
GnunuX: OK

lemur/migrations/versions/c301c59688d2_.py
GnunuX: Oh! Sorry for this mistake!

charhate (approved):

PR #3209, opened 2020-10-23
Fix plugin field on notification edit

jtschladen:
Previously, the "Plugin" field of the notification edit modal would always be empty on load. If you switch to a different type and back (say, switch from Email to Slack and back to Email), all your existing configuration would be lost. This appears to have been due to some bad Javascript, which was essentially checking for the wrong field names. This PR corrects the bug, resulting in (1) the Plugin field being filled in when the modal is opened, and (2) switching back and forth between plugins not clearing all of the notification data.

I've verified that editing a notification still works fine (as does creating a new one).

Note: Javascript guidance very welcome, this is absolutely not my area of expertise.

### DIFF
--- a/lemur/destinations/schemas.py
+++ b/lemur/destinations/schemas.py
@@ -31,6 +31,9 @@ class DestinationOutputSchema(LemurOutputSchema):
     def fill_object(self, data):
         if data:
             data["plugin"]["pluginOptions"] = data["options"]
+            for option in data["plugin"]["pluginOptions"]:
+                if "export-plugin" in option["type"]:
+                    option["value"]["pluginOptions"] = option["value"]["plugin_options"]
         return data
 
 
--- a/lemur/destinations/service.py
+++ b/lemur/destinations/service.py
@@ -41,12 +41,14 @@ def create(label, plugin_name, options, description=None):
     return database.create(destination)
 
 
-def update(destination_id, label, options, description):
+def update(destination_id, label, plugin_name, options, description):
     """
     Updates an existing destination.
 
     :param destination_id:  Lemur assigned ID
     :param label: Destination common name
+    :param plugin_name:
+    :param options:
     :param description:
     :rtype : Destination
     :return:
@@ -54,6 +56,11 @@ def update(destination_id, label, options, description):
     destination = get(destination_id)
 
     destination.label = label
+    destination.plugin_name = plugin_name
+    # remove any sub-plugin objects before try to save the json options
+    for option in options:
+        if "plugin" in option["type"]:
+            del option["value"]["plugin_object"]
     destination.options = options
     destination.description = description
 
--- a/lemur/notifications/service.py
+++ b/lemur/notifications/service.py
@@ -104,12 +104,13 @@ def create(label, plugin_name, options, description, certificates):
     return database.create(notification)
 
 
-def update(notification_id, label, options, description, active, certificates):
+def update(notification_id, label, plugin_name, options, description, active, certificates):
     """
     Updates an existing notification.
 
     :param notification_id:
     :param label: Notification label
+    :param plugin_name:
     :param options:
     :param description:
     :param active:
@@ -120,6 +121,7 @@ def update(notification_id, label, options, description, active, certificates):
     notification = get(notification_id)
 
     notification.label = label
+    notification.plugin_name = plugin_name
     notification.options = options
     notification.description = description
     notification.active = active
--- a/lemur/sources/service.py
+++ b/lemur/sources/service.py
@@ -264,13 +264,14 @@ def create(label, plugin_name, options, description=None):
     return database.create(source)
 
 
-def update(source_id, label, options, description):
+def update(source_id, label, plugin_name, options, description):
     """
     Updates an existing source.
 
     :param source_id:  Lemur assigned ID
     :param label: Source common name
     :param options:
+    :param plugin_name:
     :param description:
     :rtype : Source
     :return:
@@ -278,6 +279,7 @@ def update(source_id, label, options, description):
     source = get(source_id)
 
     source.label = label
+    source.plugin_name = plugin_name
     source.options = options
     source.description = description
 
--- a/lemur/static/app/angular/destinations/destination/destination.js
+++ b/lemur/static/app/angular/destinations/destination/destination.js
@@ -52,19 +52,19 @@ angular.module('lemur')
           if (plugin.slug === $scope.destination.plugin.slug) {
             plugin.pluginOptions = $scope.destination.plugin.pluginOptions;
             $scope.destination.plugin = plugin;
-            _.each($scope.destination.plugin.pluginOptions, function (option) {
-              if (option.type === 'export-plugin') {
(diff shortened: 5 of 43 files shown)


lemur/static/app/angular/notifications/notification/notification.js
```diff
@@ -42,25 +42,15 @@ angular.module('lemur')
...
-    PluginService.getByType('notification').then(function (plugins) {
-      $scope.plugins = plugins;
-      _.each($scope.plugins, function (plugin) {
-        if (plugin.slug === $scope.notification.pluginName) {
-          plugin.pluginOptions = $scope.notification.notificationOptions;
-          $scope.notification.plugin = plugin;
-        }
-      });
-    });
-
```
charhate, line 63: PR looks good!
I was going through older PR to find if this code is needed or may be left error (I believe so). While doing that I came across code in [source.js](https://github.com/Netflix/lemur/blob/master/lemur/static/app/angular/sources/source/source.js#L44) which may need similar fix. Do you think we can fix it the same way?

jtschladen: That looks probable. Let me dig into it and see.

jtschladen: In testing that, I've discovered another issue: it appears to be impossible to ever change the plugin associated with either a notification or a source. The options can be updated, but not the plugin itself (and attempting to change it puts the notification/source into a bad state). I'll try to figure out how to fix that and update this PR.

jtschladen:
Fixing the issue @charhate noticed with sources and destinations ended up requiring significantly more changes. It should now be possible to update the plugin type and options for notifications, sources, and destinations. I'll add a little more detail to each file to explain the changes.

lemur/destinations/schemas.py
```diff
@@ -31,6 +31,9 @@ class DestinationOutputSchema(LemurOutputSchema):
     def fill_object(self, data):
         if data:
             data["plugin"]["pluginOptions"] = data["options"]
+            for option in data["plugin"]["pluginOptions"]:
+                if "export-plugin" in option["type"]:
+                    option["value"]["pluginOptions"] = option["value"]["plugin_options"]
```
jtschladen, line 36: Function: This change is required to convert the `plugin_options` field (what's store in the DB) to the `PluginOptions` field that the Javascript expects. For the top level plugin, the options are stored as `options`; this inner plugin isn't actually converted to a plugin object and therefore just stays as `plugin_options` in the DB.

This feels rather hardcoded to me, but I am unsure what better options we have. Anyone more familiar with this is welcome to make suggestions.

lemur/destinations/service.py
```diff
@@ -41,19 +41,26 @@ def create(label, plugin_name, options, description=None):
     return database.create(destination)
 
 
-def update(destination_id, label, options, description):
+def update(destination_id, label, plugin_name, options, description):
```
jtschladen, line 44: Function: Changing the plugin selected required updating the `plugin_name` field as well as the options.

lemur/destinations/service.py
```diff
@@ -41,19 +41,26 @@ def create(label, plugin_name, options, description=None):
...
     :return:
     """
     destination = get(destination_id)
 
     destination.label = label
+    destination.plugin_name = plugin_name
+    # remove any sub-plugin objects before try to save the json options
+    for option in options:
+        if "plugin" in option["type"]:
+            del option["value"]["plugin_object"]
```
jtschladen, line 63: This delete logic is required or you get an error like "Object of type OpenSSLExportPlugin is not JSON serializable". This logic already existed in the `create` method, I just copied it here.

lemur/notifications/service.py
```diff
@@ -104,12 +104,13 @@ def create(label, plugin_name, options, description, certificates):
     return database.create(notification)
 
 
-def update(notification_id, label, options, description, active, certificates):
+def update(notification_id, label, plugin_name, options, description, active, certificates):
```
jtschladen, line 107: Function: Changing the plugin selected required updating the `plugin_name` field as well as the options.

lemur/sources/service.py
```diff
@@ -264,20 +264,22 @@ def create(label, plugin_name, options, description=None):
     return database.create(source)
 
 
-def update(source_id, label, options, description):
+def update(source_id, label, plugin_name, options, description):
```
jtschladen, line 267: Function: Changing the plugin selected required updating the `plugin_name` field as well as the options.

lemur/static/app/angular/destinations/destination/destination.js
```diff
@@ -52,19 +52,19 @@ angular.module('lemur')
           if (plugin.slug === $scope.destination.plugin.slug) {
             plugin.pluginOptions = $scope.destination.plugin.pluginOptions;
             $scope.destination.plugin = plugin;
-            _.each($scope.destination.plugin.pluginOptions, function (option) {
-              if (option.type === 'export-plugin') {
-                PluginService.getByType('export').then(function (plugins) {
-                  $scope.exportPlugins = plugins;
+            PluginService.getByType('export').then(function (plugins) {
+              $scope.exportPlugins = plugins;
```
jtschladen, line 56: Order of these calls swapped. If a destination was edited and it did NOT use an export plugin yet (e.g. if "AWS" was selected), then selecting a new destination plugin that _did_ require an export plugin resulted in the export plugin selection list being empty, since we never loaded the possible plugins. By swapping, we _always_ load the export plugins (even if we don't need them yet), so that we have them ready if the user tries to change the plugin type.

lemur/static/app/angular/notifications/notification/notification.js
```diff
@@ -42,25 +42,15 @@ angular.module('lemur')
       PluginService.getByType('notification').then(function (plugins) {
         $scope.plugins = plugins;
         _.each($scope.plugins, function (plugin) {
-          if (plugin.slug === $scope.notification.pluginName) {
-            plugin.pluginOptions = $scope.notification.notificationOptions;
+          if (plugin.slug === $scope.notification.plugin.slug) {
+            plugin.pluginOptions = $scope.notification.plugin.pluginOptions;
```
jtschladen, line 46: These fields were using incorrect variable names, resulting in the if statement never being executed and the plugin selection list showing up as empty.

lemur/static/app/angular/sources/source/source.js
```diff
@@ -41,22 +41,14 @@ angular.module('lemur')
       PluginService.getByType('source').then(function (plugins) {
         $scope.plugins = plugins;
         _.each($scope.plugins, function (plugin) {
-          if (plugin.slug === $scope.source.pluginName) {
+          if (plugin.slug === $scope.source.plugin.slug) {
+            plugin.pluginOptions = $scope.source.plugin.pluginOptions;
```
jtschladen, line 45: These fields were using incorrect variable names, resulting in the if statement never being executed and the plugin selection list showing up as empty.

charhate (approved):

hosseinsh (approved):

PR #3210, opened 2020-10-23
Digicert time bug fix

hosseinsh:
fixing a nasty old bug in the date for valid_to:  where `MM` stands for month, and needed to be `mm`
we introduced a new bug with `SS` fo sub-seconds, which needed to be `ss`.
https://arrow.readthedocs.io/en/stable/#supported-tokens

DigiCert currently does not honor the requested date, and sets the expiration to last second of previous day, from valid_to.


charhate (approved):

hosseinsh:
github seems stuck with integration tests.
Going ahead with the merge.

PR #3211, opened 2020-10-24
Modify description during reissue

charhate:
Include the certificate ID being reissued and mention that this is created by Lemur as part of reissue


hosseinsh:
looks good to me, I just wonder if we would also want to expand the tests for the cases of already reissue message present, and no message present.

https://github.com/Netflix/lemur/blob/cd29b2b870ae7bc89be5bf92491cea0641b48b1b/lemur/tests/test_certificates.py#L790-L809

hosseinsh (approved):

PR #3212, opened 2020-10-24
Issuer retry

hosseinsh:
This adds better retry strategy to Entrust, and also makes the delay shorter for DigiCert, as keeping the request alive for 10x 10s seems too long.


hosseinsh:
@sirferl we have been experiencing  this Error with Entrust, so hopefully retries would help

```
('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

charhate (approved):

PR #3214, opened 2020-10-27
Improving cn/owner search endpoint

hosseinsh:
This PR aims at improving the performance of this endpoint, but eliminating unnecessary elements.


lemur/certificates/service.py
```diff
@@ -560,18 +560,35 @@ def query_common_name(common_name, args):
...
+            .filter(Certificate.owner == owner)
+            .filter(not_(Certificate.revoked))
+            .filter(not_(Certificate.replaced.any()))  # ignore rotated certificates to avoid duplicates
+            .all()
+        )
+    else:
+        # search based on owner and cn
+        result = (
+            Certificate.query.filter(Certificate.not_after >= current_time.format("YYYY-MM-DD"))
+            .filter(Certificate.cn.like(common_name))
```
csine-nflx, line 586: The only difference between like and ilike is that ilike is case insensitive.  So in this case, if common_name=FOO.net and Certificate.cn = foo.net, there will be not be a match.  Not sure if that is an issue, but it might be.  Might be better to keep it as ilike to meet expectations.

hosseinsh: done

lemur/certificates/service.py
```diff
@@ -560,18 +560,35 @@ def query_common_name(common_name, args):
...
+            Certificate.query.filter(Certificate.not_after >= current_time.format("YYYY-MM-DD"))
+            .filter(not_(Certificate.revoked))
+            .filter(not_(Certificate.replaced.any()))  # ignore rotated certificates to avoid duplicates
+            .all()
+        )
+    elif common_name == "%":
+        # all valid certs from the owner
+        result = (
+            Certificate.query.filter(Certificate.not_after >= current_time.format("YYYY-MM-DD"))
+            .filter(Certificate.owner == owner)
```
csine-nflx, line 577: Is it possible that a request might come in where owner = '%foobar%'?  If so, then this request will likely fail.

Maybe something like the following?
if "%" in owner:
  Query using Certificate.owner.ilike(owner)
else:
  Query using Certificate.owner == owner

lemur/certificates/service.py
```diff
@@ -560,18 +560,35 @@ def query_common_name(common_name, args):
...
     # only not expired certificates
     current_time = arrow.utcnow()
 
-    result = (
-        Certificate.query.filter(Certificate.cn.ilike(common_name))
-        .filter(Certificate.owner.ilike(owner))
-        .filter(Certificate.not_after >= current_time.format("YYYY-MM-DD"))
-        .all()
-    )
+    if common_name == "%" and not owner:
```
csine-nflx, line 566: if "%" in common_name and not owner:

Done this way, you can change between using == and ilike.

csine-nflx (changes requested):
Made some comments.    I am worried that there may be a few failure conditions. Still, this is good work and will improve performance.

hosseinsh:
thanks for the feedback @csine-nflx 
moving to a simpler structurer.
`==` is not helping much compared to `ilike`, so removing that optimization

lemur/certificates/service.py
hosseinsh: moving to `ilike` since `==` was helping significantly.

charhate (approved):

PR #3219, opened 2020-10-28
Fix notification view to actually show associated certs

jtschladen:
This fixes the issue where editing a notification doesn't show the existing certificate associations for the notification. At present (without the fix), submitting an edited notification clears all certificate associations due to this bug.

I think the addition of the `showExpired` option missed this particular method. Without this fix, the following exception is thrown:

```
[2020-10-28 16:05:53,845] ERROR in schema: 'showExpired'
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/common/schema.py", line 160, in decorated_function
    resp = f(*args, **kwargs)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/certificates/views.py", line 1162, in get
    return service.render(args)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/certificates/service.py", line 425, in render
    show_expired = args.pop("showExpired")
KeyError: 'showExpired'
'showExpired'
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/common/schema.py", line 160, in decorated_function
    resp = f(*args, **kwargs)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/certificates/views.py", line 1162, in get
    return service.render(args)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/certificates/service.py", line 425, in render
    show_expired = args.pop("showExpired")
KeyError: 'showExpired'
```

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -1155,6 +1155,7 @@ class NotificationCertificatesList(AuthenticatedResource):
         )
         parser.add_argument("creator", type=str, location="args")
         parser.add_argument("show", type=str, location="args")
+        parser.add_argument("showExpired", type=int, location="args")
 
         args = parser.parse_args()
         args["notification_id"] = notification_id
--- a/lemur/static/app/angular/notifications/services.js
+++ b/lemur/static/app/angular/notifications/services.js
@@ -27,7 +27,7 @@ angular.module('lemur')
     };
 
     NotificationService.getCertificates = function (notification) {
-      notification.getList('certificates').then(function (certificates) {
+      notification.getList('certificates', {showExpired: 0}).then(function (certificates) {
         notification.certificates = certificates;
       });
     };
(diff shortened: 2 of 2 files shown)


lemur/certificates/views.py
```diff
@@ -1155,6 +1155,7 @@ def get(self, notification_id):
         )
         parser.add_argument("creator", type=str, location="args")
         parser.add_argument("show", type=str, location="args")
+        parser.add_argument("showExpired", type=int, location="args")
```
hosseinsh, line 1158: good catch @jtschladen.
would it make sense to default this to only not expired certs?
From user experience, probably only notifications on not expired certs would add value, as while editing the notification, you can drop them.

from Performance point of view, it would make a huge difference searching the entire DB or only valid certs.

hosseinsh (approved):

lemur/certificates/views.py
jtschladen: Added this, although I'd like to note for the record that hiding expired certs means those expired certs will not longer be associated with notifications after editing. I think this is fine, as we won't notify on expired certs anyway.

PR #3220, opened 2020-10-28
Add ability to override SourceArn for SES

jtschladen:
This is a requirement in order to use an [SES authorization policy](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-authorization-policies.html).

### DIFF
--- a/lemur/plugins/lemur_email/plugin.py
+++ b/lemur/plugins/lemur_email/plugin.py
@@ -38,7 +38,7 @@ def render_html(template_name, options, certificates):
 
 def send_via_smtp(subject, body, targets):
     """
-    Attempts to deliver email notification via SES service.
+    Attempts to deliver email notification via SMTP.
 
     :param subject:
     :param body:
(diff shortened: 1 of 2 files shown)


lemur/plugins/lemur_email/plugin.py
```diff
@@ -55,21 +55,25 @@ def send_via_smtp(subject, body, targets):
...
 def send_via_ses(subject, body, targets):
     """
-    Attempts to deliver email notification via SMTP.
+    Attempts to deliver email notification via SES service.
     :param subject:
     :param body:
     :param targets:
     :return:
     """
     client = boto3.client("ses", region_name="us-east-1")
```
hosseinsh, line 64: curious, if this should be a config: 
```
region_name="us-east-1"
```

hosseinsh (approved):

lemur/plugins/lemur_email/plugin.py
jtschladen: I think it absolutely should, though I noticed it's a widespread problem throughout all the AWS code. I think for now it might be best to leave it hardcoded and consider reworking the whole code base to take region as a configuration parameter.

jtschladen: Actually, I think I misread something. At a second glance, most regions come from plugin configs and aren't hardcoded. I'll fix it here now.

lemur/plugins/lemur_email/plugin.py
```diff
@@ -55,21 +55,28 @@ def send_via_smtp(subject, body, targets):
...
     :param body:
     :param targets:
     :return:
     """
-    client = boto3.client("ses", region_name="us-east-1")
-    client.send_email(
-        Source=current_app.config.get("LEMUR_EMAIL"),
-        Destination={"ToAddresses": targets},
-        Message={
+    ses_region = current_app.config.get("LEMUR_SES_REGION")
```
hosseinsh, line 64: minor suggestion, adding the default, when `LEMUR_SES_REGION` is not set

```
ses_region = current_app.config.get("LEMUR_SES_REGION", "us-east-1")
```

hosseinsh (approved):

PR #3221, opened 2020-10-29
Delegated CNAME DNS validation for ACME

csine-nflx:
The ACME protocol supports DNS delegation for validation records via CNAME, due to the nature of DNS,.  Rather than creating TXT records in the target domain, CNAME records can be used to delegate them to to other domains.  For example, lets say you want to request a certificate for foobar.com.  Normally with DNS-01, a TXT record would need to be created at _acme-challenge.foobar.com.  With CNAME delegation, you could instead create a CNAME at _acme-challenge.foobar.com that points to foobar.com.validationdomain.yourdomain.com, which is a CNAME containing the appropriate token.  This capability is enabled by setting __ACME_ENABLE_DELEGATED_CNAME = True_ in your config file.  When enabled, Lemur will follow CNAMEs and attempt to create TXT records in the target domains.

[The advantages of a CNAME](https://letsencrypt.org/2019/10/09/onboarding-your-customers-with-lets-encrypt-and-acme.html#the-advantages-of-a-cname)


lemur/plugins/lemur_acme/plugin.py
```diff
@@ -311,19 +316,32 @@ def get_authorizations(self, acme_client, order, order_info):
         authorizations = []
 
         for domain in order_info.domains:
-            if not self.dns_providers_for_domain.get(domain):
+
+            # If CNAME exists, set host to the target address
+            host = domain
+            if current_app.config.get("ACME_ENABLE_DELEGATED_CNAME", False):
+                val_domain, _ = self.strip_wildcard(domain)
+                val_domain = challenges.DNS01().validation_domain_name(val_domain)
```
hosseinsh, line 324: Maybe this would help with better readability: 
```val_domain --> acme_challenge_val_domain```

```
acme_challenge_val_domain = challenges.DNS01().validation_domain_name(val_domain)
cname_res = self.get_cname(acme_challenge_val_domain)

```

File: lemur/plugins/lemur_acme/plugin.py [hunk 2]
```diff
@@ -91,6 +92,7 @@ def start_dns_challenge(
         self,
         acme_client,
         account_number,
+        domain,
```
hosseinsh, line 95: I wonder if we could go with a better naming for `domain` and `host`, since they are overloaded here.
maybe `cnamed_host`, to indicate the purpose of host, which should help readability.

lemur/plugins/lemur_acme/plugin.py
```diff
@@ -431,6 +452,19 @@ def get_dns_provider(self, type):
             raise UnknownProvider("No such DNS provider: {}".format(type))
         return provider
 
+    def get_cname(self, domain):
+        """
+        :param domain: Domain name to look up a CNAME for.
+        :param record_type: Type of DNS record to lookup.
```
charhate, line 458: `record_type` can be removed if not needed.

lemur/plugins/lemur_acme/plugin.py
csine-nflx: These names reflect what ultimately ends up in the AuthorizationRecord. If renamed, it should probably be consistent.  The challenge is that the host is used regardless of whether there is a CNAME or not, so I don't think cname_host is ideal.  Perhaps something like target_domain.

lemur/plugins/lemur_acme/plugin.py
csine-nflx: This is essentially a temp variable, so I am not sure how much value there is in making it extra descriptive. I went ahead and just made it cname_result the whole way through this code block.

lemur/plugins/lemur_acme/plugin.py
csine-nflx: Thanks!

lemur/plugins/lemur_acme/plugin.py
hosseinsh: naming is hard. let's go with `target_name` for now

hosseinsh (approved):

lemur/plugins/lemur_acme/plugin.py
hosseinsh: thanks. it helps readability.

charhate (approved):

Issue #2635, opened 2019-03-05
Expiration notifications for security email stacking and sending multiple times


ecterun:
It seems that when the notify expiration call is being called we get some interesting results regarding the security email.  What seems to happen is if multiple notifications are going out in a single run the security email will get multiple emails, growing in number of incidents as it works through the list.

i.e. first notification security email gets one email with one certificate info in it, second it will get another email with both certificates information, third notification it will get another email with all three certificates info in the email, and so on.

I don't know if the logic behind lemur.notifications send_expiration_notifications() was to send a single email to the security email address with a single certificates info in each email, or to save some space and send a single email with a list off all certificates notifications at once.  Either way this seems to not be happening, that is unless i am not understanding the use case that was implemented.

I think we can fix it pretty easily depending on what the intended outcome was, either add a security_data initialization were the notification_data is being done.  Or we can move the send_notification for the security_data outside of the notification_group.items() for loop so it runs after all notifications have been processed and all notifications have been added to security_data.

PR #3235, opened 2020-11-04
Stop repeating certs when sending expiration notifications to security team email

jtschladen:
This solves #2635. There remains an open question (raised in that issue) as to whether we should be sending a single email to the security team with all expiring certs, or if it should one email per cert-owner-notification group, as it is now. For the moment, I'm taking the easier approach and leaving it a single email per cert, but may revisit later.


hosseinsh (approved):
I believe long-term, have one email per day to security with all the expiring certs under different categories, e.g., attached to load balancer or not, would be more valuable.

PR #3236, opened 2020-11-05
Version updates

charhate:
To fix build errors like :

```
ERROR: flake8 3.5.0 has requirement pyflakes<1.7.0,>=1.5.0, but you'll have pyflakes 2.2.0 which is incompatible.
ERROR: black 20.8b1 has requirement click>=7.1.2, but you'll have click 7.1.1 which is incompatible.
```

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -12,6 +12,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from flask import current_app
 from sqlalchemy import func, or_, not_, cast, Integer
+from sqlalchemy.sql.expression import false, true
 
 from lemur import database
 from lemur.authorities.models import Authority
@@ -150,7 +151,7 @@ def get_all_certs_attached_to_endpoint_without_autorotate():
         """
     return (
         Certificate.query.filter(Certificate.endpoints.any())
-        .filter(Certificate.rotation == False)
+        .filter(Certificate.rotation == false())
         .filter(Certificate.not_after >= arrow.now())
         .filter(not_(Certificate.replaced.any()))
         .all()  # noqa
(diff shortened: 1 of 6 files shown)


lemur/certificates/service.py
```diff
@@ -150,7 +151,7 @@ def get_all_certs_attached_to_endpoint_without_autorotate():
         """
     return (
         Certificate.query.filter(Certificate.endpoints.any())
-        .filter(Certificate.rotation == False)
+        .filter(Certificate.rotation == false())
```
charhate, line 154: lint does not like `X == boolean` comparision, which is needed in filter. Thus using `sqlalchemy.sql.expression`

lemur/certificates/utils.py
```diff
@@ -82,4 +82,4 @@ def get_key_type_from_csr(data):
             raise Exception("Unsupported key type")
 
     except NotImplemented:
-        raise NotImplemented()
+        raise NotImplementedError
```
charhate, line 85: for lint error

requirements-dev.txt
```diff
@@ -15,7 +15,7 @@ cryptography==3.2.1       # via secretstorage
 distlib==0.3.0            # via virtualenv
 docutils==0.16            # via readme-renderer
 filelock==3.0.12          # via virtualenv
-flake8==3.5.0             # via -r requirements-dev.in
+flake8==3.8.0             # via -r requirements-dev.in
```
hosseinsh, line 18: this file is auto-generated based on the input in
https://github.com/Netflix/lemur/blob/7ea6e2059a6cacf4284cf901de26ab335d6775f2/requirements-dev.in#L3

once the requirement is modified, one can either run 
```
make up-reqs
```
to compute the requirements-dev.txt

or delegate to dependbot to file a PR

hosseinsh:
nice to see these errors go away

requirements-dev.txt
charhate: Thanks for the details, flake8 update needed more version update and lint fixes. So I will do the manual steps and update the PR.

hosseinsh (approved):

PR #3246, opened 2020-11-11
Send a single email to multiple recipients instead of multiple emails

jtschladen:
Modify email notification logic to send a single email that combines owner, security, and notification recipients, instead of sending three separate emails.

The intent of this change is that sending a single email that includes the security team should make it more clear exactly which emails were sent, as (for now) the security team will receive a copy of each email.

Note: More extensive changes to security emails will be coming, but this a short-term fix to increase email visibility.

### DIFF
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -161,8 +161,10 @@ def send_expiration_notifications(exclude):
                 success += len(email_recipients)
             else:
                 failure += len(email_recipients)
-            # If we're using an email plugin, we're done;
-            # if not, we also need to send an email notification to the security team and owner
+            # If we're using an email plugin, we're done,
+            #   since "security_email + [owner]" were added as email_recipients.
+            # If we're not using an email plugin, we also need to send an email to the security team and owner,
+            #   since the plugin notification didn't send anything to them.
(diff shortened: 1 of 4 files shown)


charhate (approved):

lemur/notifications/messaging.py
```diff
@@ -152,33 +152,30 @@ def send_expiration_notifications(exclude):
...
-            else:
-                failure += 1
+                if send_default_notification(
+                        "expiration", notification_data, email_recipients, notification.options
+                ):
+                    success += len(email_recipients)
+                else:
+                    failure += len(email_recipients)
+
+                if send_plugin_notification(
```
hosseinsh, line 173: would it make any difference if for all plugins this is called?
```
if send_plugin_notification(
                        "expiration", notification_data, email_recipients, notification
                ):
``` 


sns and slack plugins would be ignoring `email_recipients` anyways, right?

jtschladen: That's true, it makes no difference. The API is a bit odd since only the email plugin needs recipients, and the others accept but ignore them. We might be able to condense down to two send calls (one plugin, one default) if I do it that way, let me double check and make sure it works.

hosseinsh (approved):

PR #3248, opened 2020-11-12
Entrust: Improve organization handling 

sirferl:
Entrust does not regard the organization in the CSR but needs the information send in the Additional Data sent with the request. The plugin now get's the organizations from Entrust and looks up the ID of the wanted organization from the "Organization"-field. If the Organization is not valid, possible values are displayed in an error message.


lemur/plugins/lemur_entrust/plugin.py
```diff
@@ -78,11 +78,35 @@ def process_options(options):
         "eku": "SERVER_AND_CLIENT_AUTH",
         "certType": product_type,
         "certExpiryDate": validity_end,
+        # "keyType": "RSA", Entrust complaining about this parameter
```
hosseinsh, line 81: the `client_id` is missing from the data.

hosseinsh (commented):
I am wondering if we should move towards issuer options, and add all the CA related info as dynamic options, similar how we are doing it with the ACME plugin: 

https://github.com/Netflix/lemur/blob/95854c6f686cce7dc8b6919388fe57abfdd03fa4/lemur/plugins/lemur_acme/plugin.py#L41-L76


This would mean one issuer per Org-ID. 
depending on the number of orgs at Entrust, it might be worthwhile to manage one or two org-based issuers, and save the extra query for each cert issuance. 

On the other hand, if the number of orgs is too high, this approach might be worth it.

lemur/plugins/lemur_entrust/plugin.py
sirferl: *sigh* Had problems syncing my dev-enviroment. So somehow the two lines got lost. Thank you. I repaired it.

sirferl:
Hi, I appreciate your comment. Actually I have thought in the same direction. But we have a varying number of organizations due to M&A. So it seems better to use the check always approach. Maybe we should implement a config-switch to turn off this check when there is only one organization?

hosseinsh:
looks like there is an Azure plugin coming in :D 
looking forward to that PR (it suppose it is added here by mistake)

hosseinsh:
> Hi, I appreciate your comment. Actually I have thought in the same direction. But we have a varying number of organizations due to M&A. So it seems better to use the check always approach. Maybe we should implement a config-switch to turn off this check when there is only one organization?

Sounds good, we can add that later, if we notice the overhead is too high.

hosseinsh:
this might come handy: 
https://github.com/Netflix/lemur/blob/37e58574066921f50485398d195994b1b805c216/lemur/plugins/lemur_jks/plugin.py#L42-L46

hosseinsh:
@sirferl would be moving the Azure plugin into a separate PR?

sirferl:
Yeah, I got a bit carried away friday afternoon working ont the azure plugin. I will clean up the mess...

hosseinsh (approved):
local testing looking good. 

Thanks for the PR @sirferl. Merging shortly.

PR #3267, opened 2020-11-23
Remove certificate from AWS and cleanup after cert revoke

charhate:
Removing destination in Lemur only updates the destination record in Lemur without removing the certificate from destination. The `delete_cert` API from AWS IAM inherently checks if certificate is attached to an endpoint. Use of this API helps in cleanup of unused certificates and frees up the spots. In case of the non-IAM endpoints (supported in Lemur) we do not have sufficient information to perform such change.

Additionally, this PR contains change for Certificate update and cleanup after revocation:
1. Disabling notification
2. Disabling auto-rotation
3. Update certificate status to 'revoked'
4. Remove from AWS

### DIFF
--- a/lemur/plugins/lemur_aws/elb.py
+++ b/lemur/plugins/lemur_aws/elb.py
@@ -160,8 +160,8 @@ def get_load_balancer_arn_from_endpoint(endpoint_name, **kwargs):
     try:
         client = kwargs.pop("client")
         elbs = client.describe_load_balancers(Names=[endpoint_name])
-        for elb in elbs["LoadBalancers"]:
-            return elb["LoadBalancerArn"]
+        if "LoadBalancers" in elbs and elbs["LoadBalancers"]:
+            return elbs["LoadBalancers"][0]["LoadBalancerArn"]
 
     except Exception as e:  # noqa
         metrics.send(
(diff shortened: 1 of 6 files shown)


lemur/plugins/lemur_aws/elb.py
```diff
@@ -149,6 +149,38 @@ def get_listener_arn_from_endpoint(endpoint_name, endpoint_port, **kwargs):
...
+    """
+    Get a load balancer ARN from an endpoint.
+    :param endpoint_name:
+    :return:
+    """
+    try:
+        client = kwargs.pop("client")
+        elbs = client.describe_load_balancers(Names=[endpoint_name])
+        for elb in elbs["LoadBalancers"]:
+            return elb["LoadBalancerArn"]
```
hosseinsh, line 164: why do we iterate over elbs to just return the first `elb["LoadBalancerArn"]`?

hosseinsh (approved):
Amazing work! thanks for the tests and increasing our coverage!

lemur/plugins/lemur_aws/elb.py
charhate: Changed code

PR #3269, opened 2020-11-25
LDAP: error decoding CN when using Active Directory

sirferl:
I set up LDAP Authentication.
Certain users worked, but some not.
I got the following error:
ERROR in views: ldap error: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)

Digging around I found the problem in auth/ldap.py:214.
There a decode with "ascii" is performed. When i changed this to "utf-8" it worked for the failing users as well.

self.ldap_groups.append(values["cn"][0].decode("utf-8"))

There is a similar code a few lines below, but I can only test AD.

### DIFF
--- a/lemur/auth/ldap.py
+++ b/lemur/auth/ldap.py
@@ -211,7 +211,7 @@ class LdapPrincipal:
             for group in lgroups:
                 (dn, values) = group
                 if type(values) == dict:
-                    self.ldap_groups.append(values["cn"][0].decode("ascii"))
+                    self.ldap_groups.append(values["cn"][0].decode("utf-8"))
         else:
             lgroups = self.ldap_client.search_s(
                 self.ldap_base_dn, ldap.SCOPE_SUBTREE, ldap_filter, self.ldap_attrs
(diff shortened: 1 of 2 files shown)


hosseinsh:
Hey @sirferl, 
looks like there is even a comment on the next decoding, saying it is `utf-8` encoded!

https://github.com/Netflix/lemur/blob/6d750076442e579f78cc63916208be9843a798bb/lemur/auth/ldap.py#L219-L221


We don't use ldap, and I hope no means to test that. So leaving it for now as is.

hosseinsh:
Thanks for the PR. merging it now.

sirferl:
thank you. 
reading this, it looks like UTF-8 would be the better alternative:
https://www.python-ldap.org/en/python-ldap-3.3.0/bytes_mode.html#what-s-text-and-what-s-bytes ->
_"The LDAP protocol states that some fields (distinguished names, relative distinguished names, attribute names, queries) be encoded in UTF-8. In python-ldap, these are represented as text (str on Python 3, unicode on Python 2)."_

PR #3270, opened 2020-11-26
Change certificate status to revoked after ACME revoke_certificate

AToMiXhawK:
On successful completion of ACME revoke_certificate, change
certificate.status to revoked

Changes validity as Revoked
image


hosseinsh:
Hi @AToMiXhawK, thanks for the PR. 
Lemur has a cron job that updates the revocation status, but the better behavior is to do this right after revocation, instead of waiting for the cron job to do it. 

The DB update, however, should happen at a higher level than the plugin, otherwise we would need to update each plugin. 

There is PR that is doing exactly this and more clean-up
https://github.com/Netflix/lemur/pull/3267

hosseinsh:
#3267 is now merged, and would address this PR. Please let us know if that change is not sufficient for your case.

PR #3277, opened 2020-12-02
CRL Reason during certificate revoke

charhate:
As per [RFC 5280 Section 5.3.1](https://tools.ietf.org/html/rfc5280#section-5.3.1), Lemur UI will offer CRL reasons to select during certificate revocation. This information will further be sent to Authority as long as revoke API offers such option.
```
Unspecified
Key Compromise
CA Compromise
Affiliation Changed
Superseded
Cessation of Operation
Certificate Hold
Remove from CRL
Privilege Withdrawn
Attribute Authority Compromise
```

The revocation via CLI is as well modified to have additional parameter
e.g. 
`certificate revoke -c -p "/x/cert-ids" -r "affilicationChanged" -m "Test CLI revoke"`

The CertificateRevokeSchema now has two members, **comments** which is a string message explaining the reason of revocation and crl_reason as per [RFC 5280](https://tools.ietf.org/html/rfc5280#section-5.3.1)

Additionally, fixed the validation on Revoke form which asks to confirm the name of the certificate to be revoked

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -619,9 +619,8 @@ def clear_pending():
     v.clear_pending_certificates()
 
 
-@manager.option(
-    "-p", "--path", dest="path", help="Absolute file path to a Lemur query csv."
-)
+@manager.option("-p", "--path", dest="path", help="Absolute file path to a Lemur query csv.")
+@manager.option("-id", "--certid", dest="cert_id", help="ID of the certificate to be revoked")
 @manager.option("-r", "--reason", dest="reason", default="unspecified", help="CRL Reason as per RFC 5280 section 5.3.1")
 @manager.option("-m", "--message", dest="message", help="Message explaining reason for revocation")
 @manager.option(
@@ -632,10 +631,17 @@ def clear_pending():
     default=False,
     help="Persist changes.",
 )
-def revoke(path, reason, message, commit):
+def revoke(path, cert_id, reason, message, commit):
     """
     Revokes given certificate.
     """
+    if not path and not cert_id:
+        print("[!] No input certificates mentioned to revoke")
+        return
+    if path and cert_id:
+        print("[!] Please mention single certificate id (-id) or input file (-p)")
+        return
+
     if commit:
         print("[!] Running in COMMIT mode.")
 
@@ -645,9 +651,12 @@ def revoke(path, reason, message, commit):
         reason = CRLReason.unspecified.name
     comments = {"comments": message, "crl_reason": reason}
 
-    with open(path, "r") as f:
-        for x in f.readlines()[2:]:
-            worker(x, commit, comments)
+    if cert_id:
+        worker(cert_id, commit, comments)
+    else:
+        with open(path, "r") as f:
+            for x in f.readlines()[2:]:
+                worker(x, commit, comments)
 
 
 @manager.command
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -5,7 +5,6 @@
     :license: Apache, see LICENSE for more details.
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
-import multiprocessing
 import sys
 from flask import current_app
 from flask_principal import Identity, identity_changed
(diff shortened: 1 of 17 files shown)


lemur/certificates/cli.py
```diff
@@ -586,11 +586,10 @@ def worker(data, commit, reason):
     parts = [x for x in data.split(" ") if x]
     try:
         cert = get(int(parts[0].strip()))
-        plugin = plugins.get(cert.authority.plugin_name)
 
         print("[+] Revoking certificate. Id: {0} Name: {1}".format(cert.id, cert.name))
         if commit:
-            plugin.revoke_certificate(cert, reason)
+            revoke_certificate(cert, reason)
```
charhate, line 592: Calling `revoke_certificate` from service since it will perform needed DB updates in Lemur

lemur/certificates/cli.py
```diff
@@ -623,7 +622,8 @@ def clear_pending():
 @manager.option(
     "-p", "--path", dest="path", help="Absolute file path to a Lemur query csv."
 )
-@manager.option("-r", "--reason", dest="reason", help="Reason to revoke certificate.")
+@manager.option("-r", "--reason", dest="reason", default="unspecified", help="CRL Reason as per RFC 5280 section 5.3.1")
```
charhate, line 625: Reused `--reason` for CRL Reason as it matches the RFC guidance

lemur/certificates/cli.py
```diff
@@ -641,11 +641,13 @@ def revoke(path, reason, commit):
...
     print("[+] Starting certificate revocation.")
 
-    with open(path, "r") as f:
-        args = [[x, commit, reason] for x in f.readlines()[2:]]
+    if reason not in CRLReason.__members__:
+        reason = CRLReason.unspecified.name
+    comments = {"comments": message, "crl_reason": reason}
 
-    with multiprocessing.Pool(processes=3) as pool:
-        pool.starmap(worker, args)
```
charhate, line 648: This method was not working when used with multiprocessing (tried using ThreadPoolExecutor too), the error it gives is.
`No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.`
Even with app_context, it complained when tried to modify certificate object later in code. So moving it to sequential processing. It can be revisited if actively used and performance improvement is needed.

lemur/certificates/views.py
```diff
@@ -1459,13 +1459,9 @@ def put(self, certificate_id, data=None):
                         403,
                     )
 
-        plugin = plugins.get(cert.authority.plugin_name)
-        plugin.revoke_certificate(cert, data)
```
charhate, line 1463: moved revoke + cleanup to certificates/service.py

lemur/certificates/views.py
```diff
@@ -1398,7 +1398,7 @@ def __init__(self):
         self.reqparse = reqparse.RequestParser()
         super(CertificateRevoke, self).__init__()
 
-    @validate_schema(None, None)
+    @validate_schema(certificate_revoke_schema, None)
     def put(self, certificate_id, data=None):
         """
         .. http:put:: /certificates/1/revoke
```
hosseinsh, line 1404: should we update the expected input, in terms of `crl_reason`?

hosseinsh (approved):
thanks for the amazing PR!

lemur/certificates/cli.py
```diff
@@ -623,7 +622,8 @@ def clear_pending():
 @manager.option(
     "-p", "--path", dest="path", help="Absolute file path to a Lemur query csv."
 )
-@manager.option("-r", "--reason", dest="reason", help="Reason to revoke certificate.")
+@manager.option("-r", "--reason", dest="reason", default="unspecified", help="CRL Reason as per RFC 5280 section 5.3.1")
+@manager.option("-m", "--message", dest="message", help="Message explaining reason for revocation")
```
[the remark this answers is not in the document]
charhate: Digicert and Entrust accept a string message during revoke. Digicert rather does not accept crl_reason. Thus, digicert plugin is modified to add crl_reason as a comment. IMO it's flexible to keep an option to provide a separate message in addition to reason.

I will add an option to revoke single certificate

hosseinsh (approved):
great improvements!

lemur/certificates/views.py
charhate: Added [here](https://github.com/Netflix/lemur/pull/3277/files#diff-88f1b5db302abda0bddba15f5af7d9c1aed03efcdd87d603bd0c3a96c41d772fR1417)

lemur/certificates/service.py
```diff
@@ -828,6 +828,14 @@ def remove_from_destination(certificate, destination):
         plugin.clean(certificate=certificate, options=destination.options)
 
 
+def revoke(certificate, reason):
+    plugin = plugins.get(certificate.authority.plugin_name)
+    plugin.revoke_certificate(certificate, reason)
```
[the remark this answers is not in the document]
charhate: Added it [here](https://github.com/Netflix/lemur/pull/3277/files#diff-88f1b5db302abda0bddba15f5af7d9c1aed03efcdd87d603bd0c3a96c41d772fR1476) along with some more error handling

PR #3278, opened 2020-12-02
Preselect ECCPRIME256V1 on UI for cert minting

charhate:
This change is to pre-select **ECCPRIME256V1** algorithm on UI while creating a new certificate. Before this change, RSA2048 was default value of the algorithm. User can go to 'More Options' to select algorithm of choice (no change there).


hosseinsh (approved):
Shall we include this default in the Lemur documentation?

charhate:
Sure, I will update lemur docs.

PR #3279, opened 2020-12-02
Use --no-emit-index-url instead of deprecated --no-index in pip-compile

atugushev:
`pip-compile --no-index` is deprecated since [pip-tools==5.2.0](https://github.com/jazzband/pip-tools/releases/tag/5.2.0)


hosseinsh (approved):
thanks for the PR @atugushev.
I don't have permissions to update the branch. Once the branch is updated, I can go ahead and merge the PR.

PR #3281, opened 2020-12-02
Add email notifications for CA cert expiry

jtschladen:
Adds support for email notifications for CA expiry. Must be enabled via cron or celery.

Sample email:
Screen Shot 2020-12-02 at 8 40 20 AM

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -348,6 +348,15 @@ Lemur supports sending certificate expiration notifications through SES and SMTP
 
           LEMUR_SECURITY_TEAM_EMAIL_INTERVALS = [15, 2]
 
+.. data:: LEMUR_AUTHORITY_CERT_EXPIRATION_EMAIL_INTERVALS
+    :noindex:
+
+       Notification interval set for CA certificate expiration notifications. If unspecified, the value [365, 180] is used (roughly one year and 6 months).
+
+       ::
+
+          LEMUR_AUTHORITY_CERT_EXPIRATION_EMAIL_INTERVALS = [365, 180]
+
 
 Celery Options
 ---------------
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -821,6 +821,42 @@ def notify_expirations():
 
 
 @celery.task(soft_time_limit=3600)
+def notify_authority_expirations():
+    """
+    This celery task notifies about expiring certificate authority certs
+    :return:
+    """
+    function = f"{__name__}.{sys._getframe().f_code.co_name}"
+    task_id = None
+    if celery.current_task:
+        task_id = celery.current_task.request.id
+
+    log_data = {
+        "function": function,
+        "message": "notify for certificate authority cert expiration",
+        "task_id": task_id,
+    }
+
+    if task_id and is_task_active(function, task_id, None):
+        log_data["message"] = "Skipping task: Task is already active"
+        current_app.logger.debug(log_data)
+        return
+
+    current_app.logger.debug(log_data)
+    try:
+        cli_notification.authority_expirations()
+    except SoftTimeLimitExceeded:
+        log_data["message"] = "Notify expiring CA Time limit exceeded."
+        current_app.logger.error(log_data)
+        sentry.captureException()
+        metrics.send("celery.timeout", "counter", 1, metric_tags={"function": function})
+        return
+
+    metrics.send(f"{function}.success", "counter", 1)
+    return log_data
+
+
+@celery.task(soft_time_limit=3600)
 def enable_autorotate_for_certs_attached_to_endpoint():
     """
     This celery task automatically enables autorotation for unexpired certificates that are
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -62,6 +62,34 @@ def get_certificates(exclude=None):
     return certs
 
 
+def get_expiring_authority_certificates():
+    """
+    Finds all certificate authority certificates that are eligible for expiration notifications.
+    :return:
+    """
+    now = arrow.utcnow()
+    authority_expiration_intervals = current_app.config.get("LEMUR_AUTHORITY_CERT_EXPIRATION_EMAIL_INTERVALS",
+                                                            [365, 180])
+    max_not_after = now + timedelta(days=max(authority_expiration_intervals) + 1)
+
+    q = (
+        database.db.session.query(Certificate)
+        .filter(Certificate.not_after < max_not_after)
+        .filter(Certificate.notify == true())
+        .filter(Certificate.expired == false())
+        .filter(Certificate.revoked == false())
+        .filter(Certificate.root_authority_id.isnot(None))
+        .filter(Certificate.authority_id.is_(None))
+    )
+
+    certs = []
+    for c in windowed_query(q, Certificate.id, 10000):
+        days_remaining = (c.not_after - now).days
+        if days_remaining in authority_expiration_intervals:
+            certs.append(c)
+    return certs
+
+
 def get_eligible_certificates(exclude=None):
     """
     Finds all certificates that are eligible for certificate expiration notification.
@@ -86,6 +114,25 @@ def get_eligible_certificates(exclude=None):
         # group by notification
         for notification, items in groupby(notification_groups, lambda x: x[0].label):
             certificates[owner][notification] = list(items)
+
+    return certificates
+
+
+def get_eligible_authority_certificates():
+    """
+    Finds all certificate authority certificates that are eligible for certificate expiration notification.
+    Returns the set of all eligible CA certificates, grouped by owner and interval, with a list of applicable certs.
+    :return:
+    """
+    certificates = defaultdict(dict)
+    all_certs = get_expiring_authority_certificates()
+    now = arrow.utcnow()
+
+    # group by owner
+    for owner, owner_certs in groupby(all_certs, lambda x: x.owner):
+        # group by expiration interval
+        for interval, interval_certs in groupby(owner_certs, lambda x: (x.not_after - now).days):
+            certificates[owner][interval] = list(interval_certs)  # list(c for c in interval_certs)
 
     return certificates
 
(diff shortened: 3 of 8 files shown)


lemur/plugins/lemur_email/templates/authority_expiration.html
```diff
@@ -0,0 +1,172 @@
...
+                                        <table style="min-width:300px" border="0" cellspacing="0" cellpadding="0">
+                                            <tbody>
+                                            <tr>
+                                                <td style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:13px;color:#202020;line-height:1.5">
+                                                    Hi,
+                                                </td>
+                                            </tr>
+                                            <tr>
+                                                <td style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:13px;color:#202020;line-height:1.5">
+                                                    <br>This is a Lemur CA certificate expiration notice. The follow CA certificates are expiring soon;
```
charhate, line 78: `The follow CA certificates` --> `Following CA certificates`

charhate (approved):

lemur/tests/test_messaging.py
```diff
@@ -125,3 +126,47 @@ def test_send_pending_failure_notification(notification_plugin, async_issuer_plu
...
+    create_cert_that_expires_in_days(364)
+    create_cert_that_expires_in_days(366)
+    create_cert_that_expires_in_days(179)
+    create_cert_that_expires_in_days(181)
+    create_cert_that_expires_in_days(1)
+
+    assert send_authority_expiration_notifications() == (2, 0)
+
+
+def create_cert_that_expires_in_days(days):
```
hosseinsh, line 162: should this maybe named `create_ca_cert_..` or `create_issuer_cert_...`

lemur/plugins/lemur_email/templates/authority_expiration.html
```diff
@@ -0,0 +1,172 @@
...
+                                            <tbody>
+                                            <tr>
+                                                <td style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:13px;color:#202020;line-height:1.5">
+                                                    Hi,
+                                                </td>
+                                            </tr>
+                                            <tr>
+                                                <td style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:13px;color:#202020;line-height:1.5">
+                                                    <br>This is a Lemur CA certificate expiration notice. The following CA certificates are expiring soon;
+                                                    please take manual action to renew them if necessary. You may also disable notifications via the
```
hosseinsh, line 79: shall we add something along the lines, to ensure we point out this not a simple rotate the cert operation.

```
Note that rotating a root CA requires advanced planing and the respective trustStores need to be updated.
A subCA on the other hand, does not require any changes to the trustStore.
```

lemur/notifications/messaging.py
```diff
@@ -90,6 +118,25 @@ def get_eligible_certificates(exclude=None):
...
+    """
+    Finds all certificate authority certificates that are eligible for certificate expiration notification.
+    Returns the set of all eligible CA certificates, grouped by owner and interval, with a list of applicable certs.
+    :return:
+    """
+    certificates = defaultdict(dict)
+    all_certs = get_expiring_authority_certificates()
+    now = arrow.utcnow()
+
+    # group by owner
```
hosseinsh, line 131: fancy `groupby` use, nice!

hosseinsh (approved):
Thanks for the awesome PR!

it would be a huge value to update the lemur docs with this new crucial notification type, and how to enable it. Ideally with the latest screenshot of the notification, which would be super handy.

lemur/plugins/lemur_email/templates/authority_expiration.html
jtschladen: Good idea. Ideally we could also indicate whether the CA is a root of sub-CA in the email, but I don't see any handy DB columns that could give that info, and it seems a little excessive to parse the cert to check if it's self-signed.

lemur/notifications/messaging.py
```diff
@@ -62,6 +62,34 @@ def get_certificates(exclude=None):
...
 
+def get_expiring_authority_certificates():
+    """
+    Finds all certificate authority certificates that are eligible for expiration notifications.
+    :return:
+    """
+    now = arrow.utcnow()
+    authority_expiration_intervals = current_app.config.get("LEMUR_AUTHORITY_CERT_EXPIRATION_EMAIL_INTERVALS",
+                                                            [365, 180])
+    max_not_after = now + timedelta(days=max(authority_expiration_intervals) + 1)
```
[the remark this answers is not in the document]
jtschladen: That would be nice, but since the intervals are cert-specific I don't think it's trivial to change. I'll give it some thought, but won't change in this PR.

lemur/plugins/lemur_email/templates/authority_expiration.html
hosseinsh: ideally the number of CA certs should be low enough such that checking for self-signed would not consume too much resources.
if you would consider it as a potential improvement in the messaging, this might be what you need

https://github.com/Netflix/lemur/blob/028f0c333042346a6c155ea7692d0a5b74a27ee1/lemur/common/utils.py#L265-L275

hosseinsh: Also another improvement that comes to mind (maybe as part of a future PR), is to list the number of valid leaf certs under that CA, similar to the number of endpoints for the expiry notifications.
this might be a good indication, in terms of how actively this CA is being used.

jtschladen:
> it would be a huge value to update the lemur docs with this new crucial notification type, and how to enable it. Ideally with the latest screenshot of the notification, which would be super handy.

I ended up rewriting most of the notification section of the docs :) I did not include a screenshot yet as I'm still thinking about how it would be appropriate - I think we could do with a bit more rewriting of notifications (including how to configure them), with screenshots, but I'm going to leave that for a later PR.

docs/administration.rst
```diff
@@ -262,22 +262,101 @@ and are used when Lemur creates the CSR for your certificates.
...
+
+**Email-only notifications**
+
+These notifications can only be sent via email and cannot use other notification plugins.
+
+Supported types:
+
+* CA certificate expiration
+* Pending ACME certificate failure
+* Certificate rotation (currently disabled in code)
```
hosseinsh, line 289: shall we change this to "currently not enabled in the celery task for rotation"?

one could create a new celery task which indeed sends out notifications: 
https://github.com/Netflix/lemur/blob/e87cf040f3fa531f407a34876ae71b4970d6369f/lemur/common/celery.py#L661-L663


or 

make sure the cron task includes the notify option
```certificate rotate -c -a```

https://github.com/Netflix/lemur/blob/56169656377637b7a0e53514105fbe31b436c4a9/lemur/certificates/cli.py#L186-L192

**Review by hosseinsh — APPROVED** [review]
Thanks for the amazing documentation!
Sounds good, with can certainly break it down to smaller PRs.

File: docs/administration.rst (hunk 6 above)
Comment (jtschladen) [reply docs/administration.rst:289]: I'll change the docs, but I'm also going to add a config option that allows celery to send rotation notifications without a code change.

**Comment by jtschladen** [comment]
Updated screenshot of the notification email. I haven't yet managed to figure out how to create a root CA for testing (specifically with a 180/365 day expiration), or why there are by default 40 certs issued by my test authority - I think there's some test magic that's messing with my data. Therefore I haven't thoroughly tested the self-signed or issued cert count logic thoroughly. 

Screen Shot 2020-12-03 at 3 59 16 PM

File: lemur/certificates/service.py [hunk 7]
```diff
@@ -864,3 +864,13 @@ def cleanup_after_revoke(certificate):
...
     return error_message
+
+
+def get_issued_cert_count_for_authority(authority):
+    """
+    Returns the count of certs issued by the specified authority.
+
+    :return:
+    """
+    query = database.session_query(Certificate.id).filter(Authority.id == authority.id)
```
hosseinsh, line 875: I think I have found the reason behind 40 issued certs: 
```
Certificate.authority_id == authority.id
```

This does something close to what we need here, but not at the level of CA, but the issuer type (like all acme-issuer based CAs):

https://github.com/Netflix/lemur/blob/56169656377637b7a0e53514105fbe31b436c4a9/lemur/certificates/service.py#L118-L123

in the Lemur UI, the Authority page renders all certs issued off a CA, and I was looking for that functionality, when I came across the code behind it: 

https://github.com/Netflix/lemur/blob/37e58574066921f50485398d195994b1b805c216/lemur/authorities/views.py#L492-L517

the Authority keeps track off all the children ever issued:
https://github.com/Netflix/lemur/blob/37e58574066921f50485398d195994b1b805c216/lemur/authorities/models.py#L54-L56

lemur/common/celery.py
```diff
@@ -656,11 +656,12 @@ def certificate_rotate(**kwargs):
 
     current_app.logger.debug(log_data)
     try:
+        notify = current_app.config.get("ENABLE_ROTATION_NOTIFICATION", None)
```
hosseinsh, line 659: nice feature!

lemur/plugins/lemur_email/templates/authority_expiration.html
```diff
@@ -91,7 +91,12 @@
                                                                     <span style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:20px;color:#202020">{{ certificate.name }}</span>
                                                                     <br>
                                                                     <span style="font-family:Roboto-Regular,Helvetica,Arial,sans-serif;font-size:13px;color:#727272">
-                                                                        {{ certificate.endpoints | length }} Endpoints
+                                                                       {% if certificate.self_signed %}
+                                                                           <b>Root</b>
+                                                                       {% else %}
+                                                                           Subordinate
```
hosseinsh, line 97: shall we use Intermediate instead of Subordinate. I might be only me, but Subordinate sounds a bit odd to my ears :D

hosseinsh (commented):
nice additions @jtschladen!

lemur/certificates/service.py
jtschladen: Thanks, good catch. I'm finding it a bit difficult to locate good examples for queries for SQLAlchemy 😅  but I think it should be correct now.

PR #3282, opened 2020-12-03
Fix the docs, attempt #2

jtschladen:
[The docs](https://readthedocs.org/projects/lemur/builds/) once again broke from commit 12b2069 by dependabot. Since the workaround that was originally in place doesn't seem to work for dependabot, I wanted to take a different approach.

I am not deeply familiar with Python dependency management, so please let me know if this doesn't make sense.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -79,7 +79,6 @@ pyrfc3339==1.1            # via -r requirements.txt, acme
 python-dateutil==2.8.1    # via -r requirements.txt, alembic, arrow, botocore
 python-editor==1.0.4      # via -r requirements.txt, alembic
 python-json-logger==0.1.11  # via -r requirements.txt, logmatic-python
-python-ldap==3.3.1        # via -r requirements.txt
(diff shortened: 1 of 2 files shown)


requirements-docs.txt
```diff
@@ -79,7 +79,6 @@ pyrfc3339==1.1            # via -r requirements.txt, acme
 python-dateutil==2.8.1    # via -r requirements.txt, alembic, arrow, botocore
```
jtschladen, line 79: Note that I chose not to completely regenerate the file at this time - should I? The next time dependabot updates in, I expect significant changes as most of the dependencies will have been dropped. 

I did try building locally as follows:
```
pip-compile --no-index --output-file=requirements-docs.txt requirements-docs.in
make dev-docs
cd docs
make html
```

And it seems to build just fine without any of the removed dependencies.

hosseinsh (approved):
testing locally looks good on my end.
there is only one way of figuring out!

PR #3283, opened 2020-12-03
Lemur doc update

charhate:
- The UI fields have changed over time and thus needed an update. I've updated images in user guide and modified steps wherever applicable. 
- Added a reference to configuring automated tasks using celery in the Quickstart guide.


hosseinsh (approved):
one suggestion would be to use a different CA in the docs/guide/create_authority_options.png
for instance `Cryptography`.

PR #3285, opened 2020-12-04
Corrected delivery of TLS format

sirferl:
While testing we found out that kubernetes expects the intermediate certificates together with the public key in one file when using the "TLS" format.
This change contains the correct code for achieving this !


hosseinsh:
Thanks for the PR @sirferl. merging shortly.

Issue #2105, opened 2018-11-14
Source Plugins: Options really optional?


sirferl:
I am experimenting with Source Plugins.
When I set up a plugin without options section, I got the following error:
_2018-11-13 15:50:44,483 ERROR: 'plugin_options' [in /opt/lemur/lemur/common/schema.py:164]
Traceback (most recent call last):
  File "/opt/lemur/lemur/common/schema.py", line 161, in decorated_function
    resp = f(*args, **kwargs)
  File "/opt/lemur/lemur/sources/views.py", line 154, in post
    return service.create(data['label'], data['plugin']['slug'], data['plugin']['plugin_options'], data['description'])
KeyError: 'plugin_options'_

I added a dummy option to get going.
Can you confirm the non optional nature of options in source plugins?
regards
sirferl

castrapel:
It does look like options are required, even if they are dummy options. All of our sources have options -- Most of them are AWS sources with options being account number and some other attributes.

I'm glad you found a workaround. This task seems relatively low priority for us at the moment so we will mark it as help wanted.

PR #3295, opened 2020-12-08
make options for sources optional

sirferl:
As stated in https://github.com/Netflix/lemur/issues/2105 the options for source where not optional. I changed the code, so when no options are needed for a source plugin, no error is thrown when creating the new source.


hosseinsh:
Awesome, thanks for the PR, and making optional, really optional! :D

PR #3302, opened 2020-12-12
Updating includeCrlDp to includeCDP

csine-nflx:
Makes it more obvious what the variable does, and ensures compatibility with some custom plugins.  It does not appear that any other plugins make use of this option, despite it being passed in options['extensions'] to issuerPlugins via create_certificates.


hosseinsh:
closing this in favor of doing the change in the custom plugin, and avoiding any impact on Lemur API users.

PR #3312, opened 2020-12-15
Expiration summary email fixes

jtschladen:
1. Fix incorrect expectation of return type for summary emails in notification cli
2. Add 1 to days in summary email (as written, it ranges from 0 to 13 instead of 1 to 14)

### DIFF
--- a/lemur/notifications/cli.py
+++ b/lemur/notifications/cli.py
@@ -85,15 +85,13 @@ def security_expiration_summary(exclude):
     status = FAILURE_METRIC_STATUS
     try:
         print("Starting to notify security team about expiring certificates!")
-        success, failed = send_security_expiration_summary(exclude)
+        status = send_security_expiration_summary(exclude)
         print(
-            "Finished notifying security team about expiring certificates! "
-            f"Sent: {success} Failed: {failed}"
+            "Finished notifying security team about expiring certificates!"
         )
-        status = SUCCESS_METRIC_STATUS
     except Exception:
         sentry.captureException()
 
     metrics.send(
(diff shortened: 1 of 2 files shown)


lemur/notifications/cli.py
```diff
@@ -85,12 +85,10 @@ def security_expiration_summary(exclude):
...
     try:
         print("Starting to notify security team about expiring certificates!")
-        success, failed = send_security_expiration_summary(exclude)
+        status = send_security_expiration_summary(exclude)
         print(
-            "Finished notifying security team about expiring certificates! "
-            f"Sent: {success} Failed: {failed}"
+            "Finished notifying security team about expiring certificates!"
         )
-        status = SUCCESS_METRIC_STATUS
```
hosseinsh, line 93: don't we still need this, when `status==True`?
https://github.com/Netflix/lemur/blob/56169656377637b7a0e53514105fbe31b436c4a9/lemur/constants.py#L12

also should this method return false: 
https://github.com/Netflix/lemur/blob/f941383cd2cee0bf8c11350cc39f45cabd44502f/lemur/notifications/messaging.py#L420

jtschladen: Oops, you're right, I was getting `SUCCESS_METRIC_STATUS` and `True` confused, will fix.

Re: "should this method return false" - the existing pattern for the file seems to be return only `True`, and nothing on failure. It's a bit of an odd pattern, but I don't want to rewrite the whole thing now and I'd like to stay consistent with the rest of the methods in the file.

hosseinsh (approved):

PR #2884, opened 2020-01-13
Optimize certificates sql

ilyalabun:
This change gave us 20-30% page load time decrease when searching by certificate name and/or cn

PR #3342, opened 2021-01-05
Cast subquery for postgres planning optimisation

rajatsharma94:
Postgres by default skips index usage for the older subquery. Unless we specifically cast everything into int.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -13,6 +13,7 @@ from cryptography.hazmat.primitives import hashes, serialization
 from flask import current_app
 from sqlalchemy import func, or_, not_, cast, Integer
 from sqlalchemy.sql.expression import false, true
+from sqlalchemy.dialects.postgresql import array, ARRAY
 
 from lemur import database
 from lemur.authorities.models import Authority
@@ -478,7 +479,7 @@ def render(args):
             query = query.filter(
                 or_(
                     func.lower(Certificate.cn).like(term.lower()),
-                    Certificate.id.in_(like_domain_query(term)),
+                    (Certificate.id == cast(array(like_domain_query(term)), ARRAY(Integer)).any_()),
                 )
             )
         elif "id" in terms:


hosseinsh:
Hi @rajatsharma94, thanks for your PR. 

We are noticing a significant performance degradation associated with the array and cast.
To replicate, search for a common pattern, like `com` or `net`.

This is an example for a smaller query:
```
WHERE certificates.not_after > %(not_after_1)s AND (lower(certificates.cn) LIKE %(lower_1)s OR certificates.id = ANY (CAST(ARRAY[%(param_1)s, %(param_2)s, %(param_3)s, %(param_4)s, %(param_5)s, %(param_6)s, %(param_7)s, %(param_8)s, %(param_9)s, %(param_10)s, %(param_11)s, %(param_12)s, %(param_13)s, %(param_14)s, %(param_15)s, %(param_16)s, %(param_17)s, %(param_18)s, %(param_19)s, %(param_20)s, %(param_21)s, %(param_22)s, %(param_23)s, %(param_24)s, %(param_25)s, %(param_26)s, %(param_27)s, %(param_28)s, %(param_29)s, %(param_30)s, %(param_31)s, %(param_32)s, %(param_33)s, %(param_34)s, %(param_35)s, %(param_36)s, %(param_37)s, %(param_38)s, %(param_39)s, %(param_40)s, %(param_41)s, %(param_42)s, %(param_43)s, %(param_44)s, %(param_45)s, %(param_46)s, %(param_47)s, %(param_48)s, %(param_49)s, %(param_50)s, %(param_51)s, %(param_52)s, %(param_53)s, %(param_54)s, %(param_55)s, %(param_56)s, %(param_57)s, %(param_58)s, %(param_59)s, %(param_60)s, %(param_61)s, %(param_62)s, %(param_63)s, %(param_64)s] AS INTEGER[]))) ORDER BY certificates.id DESC 
 LIMIT %(param_65)s OFFSET %(param_66)s
 ```

hosseinsh (changes requested):
please consider a different optimization, to avoid the negative performance penalty.

rajatsharma94:
Hi @hosseinsh, sure. Even after the overhead of `array` and `cast`. This change was making our queries at least 2x faster (including for generic term like `com`). Feel free to close down the MR if that's not the case with your test cases.

Also, I couldn't find any other way of forcing index use. If there's something that you think can have the same effect without the overhead, I'm happy to change this further 🙂

hosseinsh:
Hi @rajatsharma94, we did more testing, but it seems we cannot reproduce the same improvements. are there any other changes on your side which result into the observed improvements, like DB index migrations not checked into the Lemur repo?

we are also keen to further improve the performance, and will certainly share any new findings.

rajatsharma94:
I think there were a few index changes, but those were all checked-in [https://github.com/Netflix/lemur/pull/2884](here).

Maybe it has something to do with the table sizes? We have about 2 million entries in the certificates table, ~3.2 million in the certificate_associations table, and 600k entries in domains, which is making the whole query slow.

hosseinsh:
Hi @rajatsharma94, yeah, those indexes did have measurable impact on performance.
going to close this PR, until we have a better way to ensure a consistent performance improvement.

PR #3346, opened 2021-01-06
Config to change algo to ECC during reissue

charhate:
1. Allow certificates to start using **ECCPRIME256V1** key_type during certificate reissue/rotate. This will apply to certificates with owners configured as part of **ROTATE_TO_ECC_OWNER_LIST** only if the common name is not present in exclusion list **ECC_NON_COMPATIBLE_COMMON_NAMES** (to allow flexibility of making exceptions).

This may not be a widely needed feature, thus may be removed in mid-Q2 2021 (unless identified a use case)

2. Fix @pytest.yield_fixture deprecation warning

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -795,6 +795,15 @@ def reissue_certificate(certificate, replace=None, user=None):
     else:
         primitives["description"] = f"{reissue_message_prefix}{certificate.id}"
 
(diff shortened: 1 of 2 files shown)


hosseinsh (approved):
also thanks for fixing th @pytest.yield_fixture deprecation warnings

PR #3347, opened 2021-01-06
Paginate valid certificate fetch API

charhate:
The API is modified to offer optional pagination. One can send page number(>=1) and desired count per page. The returned data contains total number of certificates which can help in determining the last page. Pagination will not be offered if page or count info is not sent (making it backward compatible) or if it is zero.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -51,17 +51,20 @@ class CertificatesListValid(AuthenticatedResource):
         """
         .. http:get:: /certificates/valid/<query>
 
-           The current list of not-expired certificates for a given common name, and owner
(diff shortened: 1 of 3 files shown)


lemur/certificates/views.py
```diff
@@ -128,10 +131,15 @@ def get(self):
            :statuscode 403: unauthenticated
 
         """
-        parser = paginated_parser.copy()
```
charhate, line 131: The paginated_parser was putting default values for sorting and paging. It was not used eventually in the old code. As commented below, using non-paginated parser to ensure backward compatibility

hosseinsh (approved):
Awesome PR!

PR #3362, opened 2021-01-16
Entrust default client

hosseinsh:
Allow Entrust to use the primary client ID and avoid an extra round for resolving the clientID.

Also this Entrust Endpoint is not reliable, and often results into a Client Reset error.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -941,6 +941,12 @@ The following parameters have to be set in the configuration files.
 
         If there is a config variable ENTRUST_PRODUCT_<upper(authority.name)> take the value as cert product name else default to "STANDARD_SSL". Refer to the API documentation for valid products names.
 
+.. data:: ENTRUST_USE_DEFAULT_CLIENT_ID
+    :noindex:
+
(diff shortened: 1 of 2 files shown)


hosseinsh:
@sirferl any thoughts on this PR?

I am wondering, if using the primary clientID shouldn't be the default behavior, and everyone else should set a config to resolve the clientID.

sirferl:
Hi @hosseinsh, all the best in the new year!
I appreciate your change here.
You are certanly right. There is room for improvement. 
I was in rush, because we had to get that Plugin working in our enviornment.
I will think about a good way to sort out the ClientID and rework that part.

About the reliability: I noticed that one, too.
Sometimes the Entrust-API seems not respond well. Maybe a retry mechanism is called for.

charhate (approved):

hosseinsh:
Hi @sirferl, thank you. 
Wish you also all the best for 2021!

I have been in touch with Entrust folks, and sharing with them some feedback with respect to this endpoint. 
and good point, we should be able to improve the experience on our side with retries.

charhate (approved):

PR #3363, opened 2021-01-16
Entrust: respecting Entrust's rate limit

hosseinsh:
deactivate has a 10 requests per 30s rate limit

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -11,6 +11,7 @@ from flask_principal import Identity, identity_changed
 from flask_script import Manager
 from sqlalchemy import or_
 from tabulate import tabulate
+from time import sleep
 
 from lemur import database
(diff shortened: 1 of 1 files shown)


lemur/certificates/cli.py
```diff
@@ -761,7 +762,10 @@ def deactivate_entrust_certificates():
 
     certificates = get_all_valid_certs(['entrust-issuer'])
     entrust_plugin = plugins.get('entrust-issuer')
-    for cert in certificates:
+    for index, cert in enumerate(certificates, start=1):
```
charhate, line 765: we can have start=0, so that it will deactivate 10 certs in each batch. Else it will be 9 in first batch and 10 for the rest.

hosseinsh: good catch!

charhate (approved):

PR #3384, opened 2021-01-28
Audit logs for Role, User and API Key

charhate:
Adding logs for API key, User and Role create/update/delete. These are info logs.
```
[2021-01-27 18:59:10,882] INFO in service: [lemur-audit] action: unassign_role, user: user@test.com, entity: admin, details: Un-assigning the role for user charhate
[2021-01-28 18:00:50,048] INFO in service: [lemur-audit] action: create_api_key, user: user@test.com, entity: testAPIKeyLog, details: Creating the API key ApiKey(name=testAPIKeyLog, user_id=164, ttl=-1, issued_at=1611910187, revoked=False)

action: The action being logged e.g. assign_role, create_role etc
user: user performing the action
entity: The entity undergoing the action e.g. name of the role
details: additional description
```

The log format is similar to [certificate update logs](https://github.com/Netflix/lemur/blob/master/lemur/logs/service.py#L27).

### DIFF
--- a/lemur/api_keys/service.py
+++ b/lemur/api_keys/service.py
@@ -7,6 +7,7 @@
 """
 from lemur import database
 from lemur.api_keys.models import ApiKey
+from lemur.logs import service as log_service
 
 
 def get(aid):
@@ -24,6 +25,7 @@ def delete(access_key):
     :param access_key:
     :return:
     """
+    log_service.audit_log("delete_api_key", access_key.name, "Deleting the API key")
     database.delete(access_key)
 
 
@@ -34,8 +36,9 @@ def revoke(aid):
     :return:
     """
     api_key = get(aid)
-    setattr(api_key, "revoked", False)
+    setattr(api_key, "revoked", True)
 
+    log_service.audit_log("revoke_api_key", api_key.name, "Revoking API key")
     return database.update(api_key)
 
 
@@ -55,6 +58,9 @@ def create(**kwargs):
     :return:
     """
     api_key = ApiKey(**kwargs)
+    # this logs only metadata about the api key
+    log_service.audit_log("create_api_key", api_key.name, f"Creating the API key {api_key}")
+
     database.create(api_key)
     return api_key
 
@@ -69,6 +75,7 @@ def update(api_key, **kwargs):
     for key, value in kwargs.items():
         setattr(api_key, key, value)
 
+    log_service.audit_log("update_api_key", api_key.name, f"Update summary - {kwargs}")
     return database.update(api_key)
 
 
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -197,7 +197,6 @@ def update_user(user, profile, roles):
     :param profile:
     :param roles:
     """
-    log_service.audit_log("TEST", user.name, "Edit role")
     # if we get an sso user create them an account
     if not user:
         user = user_service.create(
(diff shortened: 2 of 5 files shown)


lemur/auth/views.py
```diff
@@ -198,7 +197,7 @@ def update_user(user, profile, roles):
     :param profile:
     :param roles:
     """
-
+    log_service.audit_log("TEST", user.name, "Edit role")
```
hosseinsh, line 200: is this intentional?

lemur/auth/views.py
```diff
@@ -184,8 +185,6 @@ def create_user_roles(profile):
                 current_app.config["LEMUR_DEFAULT_ROLE"],
                 description="This is the default Lemur role.",
             )
-        if not default.third_party:
```
hosseinsh, line 187: are we removing this part?

hosseinsh (commented):
nice work!

lemur/auth/views.py
charhate: Thanks for catching, removed it.

lemur/auth/views.py
charhate: Added it back since it may have side-effect of adding default role with each login

lemur/auth/views.py
```diff
@@ -215,6 +215,8 @@ def update_user(user, profile, roles):
         for ur in user.roles:
             if not ur.third_party:
                 roles.append(ur)
+            else:
+                log_service.audit_log("unassign_role", ur.name, f"Un-assigning the role for {user.name}")
```
hosseinsh, line 219: this might be very verbose, shall only log those which will be truly un-assigned?

like:
```
elif ur not in roles
    log_service.audit_log("unassign_role", ur.name, f"Un-assigning the role for {user.name}")
```

lemur/roles/service.py
```diff
@@ -44,6 +47,8 @@ def set_third_party(role_id, third_party_status=False):
     role = get(role_id)
     role.third_party = third_party_status
     database.update(role)
+
+    log_service.audit_log("update_role", role.name, f"Updated third_party_status={third_party_status}")
```
hosseinsh, line 51: isn't this very verbose? basically all GG for each login per each user?

charhate: This will be maximum once per role, so once per GG or email. `set_third_party` is called only if existing role is not marked as third-party.
Most GG will be auto created by lemur as third party ([during create](https://github.com/Netflix/lemur/blob/master/lemur/auth/views.py#L151-L159) this flag is set than setting separately). So I believe `set_third_party` is not used frequently.

lemur/auth/views.py
charhate: That's a good catch! I'll also change this to log single message with a list of removed roles.

lemur/api_keys/service.py
```diff
@@ -34,8 +36,9 @@ def revoke(aid):
     :return:
     """
     api_key = get(aid)
-    setattr(api_key, "revoked", False)
+    setattr(api_key, "revoked", True)
```
charhate, line 39: Happened to observe this while working on log changes. This method is used only via CLI.
The revoke via lemur UI calls update method instead, so we may have never seen issue with revoke.

hosseinsh (approved):
thanks for the extensive PR!

PR #3385, opened 2021-01-29
Avoid Failure when Rotating Endpoint with multiple certificates

hosseinsh:
Lemur currently gives up, when rotating an endpoint with multiple new certificates to replace the old one. 
In case of duplicate certificates replacing an expiring one, Lemur shall select the new certificate at index 0, basically the first available certificate.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -231,7 +231,7 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
             print(
                 f"[+] Rotating endpoint: {endpoint.name} to certificate {new_cert.name}"
             )
-            log_data["message"] = "Rotating endpoint"
+            log_data["message"] = "Rotating one endpoint"
             log_data["endpoint"] = endpoint.dnsname
             log_data["certificate"] = new_cert.name
             request_rotation(endpoint, new_cert, message, commit)
@@ -425,7 +425,6 @@ def rotate_region(endpoint_name, new_certificate_name, old_certificate_name, mes
                         1,
                         metric_tags={
                             "region": region,
-                            "old_certificate_name": str(old_cert),
                             "new_certificate_name": str(endpoint.certificate.replaced[0].name),
                             "endpoint_name": str(endpoint.dnsname),
                         },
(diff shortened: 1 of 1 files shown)


lemur/certificates/cli.py
```diff
@@ -244,41 +249,24 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
                 current_app.logger.info(log_data)
 
         else:
+            # no certificate name or endpoint was provided, so we will now fetch all endpoints,
+            # which have are attached to a certificate that has been replaced
```
charhate, line 253: `which have are attached` --> `which are attached`

lemur/certificates/cli.py
```diff
@@ -244,41 +249,24 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
-                    )
+                log_data["message"] = "Rotating endpoint from old to new cert"
+                if len(endpoint.certificate.replaced) > 1:
+                    log_data["message"] = f"Multiple replacement certificates found, going with the first one out of " \
+                                          f"{len(endpoint.certificate.replaced)}"
+
+                log_data["endpoint"] = endpoint.dnsname
+                log_data["certificate"] = endpoint.certificate.replaced[0].name
+                request_rotation(endpoint, endpoint.certificate.replaced[0], message, commit)
+                print(log_data)
```
charhate, line 265: Is explicit `print(log_data)` given `current_app.logger.info(log_data)`

hosseinsh: good point, we can skip them.

lemur/certificates/cli.py
```diff
@@ -418,32 +406,27 @@ def rotate_region(endpoint_name, new_certificate_name, old_certificate_name, mes
...
+                log_data["certificate"] = endpoint.certificate.replaced[0].name
+                log_data["message"] = "Rotating all endpoints in region"
+                if len(endpoint.certificate.replaced) > 1:
+                    log_data["message"] = f"Multiple replacement certificates found, going with the first one out of " \
+                                          f"{len(endpoint.certificate.replaced)}"
+
+                request_rotation(endpoint, endpoint.certificate.replaced[0], message, commit)
+                current_app.logger.info(log_data)
 
                 metrics.send(
```
charhate, line 424: Are the metrics missing in above if and elif?

hosseinsh: no, they follow later, line 467

charhate (approved):

PR #3386, opened 2021-01-29
docs: update OS and deps versions

sergerdn:
Docs how to install python 3.7 and Node js 10.x are missing. Because writing docs in English not easy for me.


hosseinsh (approved):
Thanks @sergerdn for the great PR, and improving the quality of Lemur docs!

I just removed `or greater` from Node, since we haven't tested that yet, and past experience has pointed to some difficulties.

PR #3390, opened 2021-01-29
Acme reduce attempts

hosseinsh:
ACME plugin attempts to resolve a pending Certificate up to 5 times. 
This happens to also be the rate limit of LE for duplicate certificates. 
Each attempt to resolve a pending cert consists of a new cert requesting, the DNS challenge dance, and attempt to finalize the order. 

This would mean in case there is any issues with Lemur, it can quickly get a domain blocked for issuance for 7 days. Though there are workaround for that (adding a SAN), ideally we should not consume all 5 attempts at once. 

This PR is setting the number of attempts to 3. We can consider to make this variable configurable, if there are other use-cases demanding it.

### DIFF
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -274,7 +274,8 @@ def fetch_acme_cert(id):
         real_cert = cert.get("cert")
         # It's necessary to reload the pending cert due to detached instance: http://sqlalche.me/e/bhk3
         pending_cert = pending_certificate_service.get(cert.get("pending_cert").id)
-        if not pending_cert:
+        if not pending_cert or pending_cert.resolved:
+            # pending_cert is cleared or it was resolved by another process
             log_data[
                 "message"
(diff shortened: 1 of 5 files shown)


lemur/common/celery.py
```diff
@@ -273,7 +274,8 @@ def fetch_acme_cert(id):
         real_cert = cert.get("cert")
         # It's necessary to reload the pending cert due to detached instance: http://sqlalche.me/e/bhk3
         pending_cert = pending_certificate_service.get(cert.get("pending_cert").id)
-        if not pending_cert:
+        if not pending_cert or pending_cert.resolved:
```
hosseinsh, line 277: I am hoping `pending_cert.resolved` would address the issue of 1 duplicate certificate. 

This seems to be a corner case, since currently cert issuance triggers `fetch_acme_cert ` and also celery's schedule. If they happen to be both within the same time range, celery wont notice that the pending certificate is resolved.

charhate (approved):

PR #3416, opened 2021-02-11
ECCPRIME256V1 as default for cert create API

charhate:


lemur/certificates/schemas.py
```diff
@@ -160,7 +160,7 @@ def load_data(self, data):
             if data.get("body"):
                 data["key_type"] = utils.get_key_type_from_certificate(data["body"])
             else:
-                data["key_type"] = "RSA2048"  # default value
+                data["key_type"] = "ECCPRIME256V1"  # default value
```
hosseinsh, line 163: I might be forgetting the reasoning behind why we had a default `key_type` here, when a CSR and cert body are missing?

hosseinsh (approved):

lemur/certificates/schemas.py
charhate: To my best recollection, this code was added to fix key_type during import.
Since this line is in `@pre_load`, it seems redundant given `missing="ECCPRIME256V1"` option when we define `key_type `. Let me check history of this file. Nevertheless, it is safe to keep it here for now and remove later.

PR #3422, opened 2021-02-17
Fix TTL calculation for API keys

jtschladen:


lemur/auth/service.py
```diff
@@ -75,7 +75,7 @@ def create_token(user, aid=None, ttl=None):
         if ttl == -1:
             del payload["exp"]
         else:
-            payload["exp"] = ttl
+            payload["exp"] = datetime.utcnow() + timedelta(days=ttl)
```
jtschladen, line 78: I ended up needing to make this a normal expiration time since the JWT library we use to decode the token (on line 109) correctly interprets it as a timestamp and was failing out even with the modified check on line 120. In fact, I believe the check on line 120 is redundant, but I feel safer leaving it than removing it at this point.

jtschladen: I was mistaken, that check is not technically redundant. The JWT has an immutable timestamp (since it's signed), but the API key in the DB can technically be changed; I was able to test expiration by modifying the DB API key timestamp but using a technically-still-valid JWT; this exercises the line 120 check.

hosseinsh (approved):

PR #3428, opened 2021-02-20
Add/remove certs for notification UI

jtschladen:
This fixes the pre-existing bug where submitting a notification edit dialog when only a subset of attached certs are shown causes all the remaining certs to be dropped. For example, prior to this change, it was effectively impossible to edit any default notification because it would drop all but 10 certs from that notification.


hosseinsh (approved):

lemur/static/app/angular/notifications/services.js
```diff
@@ -8,10 +8,35 @@ angular.module('lemur')
           if (this.certificates === undefined) {
             this.certificates = [];
           }
+          if (this.addedCertificates === undefined) {
+            this.addedCertificates = [];
+          }
+          if (_.some(this.addedCertificates, function (cert) {
+            return cert.id === certificate.id;
+          })) {
+            return;
```
charhate, line 17: Do you see any need to check if this cert is in `removedCertificates`, just like `line 21-26`, before returning from here?

jtschladen: My feeling is no. To elaborate, here are some possible orders of operations:

- add: we wouldn't hit this `return` since it's not present in `addedCertificates`
- add -> add: we'll hit this `return`, but don't need to worry about `removedCertificates`
- add -> remove -> add: we won't hit this `return`, since the remove call will have removed the cert from `addedCertificates` (lines 34-39)
- remove -> add -> add: we'll hit this `return`, but we can assume the cert isn't in `removedCertificates` because the last time they added it would have removed it from `removedCertificates` (lines 21-26)

The other key is that you can't ever remove twice in a row, since the "remove" button will go away.

Of course you can extend these pattern infinitely, but I think the same set of base cases should always be true, and we should be okay without checking `removedCertificates` when adding a second time.

charhate (approved):

PR #3439, opened 2021-02-24
Include notification interval (converted to days) in SNS notifications

jtschladen:
Updates the SNS notification format to include a new field `notification_interval_days` to specify the interval that triggered the notification.

### DIFF
--- a/lemur/plugins/lemur_aws/sns.py
+++ b/lemur/plugins/lemur_aws/sns.py
@@ -12,20 +12,21 @@ import boto3
 from flask import current_app
 
 
-def publish(topic_arn, certificates, notification_type, **kwargs):
+def publish(topic_arn, certificates, notification_type, options, **kwargs):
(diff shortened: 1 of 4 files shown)


lemur/plugins/lemur_aws/sns.py
```diff
@@ -57,4 +58,25 @@ def format_message(certificate, notification_type):
...
+        return interval * 7
+
+    elif unit == "months":
+        return interval * 30
+
+    elif unit == "days":
+        return interval
+
+
+def get_option(options, option_name):
```
hosseinsh, line 79: you might have already considered this, but would it be convenient to employ this:

https://github.com/Netflix/lemur/blob/98668b09722695c5c4f353c122359383edb1c811/lemur/plugins/base/v1.py#L113

hosseinsh (approved):

charhate (approved):

PR #3487, opened 2021-03-17
codeowners

hosseinsh:
tag lemur squad for review

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# These owners will be the default owners for everything in the repo.
+*                            @hosseinsh, @csine-nflx, @charhate, @jtschladen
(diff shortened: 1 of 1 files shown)


charhate (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# These owners will be the default owners for everything in the repo.
+*                            @hosseinsh, @csine-nflx, @charhate, @jtschladen
```
jtschladen, line 2: Should this use spaces rather than commas? [Doc](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) is unclear but the one example with multiple usernames has spaces

hosseinsh: you are right, I see spaces in the examples

jtschladen (approved):

PR #3489, opened 2021-03-17
Change ssh missing host key policy used by sftp plugin from AutoAddPolicy

jtschladen:
This change is a result of this CodeQL error: 
https://github.com/Netflix/lemur/security/code-scanning/8?query=ref%3Arefs%2Fheads%2Fmaster

The comment above the line flagged by the scanner seems to indicate that allowing unknown hosts is intentional, but this isn't ideal from a security perspective.

### DIFF
--- a/lemur/plugins/lemur_sftp/plugin.py
+++ b/lemur/plugins/lemur_sftp/plugin.py
@@ -113,7 +113,7 @@ class SFTPDestinationPlugin(DestinationPlugin):
             ssh = paramiko.SSHClient()
 
             # allow connection to the new unknown host
-            ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+            ssh.set_missing_host_key_policy(paramiko.RejectPolicy())
 
             # open the ssh connection
             if password:


lemur/plugins/lemur_sftp/plugin.py
```diff
@@ -113,7 +113,7 @@ def open_sftp_connection(self, options):
             ssh = paramiko.SSHClient()
 
             # allow connection to the new unknown host
-            ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+            ssh.set_missing_host_key_policy(paramiko.RejectPolicy())
```
jtschladen, line 116: @dmitryzykov as the original plugin author, do you have any input on whether `AutoAddPolicy` was really required?

hosseinsh: tagging @peschmae to also weigh in, since he has contributed to this plugin, and use it as part of the http01 flow. 
if no objections raised, I would suggest, we move forward with merging this PR.

peschmae (approved):
Lgtm

PR #3492, opened 2021-03-18
Add config option DISABLE_NOTIFICATION_PLUGINS to disable specific notification plugins

jtschladen:
Provides a new config option `DISABLE_NOTIFICATION_PLUGINS` that disables *sending* notifications via the specified plugins. This does not prevent notifications from being *configured* with those plugins, but any configured ones will not be sent.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -404,13 +404,14 @@ The following configuration options are supported:
 
        ::
 
-          DISABLE_NOTIFICATION_PLUGINS = ['exclude', 'also exclude']
+          EXCLUDE_CN_FROM_NOTIFICATION = ['exclude', 'also exclude']
 
 
 .. data:: DISABLE_NOTIFICATION_PLUGINS
     :noindex:
 
     Specifies a set of notification plugins to disable. Notifications will not be sent using these plugins. Currently only applies to expiration notifications, since they are the only type that utilize plugins.
+    This option may be particularly useful in a test environment, where you might wish to enable the notification job without actually sending notifications of a certain type (or all types).
 
     .. note::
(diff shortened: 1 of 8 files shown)


docs/administration.rst
```diff
@@ -390,6 +390,36 @@ This notification type also supports the same ``--exclude`` and ``EXCLUDE_CN_FRO
...
+
+       ::
+
+          DISABLE_NOTIFICATION_PLUGINS = ['exclude', 'also exclude']
+
+
+.. data:: DISABLE_NOTIFICATION_PLUGINS
+    :noindex:
+
+    Specifies a set of notification plugins to disable. Notifications will not be sent using these plugins. Currently only applies to expiration notifications, since they are the only type that utilize plugins.
```
hosseinsh, line 413: shall we add that this could be off value in the test environment, where you would want to trigger and test the notification job, but not actually send out the notification.

lemur/notifications/messaging.py
```diff
@@ -241,32 +241,43 @@ def send_expiration_notifications(exclude):
...
+                        certificate
+                    ).data
+                    notification_data.append(cert_data)
+
+                email_recipients = notification.plugin.get_recipients(notification.options, security_email + [owner])
+                # Plugin will ONLY use the provided recipients if it's email; any other notification plugin ignores them
+                if send_plugin_notification(
+                        "expiration", notification_data, email_recipients, notification
                 ):
-                    success = 1 + len(email_recipients)
```
hosseinsh, line 267: good catch

hosseinsh (approved):
appreciate the extensive tests!  ＼(๏◡๏)／

docs/administration.rst
jtschladen: Good idea, added!

PR #3498, opened 2021-03-23
Upgrade virtualenv to allow twine upgrade

jtschladen:


requirements-dev.txt
```diff
@@ -87,14 +89,16 @@ toml==0.10.0
     # via pre-commit
 tqdm==4.45.0
     # via twine
-twine==3.3.0
+twine==3.4.1
     # via -r requirements-dev.in
 urllib3==1.25.8
     # via requests
-virtualenv==20.0.17
+virtualenv==20.4.3
```
jtschladen, line 96: I'm unsure where the prior value of 20.0.17 came from; I suppose since not a direct dependency, it just never got updated since it was first pulled in?

hosseinsh (approved):

requirements-dev.txt
hosseinsh: yeah, would have expected the dependency update to have caught this

jtschladen: I'll go ahead and merge and hope it's not overwritten with the old version on the next dependabot run. If it is, I'll take another look to see why.

PR #3506, opened 2021-03-30
Fix usage of groupby in notifications logic

jtschladen:
`groupby` requires the collection to be pre-sorted, as it simply groups the collection by contiguous chunks. Therefore, the prior logic would only include the _latest_ contiguous chunk of certs for each `groupby` operation performed. This changed sorts before calling `groupby` to ensure this isn't a problem.

Reference: https://docs.python.org/3/library/itertools.html#itertools.groupby
> Make an iterator that returns consecutive keys and groups from the _iterable_. The _key_ is a function computing a key value for each element. If not specified or is `None`, _key_ defaults to an identity function and returns the element unchanged. Generally, the iterable needs to already be sorted on the same key function.
> 
> The operation of `groupby()` is similar to the `uniq` filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function). That behavior differs from SQL’s GROUP BY which aggregates common elements regardless of their input order.

### DIFF
--- a/lemur/notifications/messaging.py
+++ b/lemur/notifications/messaging.py
@@ -135,7 +135,7 @@ def get_eligible_certificates(exclude=None):
     certs = get_certificates(exclude=exclude)
 
     # group by owner
-    for owner, items in groupby(certs, lambda x: x.owner):
+    for owner, items in groupby(sorted(certs, key=lambda x: x.owner), lambda x: x.owner):
         notification_groups = []
 
         for certificate in items:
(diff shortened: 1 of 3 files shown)


lemur/notifications/messaging.py
```diff
@@ -146,22 +146,27 @@ def get_eligible_certificates(exclude=None):
...
+    for interval, interval_certs in groupby(sorted(all_certs, key=lambda x: (x.not_after - now).days),
+                                            lambda x: (x.not_after - now).days):
+        cert_data = []
+        for certificate in interval_certs:
+            cert_data.append(certificate_notification_output_schema.dump(certificate).data)
+        interval_data = {"interval": interval, "certificates": cert_data}
+        message_data.append(interval_data)
 
-    return certificates
+    return message_data
```
jtschladen, line 169: I refactored this to make it more testable.

hosseinsh (approved):
excellent testing! awesome work!

charhate (approved):

PR #3508, opened 2021-04-01
Remove external_id requirement to revoke certificate

csine-nflx:
Certificate providers may use certificate.serial instead of certificate.external_id to identify certificates to revoke.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -1471,9 +1471,6 @@ class CertificateRevoke(AuthenticatedResource):
                     403,
                 )
 
-        if not cert.external_id:
-            return dict(message="Cannot revoke certificate. No external id found."), 400
-
         if cert.endpoints:
             for endpoint in cert.endpoints:
                 if service.is_attached_to_endpoint(cert.name, endpoint.name):


hosseinsh:
Looks good to me. At the level of plugin, we could then potentially check if external_id is missing, where applicable.


https://github.com/Netflix/lemur/blob/56169656377637b7a0e53514105fbe31b436c4a9/lemur/plugins/lemur_digicert/plugin.py#L377

https://github.com/Netflix/lemur/blob/47d617f066014faa9f96b420230acadaffa62d61/lemur/plugins/lemur_entrust/plugin.py#L279

csine-nflx:
> Looks good to me. At the level of plugin, we could then potentially check if external_id is missing, where applicable.
> 
> https://github.com/Netflix/lemur/blob/56169656377637b7a0e53514105fbe31b436c4a9/lemur/plugins/lemur_digicert/plugin.py#L377
> 
> https://github.com/Netflix/lemur/blob/47d617f066014faa9f96b420230acadaffa62d61/lemur/plugins/lemur_entrust/plugin.py#L279

I don't think this is needed, as it should not happen. In all cases where a particular plugin is used to revoke a certificate, the certificate was created with that same plugin and the external_id was created. If during creation the certificate returned an external_id, then it will be stored and available for revocation.  This holds true for the cfssl, digicert, and entrust plugins.  Is there a use case where a certificate would be revoked via a plugin but was not created by that plugin?

hosseinsh:
> Is there a use case where a certificate would be revoked via a plugin but was not created by that plugin?
The only use-cases that come to mind are discovered certificates and imported certificates.

In both cases, the `external_id` is missing, which makes revocation not possible. In addition to that, such certificates would also be missing the `authority_id`, which makes finding the plugin not possible. 

For the class of certificates, where `external_id` and `authority_id` is missing, the error is likely to be caught here:
https://github.com/Netflix/lemur/blob/3783fbeaa1645bbee022827f4f53ffb12dd65a61/lemur/certificates/views.py#L1487-L1498

I believe the `verisign_plugin` is also an exception here, where issued certificates do not exhibit an external_id; but that plugin is deprecated and doesn't offer revocation.

PR #3529, opened 2021-04-16
Expanding audit logs

charhate:
Additional audit logs similar to https://github.com/Netflix/lemur/pull/3384
Moved few delete operations inside None check, no behavior change seen with that.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -694,6 +694,9 @@ class CertificatePrivateKey(AuthenticatedResource):
         response = make_response(jsonify(key=cert.private_key), 200)
         response.headers["cache-control"] = "private, max-age=0, no-cache, no-store"
         response.headers["pragma"] = "no-cache"
+
+        log_service.audit_log("export_private_key", cert.name,
+                              "Private key exported the certificate")
         return response
 
 
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -696,7 +696,7 @@ class CertificatePrivateKey(AuthenticatedResource):
         response.headers["pragma"] = "no-cache"
 
         log_service.audit_log("export_private_key", cert.name,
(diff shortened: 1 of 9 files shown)


lemur/certificates/views.py
```diff
@@ -694,6 +694,9 @@ def get(self, certificate_id):
         response = make_response(jsonify(key=cert.private_key), 200)
         response.headers["cache-control"] = "private, max-age=0, no-cache, no-store"
         response.headers["pragma"] = "no-cache"
+
+        log_service.audit_log("export_private_key", cert.name,
+                              "Private key exported for the certificate")
```
hosseinsh, line 699: shall we use a consistent language? like `Exported private key.`

lemur/sources/service.py
```diff
@@ -292,7 +294,10 @@ def delete(source_id):
 
     :param source_id: Lemur assigned ID
     """
-    database.delete(get(source_id))
+    source = get(source_id)
+    if source:
+        log_service.audit_log("delete_source", source.label, "Deleting source")
+    database.delete(source)
```
hosseinsh, line 300: should this also be part of `if source:`?

lemur/certificates/views.py
hosseinsh: specially looking how the log will appear we can go with a common theme:
https://github.com/Netflix/lemur/blob/3783fbeaa1645bbee022827f4f53ffb12dd65a61/lemur/logs/service.py#L47

hosseinsh (approved):

lemur/sources/service.py
charhate: Good call, I lost track of this with some other changes. I have confirmed that the original code did not throw any error is input to delete was `None`, thus moving delete inside if should be safe. It will not silence any past error.

PR #3535, opened 2021-04-19
Confirmation before delete

charhate:
The `Remove` option on UI for API Keys, DNS Provider, Role etc does not ask for confirmation. One can accidentally click on Remove in the wrong row from the table and end up deleting wrong record. These changes will add a minimal confirmation box allowing to confirm the action. I've added the name/label of entity being removed to the message to make it more user friendly.
This change also allows adding confirmation to any action, simply add html attribute `confirm-click="<desired message here>"` to the submit button.


jtschladen (approved):
Very helpful change, thank you!

PR #3536, opened 2021-04-20
Add notification emails for expiring deployed certificates

jtschladen:
Adds a new notification type (for Lemur security team only) that notifies when an expiring certificate is still detected at any of its associated domains.

I also considered adding this logic to the regular expiration notification and notification summary emails; it's possible to do this, but I'm a little hesitant about how much additional load it might cause to add it in multiple places.


lemur/tests/test_messaging.py
```diff
@@ -225,6 +225,30 @@ def test_send_authority_expiration_notifications():
     assert send_authority_expiration_notifications() == (2, 0)
 
 
+@mock_ses
+def test_send_expiring_deployed_certificate_notifications():
```
jtschladen, line 229: I don't particularly like how I've written this test, but I also haven't come up with a better way to do it. At the moment, it makes an actual network call to these domains to check for their current certificates. In order to get the notification to actually send, I need to make the code believe a cert in the DB matches a cert deployed somewhere - unless I were to mock the TLS cert download logic, which I'm not sure how to do in this code base.

I could also remove the test, though I don't love that option either.

Suggestions welcome.

hosseinsh: yeah, you are right, the actual network call might raise some eyebrows.

would it be worthwhile to mock the server?
https://dev.to/bowmanjd/two-methods-for-testing-https-api-calls-with-python-and-pytest-and-also-communicating-with-the-in-laws-388n

and serve a test certificates?
https://github.com/Netflix/lemur/blob/3783fbeaa1645bbee022827f4f53ffb12dd65a61/lemur/tests/vectors.py

hosseinsh (commented):
wow, awesome work!
only some minor ideas about code structure, and potentially a way to avoid a network call during the tests

jtschladen:
Still in progress; I will split up the current logic into two separate celery tasks in order to decouple the domain/port detection from the notification.

lemur/tests/test_endpoints.py
```diff
@@ -32,7 +32,7 @@ def test_rotate_certificate(client, source_plugin):
 )
 def test_endpoint_get(client, token, status):
     assert (
-        client.get(api.url_for(Endpoints, endpoint_id=2), headers=token).status_code
+        client.get(api.url_for(Endpoints, endpoint_id=999), headers=token).status_code
```
jtschladen, line 35: I changed this because the test started returning 200 instead of 404; I'm assuming some other test changes I made (that add more endpoints) caused ID 2 to be valid.

jtschladen:
Should be ready for review now.

lemur/plugins/lemur_email/templates/expiring_deployed_certificate.html
```diff
@@ -0,0 +1,186 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
```
jtschladen, line 1: Screen Shot 2021-05-18 at 12 21 04 PM

lemur/certificates/service.py
```diff
@@ -1005,3 +1009,105 @@ def remove_destination_association(certificate, destination):
...
+    matched_domains = defaultdict(list)
+    # filter out wildcards, we can't check them
+    for cert_association in [assoc for assoc in certificate.certificate_associations if '*' not in assoc.domain.name]:
+        domain = cert_association.domain
+        matched_ports_for_domain = []
+        for port in current_app.config.get("LEMUR_PORTS_FOR_DEPLOYED_CERTIFICATE_CHECK", [443]):
+            try:
+                parsed_serial = parse_serial(get_certificate_via_tls(domain.name, port,
+                                                                     timeout_seconds_per_network_call))
+                if parsed_serial == int(certificate.serial):
```
hosseinsh, line 1074: I wonder if we can do something valuable with the information about the retrieved certificate. 
For instance, search for the existing cert in Lemur and update it.

In case the cert is not in Lemur, likely the case for self-signed certs and instance certs, I am wondering if emitting the certificate in the logs via some interesting properties, like CN, issuer, expiration date, would be a ground to alert on expired certs. 

This might complicate the current PR, and we can think about it in the future.

lemur/certificates/service.py
```diff
@@ -1005,3 +1009,105 @@ def remove_destination_association(certificate, destination):
...
+                    matched_ports_for_domain.append(port)
+            except Exception:
+                current_app.logger.info(f'Unable to check certificate for domain {domain} on port {port}',
+                                        exc_info=True)
+        if len(matched_ports_for_domain) > 0:
+            matched_domains[domain.name] = matched_ports_for_domain
+            # Update the DB
```
hosseinsh, line 1081: would it help to have the DB update take place here?

hosseinsh (commented):
wow, this PR is just amazing, with great Test cases!
Thank you!

lemur/certificates/service.py
jtschladen: The DB update is happening here for the individual cert (line 1082 `cert_association.ports = matched_ports_for_domain` is the update), but I placed the commit outside after _all_ the certs are modified. I'm not actually sure if it's preferable to commit once or many times; if there's not reason to avoid many individual commits, then committing here would be "safer" in that there's less risk of a single cert preventing all the others from being updated.

lemur/certificates/service.py
jtschladen: Yeah, that could make sense. I'll table it for now and we can consider additional enhancements in the future.

lemur/common/utils.py
```diff
@@ -381,3 +384,37 @@ def convert_pkcs7_bytes_to_pem(certs_pkcs7):
...
+    context.check_hostname = False  # we don't care about validating the cert
+    context.verify_mode = ssl.CERT_NONE  # we don't care about validating the cert; it may be self-signed
+    conn = socket.create_connection((host, port), timeout=timeout)
+    sock = context.wrap_socket(conn, server_hostname=host)
+    sock.settimeout(timeout)
+    try:
+        der_cert = sock.getpeercert(True)
+    finally:
+        sock.close()
+    return ssl.DER_cert_to_PEM_cert(der_cert)
```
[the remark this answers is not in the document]
jtschladen: This error will be raised:
```
TypeError: a bytes-like object is required, not 'NoneType'
```
It would then be caught caught on line 1082 of `service.py`, and we'll simply log that it couldn't be checked and skip it.

lemur/certificates/service.py
jtschladen: I just moved it to commit after every update, but let me know if you have other thoughts.

hosseinsh: yeah, +1 one committing more granularly, to avoid losing the entire state.

hosseinsh (approved):

PR #3542, opened 2021-04-28
Task to disable auto-rotation of duplicate certificates

charhate:
Async cert issuance sometimes results in duplicate certificate creation give the retry logic. These certificates keep gettin reissued. The task attempts to find duplicate certificates for configured authorities `AUTHORITY_TO_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES ` and turns off auto-rotation for duplicates if no endpoint attached to the cert leaving at least one cert to be auto-rotated. To identify certs to process in this task, it uses the naming convention by Lemur which appends the serial number to keep cert names unique. The certs with same prefix followed by different serial numbers help to identify duplicate candidates to do further decision making.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -16,6 +16,7 @@ from time import sleep
 from lemur import database
 from lemur.authorities.models import Authority
 from lemur.authorities.service import get as authorities_get_by_id
+from lemur.authorities.service import get_authorities_by_name
 from lemur.certificates.models import Certificate
 from lemur.certificates.schemas import CertificateOutputSchema
 from lemur.certificates.service import (
@@ -797,7 +798,7 @@ def disable_rotation_of_duplicate_certificates(commit):
 
     log_data["authorities"] = authority_names
 
-    authority_ids = [a.id for a in Authority.query.filter(Authority.name.in_(authority_names)).all()]
+    authority_ids = [a.id for a in get_authorities_by_name(authority_names)]
     duplicate_candidate_certs = list_duplicate_certs_by_authority(authority_ids)
 
     log_data["certs_with_serial_number_count"] = len(duplicate_candidate_certs)
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -896,9 +896,9 @@ def process_duplicates(duplicate_candidate_cert, skipped_certs, rotation_disable
                          )
         else:
             # disable rotation and update DB
-            # matching_cert.rotation = False
-            # if commit:
-                # database.update(matching_cert)
+            matching_cert.rotation = False
+            if commit:
+                database.update(matching_cert)
             rotation_disabled_certs.append(matching_cert.name)
             metrics.send("disable_rotation_duplicates", "counter", 1,
                          metric_tags={"status": "success", "certificate": matching_cert.name}
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -27,6 +27,8 @@ from lemur.certificates.service import (
     get,
     get_all_certs_attached_to_endpoint_without_autorotate,
     revoke as revoke_certificate,
+    list_duplicate_certs_by_authority,
+    get_certificates_with_same_prefix_with_rotate_on
 )
 from lemur.certificates.verify import verify_string
 from lemur.constants import SUCCESS_METRIC_STATUS, FAILURE_METRIC_STATUS, CRLReason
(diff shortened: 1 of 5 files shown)


lemur/certificates/cli.py
```diff
@@ -775,3 +777,146 @@ def deactivate_entrust_certificates():
...
+            fallback_cert_to_rotate = name_without_serial_num
+
+        if matching_cert.name == duplicate_candidate_cert.name:
+            # Same cert, no need to compare
+            continue
+
+        # Even if one of the cert with same prefix has different details, skip this set of certs
+        # it's safe to do so and this logic can be revisited
+        if not is_duplicate(matching_cert, duplicate_candidate_cert):
+            skip_cert = True
```
hosseinsh, line 876: would it help to just drop the non-duplicate cert from `certs_with_same_prefix`?

I am thinking about the edge case of:

foo.net
foo.net-1
foo.net-2
foo.net-3
foo.net-4
foo.net-non-duplicate

It is certainly okay to not optimize for this edge case.

lemur/certificates/cli.py
```diff
@@ -775,3 +777,146 @@ def deactivate_entrust_certificates():
...
+    if skip_cert:
+        return False, certs_with_same_prefix
+
+    # Find certs with endpoint, auto-rotate needs to be on for these
+    certs_with_endpoint = []
+    for matching_cert in certs_with_same_prefix:
+        if matching_cert.endpoints:
+            certs_with_endpoint.append(matching_cert.name)
+
+    # If no cert has endpoint
```
hosseinsh, line 887:
```
# If no cert has endpoint
```

to 

```
# If no certificate has an endpoint, we will add one to the `certs_with_endpoint`, to allow one certificate to auto-rotate.
```
also maybe changing `certs_with_endpoint` to `certs_to_stay_on_autorotate` would help with context .

lemur/certificates/cli.py
```diff
@@ -775,3 +777,146 @@ def deactivate_entrust_certificates():
...
+    Process duplicates with same prefix as duplicate_candidate_cert
+
+    :param duplicate_candidate_cert: Name of the certificate which has duplicates
+    :param skipped_certs: List of certificates which will continue to have rotation on (no change)
+    :param rotation_disabled_certs: List of certificates for which rotation got disabled as part of this job
+    :param unique_prefix: List of uniq prefixes to avoid rework
+    :return: Success - True or False; If False, set of duplicates which were not processed
+    """
+    name_without_serial_num = duplicate_candidate_cert.name[:duplicate_candidate_cert.name.rindex("-")]
+    if name_without_serial_num in unique_prefix:
```
hosseinsh, line 845: maybe `unique_prefix` --> `unique_prefix_processed` or `processed_unique_prefix`

lemur/certificates/service.py
```diff
@@ -232,6 +233,34 @@ def find_duplicates(cert):
...
+        .filter(Certificate.rotation == true())
+        .filter(not_(Certificate.replaced.any()))
+        .filter(text("name ~ '.*-[0-9]{8}-[0-9]{8}-.*'"))
+        .all()
+    )
+
+
+def get_certificates_with_same_prefix_with_rotate_on(prefix):
+    return (
+        Certificate.query.filter(Certificate.name.like(prefix))
```
hosseinsh, line 258: this won't add performance, but might help to set the context around what this method is returning:

```
        .filter(Certificate.not_after >= now)
        .filter(not_(Certificate.replaced.any()))
```

hosseinsh (commented):
great work!

lemur/certificates/cli.py
```diff
@@ -775,3 +777,146 @@ def deactivate_entrust_certificates():
...
+
+    authority_names = current_app.config.get("AUTHORITY_TO_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES")
+    if not authority_names:
+        log_data["message"] = "Skipping task: No authorities configured"
+        current_app.logger.debug(log_data)
+        return
+
+    log_data["authorities"] = authority_names
+
+    authority_ids = [a.id for a in Authority.query.filter(Authority.name.in_(authority_names)).all()]
```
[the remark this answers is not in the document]
charhate: I'll add new method to `authorities/service.py` which takes multiple names - want to avoid separate DB call for each authority. I did consider using IDs in `AUTHORITY_TO_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES`, but it gets less readable. We do not display authority ID on UI too, so needs a DB query to do the mapping.

lemur/certificates/cli.py
```diff
@@ -775,3 +777,146 @@ def deactivate_entrust_certificates():
...
+    log_data["certs_with_serial_number_count"] = len(duplicate_candidate_certs)
+    current_app.logger.info(log_data)
+
+    skipped_certs = []
+    rotation_disabled_certs = []
+    unique_prefix = []
+    failed_certs = []
+
+    for duplicate_candidate_cert in duplicate_candidate_certs:
+        success, duplicates = process_duplicates(duplicate_candidate_cert,
```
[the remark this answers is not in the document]
charhate: Since boolean, I've used `success` and then further check is `if not success:`
For `status`, I feel appropriate value will be `success/failure` and further check will be string comparison. I prefer boolean over that. Let me know more if I missed the thought behind this.

lemur/certificates/cli.py
charhate: Thanks for catching, I thought I had `break` here

PR #3554, opened 2021-05-03
Force source and destination delete to remove associations with certificate

charhate:
Source and destination delete is admin only operation. For situations like deleting account from AWS before doing necessary certificate cleanup from AWS and in Lemur, there does not exist a way to remove association of source or destination with a certificate. Modified the source and destination delete operation to remove such association of valid certificates (expired ones are already cleaned up via celery).
Modified the confirmation message for delete to reflect this change.
Also added missing form validation for source/destination create form.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -20,6 +20,7 @@ from lemur.authorities.models import Authority
 from lemur.certificates.models import Certificate
 from lemur.certificates.schemas import CertificateOutputSchema, CertificateInputSchema
 from lemur.common.utils import generate_private_key, truthiness
+from lemur.constants import SUCCESS_METRIC_STATUS
 from lemur.destinations.models import Destination
 from lemur.domains.models import Domain
 from lemur.endpoints import service as endpoint_service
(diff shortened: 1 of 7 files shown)


lemur/destinations/service.py
```diff
@@ -78,6 +79,14 @@ def delete(destination_id):
     """
     destination = get(destination_id)
     if destination:
+        # remove association of this source from all valid certificates
+        certificates = certificate_service.get_all_valid_certificates_with_destination(destination_id)
+        for certificate in certificates:
+            certificate_service.remove_source_and_destination_association(certificate, destination.label)
```
hosseinsh, line 85: there might value in only removing destination associations.

lemur/sources/service.py
```diff
@@ -296,6 +296,13 @@ def delete(source_id):
     """
     source = get(source_id)
     if source:
+        # remove association of this source from all valid certificates
+        certificates = certificate_service.get_all_valid_certificates_with_source(source_id)
+        for certificate in certificates:
+            certificate_service.remove_source_and_destination_association(certificate, source.label)
```
hosseinsh, line 302: similarly, there might value in only removing source associations.

hosseinsh (approved):

lemur/certificates/service.py
```diff
@@ -939,3 +940,68 @@ def get_issued_cert_count_for_authority(authority):
...
+
+def remove_destination_association(certificate, destination):
+    certificate.destinations.remove(destination)
+    database.update(certificate)
+
+    try:
+        remove_from_destination(certificate, destination)
+    except Exception as e:
+        # This cleanup is the best-effort, it will capture the exception and log
+        sentry.captureException()
```
hosseinsh, line 997: I just realized we have completed the Sentry migration. Would you mind make the following changes:
```
- from lemur.extensions import sentry
+ from sentry_sdk import capture_exception
- sentry.captureException()
+ capture_exception()
```

PR #3570, opened 2021-05-11
Validate plugin options server-side

havron:
Currently, Angular does client-side validation of plugin options with the `ng-pattern` directive, which (in most cases) calls the plugin's defined validation rule before a user can submit to the API server (these definitions are found in the plugin class' `options` attribute.)

This PR adds validation of plugin options on the API server. To do so, `PluginInputSchema`'s `get_object()` method is extended to attempt validation on all plugin option values before committing to postgres. (All other callers of IPlugin's `get_option` method perform validation now, too.)

Also, all field options submitted by the user other than the `"value"` field to the input schema are ignored. This prevents, for example, a user from attempting to overwrite the validation field itself or other fields.

Questions/notes:
* I noticed in the schema, there's [a call](https://github.com/Netflix/lemur/blob/master/lemur/schemas.py#L182) to [`validate_options()`](https://github.com/Netflix/lemur/blob/master/lemur/schemas.py#L36); I wasn't sure if I could (or should) put the rest of the validation logic there instead of the `get_object()` body?
* I've removed Angular-specific regex syntax (the `/` slashes) from all the validation rules I could find (fingers crossed I haven't missed any!). Alternatively, I could also strip the slashes right before `re.match` runs, and leave the regexes as they are?
* I modified the `IPlugin` `get_option` method directly to support validation; but there's a near identical `get_plugin_option` in [`lemur.plugins.utils`](https://github.com/Netflix/lemur/blob/master/lemur/plugins/utils.py#L13) that's called by [`validate_options()`](https://github.com/Netflix/lemur/blob/master/lemur/schemas.py#L36) - should we modify this instead?

Also happy to squash commits at the end of this PR 🙂

### DIFF
--- a/lemur/common/utils.py
+++ b/lemur/common/utils.py
@@ -302,6 +302,7 @@ def validate_conf(app, required_vars):
                 "Required variable '{var}' is not set in Lemur's conf.".format(var=var)
             )
 
+
 def check_validation(validation):
     """
     Checks that the given validation string compiles successfully.
@@ -309,7 +310,7 @@ def check_validation(validation):
     :param validation:
     :return str: The validation pattern, if compilation succeeds
     """
-    
+
     try:
         compiled = re.compile(validation)
     except re.error as e:
--- a/lemur/common/utils.py
+++ b/lemur/common/utils.py
@@ -6,12 +6,15 @@
 
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
+import base64
 import random
 import re
+import socket
+import ssl
 import string
-import pem
-import base64
 
+import OpenSSL
+import pem
 import sqlalchemy
 from cryptography import x509
 from cryptography.exceptions import InvalidSignature, UnsupportedAlgorithm
@@ -396,3 +399,37 @@ def convert_pkcs7_bytes_to_pem(certs_pkcs7):
         certificates_pem.append(pem.parse(cert.public_bytes(encoding=Encoding.PEM))[0])
 
     return certificates_pem
+
+
+def get_certificate_via_tls(host, port, timeout=10):
+    """
+    Makes a TLS network connection to retrieve the current certificate for the specified host and port.
+
+    Note that if the host is valid but the port is not, we'll wait for the timeout for the connection to fail,
+    so this should remain low when doing bulk operations.
+
+    :param host: Host to get certificate for
+    :param port: Port to get certificate for
+    :param timeout: Timeout in seconds
+    """
+    context = ssl.create_default_context()
+    context.check_hostname = False  # we don't care about validating the cert
+    context.verify_mode = ssl.CERT_NONE  # we don't care about validating the cert; it may be self-signed
+    conn = socket.create_connection((host, port), timeout=timeout)
+    sock = context.wrap_socket(conn, server_hostname=host)
+    sock.settimeout(timeout)
+    try:
+        der_cert = sock.getpeercert(True)
+    finally:
+        sock.close()
+    return ssl.DER_cert_to_PEM_cert(der_cert)
+
+
+def parse_serial(pem_certificate):
+    """
+    Parses a serial number from a PEM-encoded certificate.
+    """
+    x509_cert = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, pem_certificate)
+    x509_cert.get_notAfter()
+    parsed_certificate = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, pem_certificate)
+    return parsed_certificate.get_serial_number()
--- a/lemur/common/utils.py
+++ b/lemur/common/utils.py
@@ -299,6 +299,21 @@ def validate_conf(app, required_vars):
                 "Required variable '{var}' is not set in Lemur's conf.".format(var=var)
             )
 
+def check_validation(validation):
+    """
+    Checks that the given validation string compiles successfully.
+
+    :param validation:
+    :return str: The validation pattern, if compilation succeeds
+    """
+    
+    try:
+        compiled = re.compile(validation)
+    except re.error as e:
+        raise InvalidConfiguration(f"Validation {validation} couldn't compile. Reason: {e}")
+
+    return compiled.pattern
+
 
 # https://bitbucket.org/zzzeek/sqlalchemy/wiki/UsageRecipes/WindowedRangeQuery
 def column_windows(session, column, windowsize):
--- a/lemur/constants.py
+++ b/lemur/constants.py
@@ -4,6 +4,7 @@
     :license: Apache, see LICENSE for more details.
 """
 from enum import IntEnum
+import re
 
 SAN_NAMING_TEMPLATE = "SAN-{subject}-{issuer}-{not_before}-{not_after}"
 DEFAULT_NAMING_TEMPLATE = "{subject}-{issuer}-{not_before}-{not_after}"
@@ -38,6 +39,9 @@ CERTIFICATE_KEY_TYPES = [
     "ECCSECT571R2",
 ]
 
+# For commonly reused regexes used by plugins
+EMAIL_RE = re.compile('^([\w+-.%]+@[-\w.]+\.[A-Za-z]{2,4},?)+$')
+EMAIL_RE_HELP = "Comma delimited list of email addresses"
 
 # As per RFC 5280 section 5.3.1 (https://tools.ietf.org/html/rfc5280#section-5.3.1)
 class CRLReason(IntEnum):
--- a/lemur/constants.py
+++ b/lemur/constants.py
@@ -40,8 +40,9 @@ CERTIFICATE_KEY_TYPES = [
 ]
 
 # For commonly reused regexes used by plugins
-EMAIL_RE = re.compile('^([\w+-.%]+@[-\w.]+\.[A-Za-z]{2,4},?)+$')
+EMAIL_RE = re.compile('^([\w+-.%]+@[-\w.]+\.[A-Za-z]{2,4},?)+$')  # noqa: W605
 EMAIL_RE_HELP = "Comma delimited list of email addresses"
+
 
 # As per RFC 5280 section 5.3.1 (https://tools.ietf.org/html/rfc5280#section-5.3.1)
 class CRLReason(IntEnum):
--- a/lemur/plugins/base/v1.py
+++ b/lemur/plugins/base/v1.py
@@ -113,9 +113,13 @@ class IPlugin(local):
 
     def get_option(self, name, options):
         user_opt = self.get_user_option(name, options)
+        if user_opt is None:
+            return None
+
         value = user_opt.get("value", user_opt.get("default"))
         if value is None:
             return None
+
         return self.validate_option_value(name, value)
 
     def validate_option_value(self, option_name, value):
@@ -128,7 +132,7 @@ class IPlugin(local):
             validation = class_opt.get("validation")
             if not validation:
                 return value
-            if not re.match(validation, value): # NOTE: please review that plugin validation regexes fit re
+            if not re.match(validation, value):
                 raise ValueError(f"Option '{option_name}' cannot be validated")
         elif opt_type == "select":
             available = class_opt.get("available")
--- a/lemur/plugins/base/v1.py
+++ b/lemur/plugins/base/v1.py
@@ -6,7 +6,9 @@
 
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
+from flask import current_app
 from threading import local
+import re
 
 
 # stolen from https://github.com/getsentry/sentry/
@@ -109,11 +111,45 @@ class IPlugin(local):
         """
         return self.resource_links
 
-    @staticmethod
-    def get_option(name, options):
+    def get_option(self, name, options):
+        user_opt = self.get_user_option(name, options)
+        value = user_opt.get("value", user_opt.get("default"))
+        if value is None:
+            return None
+        return self.validate_option_value(name, value)
+
+    def validate_option_value(self, option_name, value):
+        class_opt = self.get_server_options(option_name)
+        if not class_opt:
+            current_app.logger.warning("Plugin option server-defaults not found")
+            return None
+        opt_type = class_opt["type"]
+        if opt_type == "str":
+            validation = class_opt.get("validation")
+            if not validation:
+                return value
+            if not re.match(validation, value): # NOTE: please review that plugin validation regexes fit re
+                raise ValueError(f"Option '{option_name}' cannot be validated")
+        elif opt_type == "select":
+            available = class_opt.get("available")
+            if available is None:
+                return value
+            if value not in available:
+                raise ValueError(f"Option '{option_name}' doesn't match available options")
+        return value
+
+    def get_server_options(self, name):
+        class_options = getattr(self, "options", [])
+        for o in class_options:
+            if o.get("name") == name:
+                return o
+        return None
+
+    def get_user_option(self, name, options):
         for o in options:
             if o.get("name") == name:
-                return o.get("value", o.get("default"))
+                return o
+        return None
 
 
 class Plugin(IPlugin):
--- a/lemur/plugins/lemur_slack/tests/test_slack.py
+++ b/lemur/plugins/lemur_slack/tests/test_slack.py
@@ -36,7 +36,7 @@ def get_options():
     return [
         {"name": "interval", "value": 10},
         {"name": "unit", "value": "days"},
-        {"name": "webhook", "value": "https://slack.com/api/api.test"},
+        {"name": "webhook", "value": "https://hooks.slack.com/services/api.test"},
     ]
 
 
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -318,7 +318,7 @@ These notifications can be configured to use all available notification plugins.
 
 Supported types:
 
-* Certificate expiration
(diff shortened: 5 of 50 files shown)


lemur/plugins/base/v1.py
```diff
@@ -109,11 +111,45 @@ def get_resource_links(self):
...
+        elif opt_type == "select":
+            available = class_opt.get("available")
+            if available is None:
+                return value
+            if value not in available:
+                raise ValueError(f"Option '{option_name}' doesn't match available options")
+        return value
+
+    def get_server_options(self, name):
+        class_options = getattr(self, "options", [])
```
havron, line 142: Hm, I noticed I'm [failing Travis](https://travis-ci.com/github/Netflix/lemur/jobs/504920621) due to removing the **@staticmethod** decorator from `get_option` (looks like the AWS plugin uses it statically, most others call with their instance).

The only use of `self` we really need is on this line, to retrieve the server-defined options. Is there a better way to grab those options for validation? If not, maybe we could change the static calls to `get_option` to the util version of it (which appears to do the same thing): https://github.com/Netflix/lemur/blob/master/lemur/plugins/utils.py#L13

havron: (trying the latter in https://github.com/Netflix/lemur/pull/3570/commits/d70fac722e6485e66591a37f27c96cec0153f70b)

lemur/plugins/lemur_slack/tests/test_slack.py
```diff
@@ -36,7 +36,7 @@ def get_options():
     return [
         {"name": "interval", "value": 10},
         {"name": "unit", "value": "days"},
-        {"name": "webhook", "value": "https://slack.com/api/api.test"},
+        {"name": "webhook", "value": "https://hooks.slack.com/services/api.test"},
```
havron, line 39: The fact that this needed to change to a valid value to pass the test is a good sign 🙂

jtschladen:
Thanks for PR @havron ! I just ran a copy of this branch locally and tried to open and save a notification (with no changes) I get a 500 with a log like this (portions omitted for brevity):
```
Exception on /api/1/notifications/303 [PUT]
  File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  # omitted
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/auth/service.py", line 139, in decorated_function
    return f(*args, **kwargs)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/common/schema.py", line 150, in decorated_function
    data, errors = input_schema.load(request_data)
  File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/marshmallow/schema.py", line 588, in load
    result, errors = self._do_load(data, many, partial=partial, postprocess=True)
  # omitted
  File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/marshmallow/schema.py", line 964, in _invoke_processors
    data = utils.if_none(processor(data), data)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/schemas.py", line 222, in get_object
    data["plugin_object"].validate_option_value(option_name, option_value)
  File "/Users/jschladen/repos/lemur-dev/lemur/lemur/plugins/base/v1.py", line 135, in validate_option_value
    if not re.match(validation, value):
  File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/re.py", line 175, in match
    return _compile(pattern, flags).match(string)
  # omitted
  File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/sre_parse.py", line 420, in _parse_sub
    not nested and not items))
  File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/sre_parse.py", line 569, in _parse
    raise source.error(msg, len(this) + 1 + len(that))
re.error: bad character range \w-. at position 13
```

It looks like this is checking against the [recipients validation in the email plugin](https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_email/plugin.py#L94), and it looks like that regex may not be what was intended (the `-` is being treated as a range rather than a literal). I haven't checked all the other validations but we may need to go through and confirm all of them to see if they have similar issues, since this change would block data that _should_ be valid due to the new enforcement.

havron:
> Thanks for PR @havron ! I just ran a copy of this branch locally and tried to open and save a notification (with no changes) I get a 500 with a log like this (portions omitted for brevity):
> 
> ```
> Exception on /api/1/notifications/303 [PUT]
> Traceback (most recent call last):
>   File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
>     rv = self.dispatch_request()
>   # omitted
>   File "/Users/jschladen/repos/lemur-dev/lemur/lemur/auth/service.py", line 139, in decorated_function
>     return f(*args, **kwargs)
>   File "/Users/jschladen/repos/lemur-dev/lemur/lemur/common/schema.py", line 150, in decorated_function
>     data, errors = input_schema.load(request_data)
>   File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/marshmallow/schema.py", line 588, in load
>     result, errors = self._do_load(data, many, partial=partial, postprocess=True)
>   # omitted
>   File "/Users/jschladen/repos/lemur-dev/lib/python3.7/site-packages/marshmallow/schema.py", line 964, in _invoke_processors
>     data = utils.if_none(processor(data), data)
>   File "/Users/jschladen/repos/lemur-dev/lemur/lemur/schemas.py", line 222, in get_object
>     data["plugin_object"].validate_option_value(option_name, option_value)
>   File "/Users/jschladen/repos/lemur-dev/lemur/lemur/plugins/base/v1.py", line 135, in validate_option_value
>     if not re.match(validation, value):
>   File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/re.py", line 175, in match
>     return _compile(pattern, flags).match(string)
>   # omitted
>   File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/sre_parse.py", line 420, in _parse_sub
>     not nested and not items))
>   File "/Users/jschladen/.pyenv/versions/3.7.8/lib/python3.7/sre_parse.py", line 569, in _parse
>     raise source.error(msg, len(this) + 1 + len(that))
> re.error: bad character range \w-. at position 13
> ```
> 
> It looks like this is checking against the [recipients validation in the email plugin](https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_email/plugin.py#L94), and it looks like that regex may not be what was intended (the `-` is being treated as a range rather than a literal). I haven't checked all the other validations but we may need to go through and confirm all of them to see if they have similar issues, since this change would block data that _should_ be valid due to the new enforcement.

Thanks @jtschladen! Ugh yeah, an audit might be needed--if you'd like, I could grab all the validations I can find with `grep` and attempt to see if they all compile with `re`? The main issue I'd seen is the `/`s.

For email validation specifically, I forgot that I actually wrote a fixed regex for it in `lemur/constants.py`, let me add that for all email checks 🙂

lemur/constants.py
```diff
@@ -38,6 +39,9 @@
     "ECCSECT571R2",
 ]
 
+# For commonly reused regexes used by plugins
+EMAIL_RE = re.compile('^([\w+-.%]+@[-\w.]+\.[A-Za-z]{2,4},?)+$')
```
havron, line 43: @jtschladen this regex compiles now! I swapped the literal hyphen from `[\w-.]` to `[-\w.]` since (I believe) the position is interchangeable in character ranges (ran a couple of sanity checks locally).

jtschladen: Nice, sounds good to me.

jtschladen:
> Thanks @jtschladen! Ugh yeah, an audit might be needed--if you'd like, I could grab all the validations I can find with `grep` and attempt to see if they all compile with `re`? The main issue I'd seen is the `/`s.
> 
> For email validation specifically, I forgot that I actually wrote a fixed regex for it in `lemur/constants.py`, let me add that for all email checks 🙂

I just looked at another validation as an example - the one on account number in AWSSourcePlugin. At a glance, I think this was actually never applied, because we don't actually apply validations to source plugin options on the client side. That particular one looks well-formed to me, but it makes me think we probably do need to go through one by one and test/eyeball them to make sure nothing will break. 

Here's everything I found that seems relevant (all `.py` files, filtered out `views.py` since those are just comments). This is 53, which matches what I see searching in IntelliJ.

```
find . -name "*.py" | grep -v views.py | xargs grep -i '"validation": '
./lemur/plugins/lemur_jks/plugin.py:            "validation": "",
./lemur/plugins/lemur_jks/plugin.py:            "validation": "",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^https?://[a-zA-Z0-9.:-]+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^([a-zA-Z0-9/._-]+/?)+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": r"^\S+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^([a-zA-Z0-9._-]+/?)+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "[0-9a-zA-Z.:_-]+",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^https?://[a-zA-Z0-9.:-]+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^([a-zA-Z0-9/._-]+/?)+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": r"^\S+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^(([a-zA-Z0-9._-]+|{(CN|OU|O|L|S|C)})+/?)+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": "^([0-9a-zA-Z.:_-]+|{(CN|OU|O|L|S|C)})+$",
./lemur/plugins/lemur_vault_dest/plugin.py:            "validation": ".*",
./lemur/plugins/lemur_openssl/plugin.py:            "validation": "",
./lemur/plugins/bases/notification.py:            "validation": r"^\d+$",
./lemur/plugins/bases/notification.py:            "validation": "",
./lemur/plugins/lemur_email/plugin.py:            "validation": r"^([\w+-.%]+@[\w-.]+\.[A-Za-z]{2,4},?)+$",
./lemur/plugins/lemur_acme/plugin.py:            "validation": r"/^http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+$/",
./lemur/plugins/lemur_acme/plugin.py:            "validation": r"/^?([-a-zA-Z0-9.`?{}]+@\w+\.\w+)$/",
./lemur/plugins/lemur_acme/plugin.py:            "validation": "^-----BEGIN CERTIFICATE-----",
./lemur/plugins/lemur_acme/plugin.py:            "validation": r"/^http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+$/",
./lemur/plugins/lemur_acme/plugin.py:            "validation": r"/^?([-a-zA-Z0-9.`?{}]+@\w+\.\w+)$/",
./lemur/plugins/lemur_acme/plugin.py:            "validation": "^-----BEGIN CERTIFICATE-----",
./lemur/plugins/lemur_slack/plugin.py:            "validation": r"^https:\/\/hooks\.slack\.com\/services\/.+$",
./lemur/plugins/lemur_slack/plugin.py:            "validation": "^.+$",
./lemur/plugins/lemur_slack/plugin.py:            "validation": "^(@|#).+$",
./lemur/plugins/lemur_azure_dest/plugin.py:            "validation": "^https?://[a-zA-Z0-9.:-]+$",
./lemur/plugins/lemur_azure_dest/plugin.py:            "validation": "^([a-zA-Z0-9/-/?)+$",
./lemur/plugins/lemur_azure_dest/plugin.py:            "validation": "^([a-zA-Z0-9/-/?)+$",
./lemur/plugins/lemur_azure_dest/plugin.py:            "validation": "[0-9a-zA-Z.:_-~]+",
./lemur/plugins/lemur_aws/tests/test_plugin.py:            "validation": r"[0-9a-z.-]{3,63}",
./lemur/plugins/lemur_aws/tests/test_plugin.py:            "validation": r"[0-9]{12}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "^[0-9]{12,12}$",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "[0-9]{12}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "[0-9a-z.-]{3,63}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "[0-9]{12}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "[0-9]{12}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "[0-9a-z\\-]{1,25}",
./lemur/plugins/lemur_aws/plugin.py:            "validation": "^[a-zA-Z0-9_\\-]{1,256}$",
./lemur/plugins/lemur_sftp/plugin.py:            "validation": r"^(6553[0-5]|655[0-2][0-9]\d|65[0-4](\d){2}|6[0-4](\d){3}|[1-5](\d){4}|[1-9](\d){0,3})",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "(?:[a-z0-9.-]|\\{common_name\\})+",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "https?://[a-zA-Z0-9.-]+(?::[0-9]+)?",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "[0-9a-zA-Z-_.]+",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "(/[^/]+)+",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "-----BEGIN CERTIFICATE-----[a-zA-Z0-9/+\\s\\r\\n]+-----END CERTIFICATE-----",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "(/[^/]+)+",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "[a-z0-9]([-a-z0-9]*[a-z0-9])?",
./lemur/plugins/lemur_kubernetes/plugin.py:            "validation": "(/[^/]+)+",
./lemur/acme_providers/cli.py:            "validation": r"[0-9a-z.-]{3,63}",
./lemur/acme_providers/cli.py:            "validation": r"[0-9]{12}",
./lemur/notifications/service.py:            "validation": "",
./lemur/notifications/service.py:            "validation": r"^([\w+-.%]+@[\w-.]+\.[A-Za-z]{2,4},?)+$",
./lemur/notifications/service.py:                    "validation": r"^\d+$",
```

havron:
Thanks, I can take a stab at `re.compile`ing them all--I'll post the output here!

havron:
Awesome, I ran `find . -name "*.py" | grep -v views.py | xargs grep -ni '"validation": ' > validations.txt` (your command but I added `n` to grep at the end)

Then, I ran [a script](https://gist.github.com/havron/ed1026df7bbc84a00b5f3db61d241e44) to check remaining culprits:

```python                                                                          
[['./plugins/lemur_acme/plugin.py:311:            ', 'r"/^?([-a-zA-Z0-9.`?{}]+@\\w+\\.\\w+)$/"']
['./plugins/lemur_azure_dest/plugin.py:109:            ', '"^([a-zA-Z0-9/-/?)+$"']
['./plugins/lemur_azure_dest/plugin.py:116:            ', '"^([a-zA-Z0-9/-/?)+$"']
```

First case was just another email validation I missed (replaced with `EMAIL_RE`; for the other two, looks like they were both missing `]` at the ends and never should have worked to begin with. Committing changes!

jtschladen (approved):
Thanks @havron! This all looks good to me. I'll plan on merging tomorrow, just in case there's any other feedback.

havron:
Thanks @jtschladen - the only other thing I can think of, is whether or not we'd want to have a `re.compile` check written in a utils function, that is imported and called on all validations (returns the `.pattern` if successful) - that might encourage new validations to follow the style and test themselves before a PR.

jtschladen:
@havron that suggestion makes sense to me - it would certainly be ideal to catch any errors as soon as possible. I'll hold off on merging if you want to add that first?

havron:
Sure! I can add that tomorrow morning, if that's okay - I do want to get this one out 🙂

lemur/common/utils.py
```diff
@@ -299,6 +299,21 @@ def validate_conf(app, required_vars):
...
+    :param validation:
+    :return str: The validation pattern, if compilation succeeds
+    """
+    
+    try:
+        compiled = re.compile(validation)
+    except re.error as e:
+        raise InvalidConfiguration(f"Validation {validation} couldn't compile. Reason: {e}")
+
+    return compiled.pattern
```
havron, line 318: Pushed latest commit! I will note that the `.pattern` returned by the check does escape certain characters - I'm not sure if that would affect Angular? We could return the raw `validation` instead of `compiled.pattern` if need be. Cheers! 🎉

jtschladen: I checked a few of them in the UI with your changes and they look fine to me - SNS topic, Slack webook, and Vault URL/file all look likely to break if anything too weird happened in terms of escaping (but they're fine as far as I see). If there's any others you think are likely to break I can try them, but otherwise I think it seems to be fine.

jtschladen:
Merged @havron - thanks again for this one! Appreciate the diligence getting it to completion.

PR #3571, opened 2021-05-12
Fix login bug during user creation

havron:
When an SSO login is done for the first time (i.e., when there is no `user` row in the database for the new user), roles are created properly with the role service, but the newly created user is never returned by `update_user()` in `lemur.auth.views`. This dooms the login until the second try, since the `user` variable is initially provided by `retrieve_user_*` and is used to complete the auth check.

### DIFF
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -275,6 +275,8 @@ def update_user(user, profile, roles):
             roles,
         )
 
+    return user
+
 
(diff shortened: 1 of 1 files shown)


hosseinsh (approved):

jtschladen (approved):

hosseinsh:
Thanks for the PR @havron!

PR #3573, opened 2021-05-14
generate and verify oauth state tokens

havron:
This PR adds support for generation and verification of oauth state token nonces during login. Rather than rely on storing state tokens in a database or flask session variable (good luck trying that with gunicorn threading lol), we instead go stateless 🎉  by leveraging hmac to sign all the tokens we hand out, and then verify every token that's passed to the Lemur API server. Timestamp is included in the state token, so we could do some kind of freshness check if we want.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -873,6 +873,36 @@ For more information about how to use social logins, see: `Satellizer <https://g
 
             OAUTH2_VERIFY_CERT = True
 
+.. data:: OAUTH_STATE_TOKEN_SECRET
+    :noindex:
+
+        The OAUTH_STATE_TOKEN_SECRET is used to sign state tokens to guard against CSRF attacks. Without a secret configured, Lemur will use
+        a fallback secret: ``b'0auth'``. The secret must be `bytes-like <https://cryptography.io/en/latest/glossary/#term-bytes-like>`;
+        it will be used to instantiate `HMAC <https://cryptography.io/en/latest/hazmat/primitives/mac/hmac/#cryptography.hazmat.primitives.hmac.HMAC>`.
+
+        For implementation details, see ``generate_state_token()`` and ``verify_state_token() in ``lemur/auth/views.py``.
+
+        Running lemur create_config will securely generate a key for your configuration file.
+        If you would like to generate your own, we recommend the following method:
+
+            >>> import os
+            >>> import base64
+            >>> KEY_LENGTH = 32  # tweak as needed
+            >>> base64.b64encode(os.urandom(KEY_LENGTH))
+
+    ::
+
+        OAUTH_STATE_TOKEN_SECRET = b'RVdUVEZFQU4oISFAK18oJGZka2lmZmFsMjI0NTkxMjg='
+
+.. data:: OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS
+    :noindex:
+
+        Defaults to 15 seconds if configuration is not discovered.
+
+        ::
+
+            OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS = 15
+
 .. data:: GOOGLE_CLIENT_ID
     :noindex:
 
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -931,9 +931,9 @@ For more information about how to use social logins, see: `Satellizer <https://g
 
         The OAUTH_STATE_TOKEN_SECRET is used to sign state tokens to guard against CSRF attacks. Without a secret configured, Lemur will use
         a fallback secret: ``b'0auth'``. The secret must be `bytes-like <https://cryptography.io/en/latest/glossary/#term-bytes-like>`;
-        it will be used to instantiate `HMAC <https://cryptography.io/en/latest/hazmat/primitives/mac/hmac/#cryptography.hazmat.primitives.hmac.HMAC>`.
+        it will be used to instantiate the key parameter of `HMAC <https://cryptography.io/en/latest/hazmat/primitives/mac/hmac/#cryptography.hazmat.primitives.hmac.HMAC>`.
 
-        For implementation details, see ``generate_state_token()`` and ``verify_state_token() in ``lemur/auth/views.py``.
+        For implementation details, see ``generate_state_token()`` and ``verify_state_token()`` in ``lemur/auth/views.py``.
 
         Running lemur create_config will securely generate a key for your configuration file.
         If you would like to generate your own, we recommend the following method:
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -945,7 +945,7 @@ For more information about how to use social logins, see: `Satellizer <https://g
 
     ::
 
-        OAUTH_STATE_TOKEN_SECRET = base64.b64encode(lemur.common.utils.get_random_secret(32).encode('utf8'))
+        OAUTH_STATE_TOKEN_SECRET = lemur.common.utils.get_state_token_secret()
 
 .. data:: OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS
     :noindex:
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -929,8 +929,8 @@ For more information about how to use social logins, see: `Satellizer <https://g
 .. data:: OAUTH_STATE_TOKEN_SECRET
     :noindex:
 
-        The OAUTH_STATE_TOKEN_SECRET is used to sign state tokens to guard against CSRF attacks. Without a secret configured, Lemur will use
-        a fallback secret: ``b'0auth'``. The secret must be `bytes-like <https://cryptography.io/en/latest/glossary/#term-bytes-like>`;
+        The OAUTH_STATE_TOKEN_SECRET is used to sign state tokens to guard against CSRF attacks. Without a secret configured, Lemur will create
+        a fallback secret on a per-server basis that would last for the length of the server's lifetime (e.g., between redeploys). The secret must be `bytes-like <https://cryptography.io/en/latest/glossary/#term-bytes-like>`;
         it will be used to instantiate the key parameter of `HMAC <https://cryptography.io/en/latest/hazmat/primitives/mac/hmac/#cryptography.hazmat.primitives.hmac.HMAC>`.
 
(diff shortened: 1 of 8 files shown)


hosseinsh (commented):
This is a great PR @havron protecting us against CSRF attacks.
would you be open about adding some test cases for `verify_state_token` and `verify_state_token`? This would give us a higher confidence around the correctness, and also potential future errors would get caught early.


Also would be great to add this new config to the docs, so others could also benefit from it
https://github.com/Netflix/lemur/blob/d4e742ebf6ec66d2056d1d2da4b7b830570a6f6c/docs/administration.rst

lemur/auth/views.py
```diff
@@ -278,6 +282,36 @@ def update_user(user, profile, roles):
...
+    try:
+        state = token.encode('utf-8')
+        ts, digest = state.split(b':')
+        timestamp = int(ts, 16)  # todo: check freshness of timestamp vs. current time?
+        digest = base64.b64decode(digest)
+        h = build_hmac()
+        h.update(ts)
+        h.verify(digest)  # should verification fail for any reason, an exception is thrown
+        return True
+    except Exception as e:
```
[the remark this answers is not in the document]
havron: Good point! I just added a new explicit catch for it. I noticed that there's several ways to trigger exceptions on badly formatted inputs (e.g., ValueError, binascii decoding error). Perhaps we can leave the generic `Exception` at the end to catch these?

lemur/auth/views.py
```diff
@@ -278,6 +282,36 @@ def update_user(user, profile, roles):
...
+    digest = base64.b64encode(h.finalize())
+    state = ts + b':' + digest
+    return state.decode('utf-8')
+
+
+def verify_state_token(token):
+    try:
+        state = token.encode('utf-8')
+        ts, digest = state.split(b':')
+        timestamp = int(ts, 16)  # todo: check freshness of timestamp vs. current time?
```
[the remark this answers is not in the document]
havron: Done!

havron:
> This is a great PR @havron protecting us against CSRF attacks.
> would you be open about adding some test cases for `verify_state_token` and `verify_state_token`? This would give us a higher confidence around the correctness, and also potential future errors would get caught early.
> 
> Also would be great to add this new config to the docs, so others could also benefit from it
> https://github.com/Netflix/lemur/blob/d4e742ebf6ec66d2056d1d2da4b7b830570a6f6c/docs/administration.rst

@hosseinsh thank you! Sure thing - I just added `lemur/tests/test_oauth.py` and new docs updates 🙂

lemur/tests/test_oauth.py
```diff
@@ -0,0 +1,22 @@
...
+    assert generate_state_token()
+
+
+def test_verify_state_token(client):
+    from lemur.auth.views import generate_state_token
+    from lemur.auth.views import verify_state_token
+
+    token = generate_state_token()
+    assert verify_state_token(token)
+    assert not verify_state_token('123456:f4k8')
```
havron, line 22: We could add a test for staleness:

```python
token = generate_state_token()
assert verify_state_token(token)
time.sleep(15)
assert not verify_state_token(token)
```

But that would slow down every build by 15 seconds - is it worth including? (I don't think it's possible to patch a local variable in a function with `mock`)

lemur/auth/views.py
```diff
@@ -278,6 +283,43 @@ def update_user(user, profile, roles):
...
+    digest = base64.b64encode(h.finalize())
+    state = ts + b':' + digest
+    return state.decode('utf-8')
+
+
+def verify_state_token(token):
+    stale_seconds = current_app.config.get('OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS', 15)
+    try:
+        state = token.encode('utf-8')
+        ts, digest = state.split(b':')
```
hosseinsh, line 305: to render this more robust, maybe it would be helpful to catch this error, by first checking the correctness of the format?

```
ValueError: not enough values to unpack (expected 2, got 1)
```

docs/administration.rst
```diff
@@ -926,6 +926,36 @@ For more information about how to use social logins, see: `Satellizer <https://g
...
+        If you would like to generate your own, we recommend the following method:
+
+            >>> import os
+            >>> import base64
+            >>> KEY_LENGTH = 32  # tweak as needed
+            >>> base64.b64encode(os.urandom(KEY_LENGTH))
+
+    ::
+
+        OAUTH_STATE_TOKEN_SECRET = b'RVdUVEZFQU4oISFAK18oJGZka2lmZmFsMjI0NTkxMjg='
```
hosseinsh, line 948: shall we just show the case, where the secret is created on the fly?
```
OAUTH_STATE_TOKEN_SECRET  = base64.b64encode(get_random_secret(32).encode('utf8'))
```

I am afraid (a) this will get flagged by security researchers as secret in code, and more importantly (b) someone just copy pasting this secret.

lemur/tests/test_oauth.py
jtschladen: Are you able to use `freeze_time`? Freeze in past, generate token, unfreeze, verify stale? 
Sample: https://github.com/Netflix/lemur/blob/master/lemur/tests/test_certificates.py#L203
I haven't used it extensively so perhaps not an option, but I wanted to check if you tried that already.

lemur/constants.py
```diff
@@ -12,6 +12,9 @@
 SUCCESS_METRIC_STATUS = "success"
 FAILURE_METRIC_STATUS = "failure"
 
+# fallback if we don't have an oauth state token secret configured
+OAUTH_STATE_TOKEN_SECRET_FALLBACK = b'0auth'
```
hosseinsh, line 16: I am wondering, if we could just create the secret on the fly for the fall_back secret?
```
OAUTH_STATE_TOKEN_SECRET_FALLBACK = base64.b64encode(get_random_secret(32).encode('utf8'))
```

likely between deploys the state cannot be verified correctly, and if there are multiple instances of Lemur, each will have varying keys, which shouldn't be a problem, since the request will be coming back to the same instance. 


any thoughts @havron

lemur/tests/test_oauth.py
hosseinsh: what about setting the tolerance to 0, would do that the trick?
`OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS = 0`

hosseinsh, line 22: shall we also add a test case for miss formatted token? (after adding the robustness in the code for it)

```
assert not verify_state_token('123456f4k8')
assert not verify_state_token('123456::f4k8')
assert not verify_state_token('')
```

docs/administration.rst
```diff
@@ -926,6 +926,36 @@ For more information about how to use social logins, see: `Satellizer <https://g
 
             OAUTH2_VERIFY_CERT = True
 
+.. data:: OAUTH_STATE_TOKEN_SECRET
+    :noindex:
+
+        The OAUTH_STATE_TOKEN_SECRET is used to sign state tokens to guard against CSRF attacks. Without a secret configured, Lemur will use
+        a fallback secret: ``b'0auth'``. The secret must be `bytes-like <https://cryptography.io/en/latest/glossary/#term-bytes-like>`;
```
hosseinsh, line 933: I was wondering if we can create a random fallback secret? 
Also left comment, where the fall back secret is set.

hosseinsh (commented):
Thanks @havron for the revision. I left few minor comments on ideas how to improve this functionality even further.  🚀

docs/administration.rst
havron: Makes sense to me! Updating it.

lemur/constants.py
havron: Makes sense to me - perhaps it should move to `utils`, then, since it's no longer a constant. Updating!

lemur/tests/test_oauth.py
havron: Thank you both! I used a little bit from both suggestions: I define `OAUTH_STATE_TOKEN_STALE_TOLERANCE_SECONDS` in `lemur.tests.conf` and import it in my test, then apply `freeze_time` to grab a token that's at least the tolerance seconds stale. By pulling this from test config explicitly, we can be confident the the verification will also pull from our test tolerance seconds.

lemur/auth/views.py
havron: Thanks, I split up the token parsing logic from the verification work - commit forthcoming!

lemur/constants.py
havron: Updated it, and pulled the `get_random_secret(length)` into `utils` as well.

docs/administration.rst
havron: Done! Editing docs to reflect it.

lemur/tests/test_oauth.py
havron: Awesome, adding now, thanks!

havron:
Thanks so much for feedback @hosseinsh and @jtschladen - I just updated the PR to reflect everything discussed in b96fd3a. Happy Friday!

lemur/common/utils.py
```diff
@@ -55,6 +55,18 @@ def get_psuedo_random_string():
     return challenge
 
 
+def get_random_secret(length):
```
hosseinsh, line 58: we should consider consolidating the `get_random_secrets()` and  `get_psuedo_random_string()`
for instance
```
def get_psuedo_random_string():    
"""    Create a random and strongish challenge.    """
    get_random_secret(24)
```

I have also noticed the `get_random_secret(length)` in other corners of the code:
https://github.com/Netflix/lemur/blob/176417c2400dfc60356769b26b4cee7421a91c91/docker/src/lemur.conf.py#L20-L24
https://github.com/Netflix/lemur/blob/d4e742ebf6ec66d2056d1d2da4b7b830570a6f6c/lemur/tests/conf.py#L11-L16

We can take care of it in a separate PR ;)

hosseinsh (approved):

hosseinsh:
> Thanks so much for feedback @hosseinsh and @jtschladen - I just updated the PR to reflect everything discussed: [b96fd3a](https://github.com/Netflix/lemur/commit/b96fd3a7deabe629e361d3fc56c040f2dac2fb76). Happy Friday!

Thank you for the PR @havron, and all the improvements!
Indeed Happy Friday!  🎉

will merge the PR on Monday.

PR #3580, opened 2021-05-17
Consider duplicate certs issued in last n days

charhate:
Extending https://github.com/Netflix/lemur/pull/3542 to consider certificates issued in last N days, where N is configured using config `DAYS_SINCE_ISSUANCE_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES`. If not configured, this value results in -1 and gets ignored. In that case, the job will run on all valid certificates that satisfy the duplicate check.

Not setting this parameters doesn't cause a harm and thus is optional. By setting this param value to same as the window after which deduplication job runs, one can avoid same cert from going under duplicate checks (this happens only if the cert with serial number is actively used).

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -798,6 +798,8 @@ def disable_rotation_of_duplicate_certificates(commit):
         return
 
     log_data["authorities"] = authority_names
+    days_since_issuance = current_app.config.get("DAYS_SINCE_ISSUANCE_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES", -1)
+    log_data["days_since_issuance"] = f"{days_since_issuance} (Ignored if -1)"
 
     authority_ids = []
     invalid_authorities = []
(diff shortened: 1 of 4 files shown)


lemur/certificates/cli.py
```diff
@@ -798,6 +798,8 @@ def disable_rotation_of_duplicate_certificates(commit):
         return
 
     log_data["authorities"] = authority_names
+    days_since_issuance = current_app.config.get("DAYS_SINCE_ISSUANCE_DISABLE_ROTATE_OF_DUPLICATE_CERTIFICATES", -1)
```
hosseinsh, line 801: would `none` serve better in this case, or maybe a reasonable defaults like `90` days?

hosseinsh:
would maybe adding this and the dedup task to the docs help to make this feature more visible?

lemur/certificates/cli.py
charhate: I'll change to `none` which will take care of all duplicate certificates. 90 Days is a good option too given LE certs are valid for 90 days. But `none` is more generic.

hosseinsh (approved):

PR #3581, opened 2021-05-18
Add notifications when reissue fails or a cert is reissued with no endpoints

jtschladen:
Adds two new notification types:
* When a certificate fails to be re-issued
* When a certificate is successfully re-issued, but has no associated endpoints

Both types are (currently) sent exclusively to the certificate owner and security team.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -38,7 +38,8 @@ from lemur.deployment import service as deployment_service
 from lemur.domains.models import Domain
 from lemur.endpoints import service as endpoint_service
 from lemur.extensions import metrics
-from lemur.notifications.messaging import send_rotation_notification
+from lemur.notifications.messaging import send_rotation_notification, send_reissue_no_endpoints_notification, \
+    send_reissue_failed_notification
 from lemur.plugins.base import plugins
 
 manager = Manager(usage="Handles all certificate related tasks.")
@@ -139,14 +140,16 @@ def request_rotation(endpoint, certificate, message, commit):
                                                                  "endpoint": str(endpoint.dnsname)})
 
 
-def request_reissue(certificate, commit):
+def request_reissue(certificate, notify, commit):
     """
     Reissuing certificate and handles any exceptions.
     :param certificate:
+    :param notify:
     :param commit:
     :return:
     """
     status = FAILURE_METRIC_STATUS
+    notify = certificate.notify
     try:
         print("[+] {0} is eligible for re-issuance".format(certificate.name))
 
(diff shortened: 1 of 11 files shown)


lemur/plugins/lemur_email/templates/rotation.html
```diff
@@ -98,7 +98,7 @@
                                             </tr>
```
jtschladen, line 98: I "fixed" this file a while ago, meaning I got it to successfully render. Looking at it again today, the contents were not correct. This new version should be closer to what we actually want out of it.

jtschladen: Screen Shot 2021-05-17 at 6 47 20 PM
Screen Shot 2021-05-17 at 6 47 14 PM

lemur/plugins/lemur_email/templates/reissued_with_no_endpoints.html
```diff
@@ -0,0 +1,195 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
```
jtschladen, line 1: Screen Shot 2021-05-17 at 6 47 03 PM

lemur/certificates/cli.py
```diff
@@ -139,14 +140,16 @@ def request_rotation(endpoint, certificate, message, commit):
...
+def request_reissue(certificate, notify, commit):
     """
     Reissuing certificate and handles any exceptions.
     :param certificate:
+    :param notify:
     :param commit:
     :return:
     """
     status = FAILURE_METRIC_STATUS
+    notify = certificate.notify
```
hosseinsh, line 152: I am trying to understand what is the purpose if `notify`, if it is overwritten by `certificate.notify`.

Are we thinking more like
```
notify = certificate.notify and notify
```
basically, as long as both certificate notification is on, and the cli call is requesting notify, we send out additional notifications? 

or do we want it to be just controlled via the `notify` input parameter?

hosseinsh:
Thanks for the screenshots of the messages, that makes reviewing them much more easier. 
Also great work, can't wait to see the new type of messaging in production.

lemur/certificates/cli.py
jtschladen: You're right, this is a mistake. I originally didn't have it configurable (they were always enabled, assuming `certificate.notify` was enabled) but I later added the config option and passed it via the same variable name (oops!). So, my intention was `notify = certificate.notify and notify` - but I'm open to suggestions if you disagree.

hosseinsh (approved):

lemur/certificates/cli.py
jtschladen: I think it might be a good case to _ignore_ `certificate.notify` since honoring it risks unnoticed failure. That said, _not_ honoring it might confuse people. Overall, I'm thinking that sending regardless of `certificate.notify` is probably the safest option to avoid failures.

hosseinsh: I also prefer this option, since it gives the flexibility to control this type of notification via the UI.

PR #3582, opened 2021-05-18
Set default value for AcmeDnsChallenge.cleanup()'s unused validation_target parameter

tho:
The function is called from [lemur/plugins/lemur_acme/plugin.py#L229](https://github.com/Netflix/lemur/blob/ace7a1d52460e1e96f04327a59e576939636684c/lemur/plugins/lemur_acme/plugin.py#L229) with only two arguments, which results in a `TypeError` exception when the statement is executed.

Alternatively, the unused `validation_target` parameter could be removed altogether.

### DIFF
--- a/lemur/plugins/lemur_acme/challenge_types.py
+++ b/lemur/plugins/lemur_acme/challenge_types.py
@@ -249,7 +249,7 @@ class AcmeDnsChallenge(AcmeChallenge):
     def deploy(self, challenge, acme_client, validation_target):
         pass
 
-    def cleanup(self, authorizations, acme_client, validation_target):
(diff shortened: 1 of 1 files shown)


hosseinsh:
Thanks for the PR @tho.
I suppose `validation_target` originates from the base class
https://github.com/Netflix/lemur/blob/3783fbeaa1645bbee022827f4f53ffb12dd65a61/lemur/plugins/lemur_acme/challenge_types.py#L56-L63

I think we can for now go with your suggestion, avoid the `TypeError`s.

hosseinsh (approved):

PR #3583, opened 2021-05-18
Add revocation notification

jtschladen:
I was originally planning to wait to add this until more of the notification refactoring I'm planning was done, but at this point it makes sense to add this type alongside the other two re-issue notifications I just added. As with the last ones, these will only be sent to the certificate owner and security team.

### DIFF
--- a/lemur/notifications/cli.py
+++ b/lemur/notifications/cli.py
@@ -8,6 +8,7 @@
 from flask_script import Manager
 from sentry_sdk import capture_exception
 
+from lemur.certificates.service import get_expiring_deployed_certificates
 from lemur.constants import SUCCESS_METRIC_STATUS, FAILURE_METRIC_STATUS
 from lemur.extensions import metrics
 from lemur.notifications.messaging import send_expiration_notifications, \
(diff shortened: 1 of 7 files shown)


lemur/plugins/lemur_email/templates/revocation.html
```diff
@@ -0,0 +1,163 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
```
jtschladen, line 1: Screen Shot 2021-05-18 at 11 15 12 AM

hosseinsh (approved):

charhate (approved):

lemur/notifications/cli.py
```diff
@@ -111,13 +112,15 @@ def security_expiration_summary(exclude):
...
-    and notify the certificate owner.
+    Attempt to find any certificates that are expiring soon but are still deployed, and notify the certificate owner.
+    This information is retrieved from the database, and is based on the previous run of
+    identity_expiring_deployed_certificates.
     """
     status = FAILURE_METRIC_STATUS
     try:
         print("Starting to notify owners about certificates that are expiring but still deployed!")
-        success = send_expiring_deployed_certificate_notifications(exclude)
+        certificates = get_expiring_deployed_certificates(exclude).items()
```
jtschladen, line 122: I had to restructure this due to a circular dependency issue in Lemur caused by #3536:

1. In #3536, I added a dependency from `messaging.py` on `certificates/service.py`  in order to find all relevant expiring deployed certs
2. In this PR (#3583), I added a (circular) dependency from `certificates/service.py` on `messaging.py` in order to send a notification when revocation occurs

This seemed like the simplest way to solve the issue with minimal code changes.

PR #3586, opened 2021-05-21
Generate and verify hashes for all required python packages

csine-nflx:
This change adds hash verification to all pip packages brought in as dependencies.  It does this by updating the pip-compile command with the --generate-hashes flag.  All requirements files are updated to include sha256 hashes, and they are verified using the `--require-hashes` argument to `pip install`. In creating this change, the requirements.txt files received hashes for all packages, but the package versions were not updated.  The build would fail without adding hashes at this time, which is why they are included.

For local development, this change should just work as it always has.  However, if you are building Lemur using external tools/configurations, this may impact your build.  Just running `pip install -e .` will no longer pull in requirements.txt.  To fix this, update your build script by adding `pip install --require-hashes -r requirements.txt` just after the previous command.  

[Hash-Checking Mode](https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode)


hosseinsh:
Closing for now, since we couldn't see any significant benefits.

PR #3597, opened 2021-05-25
Improving robustness of Check revoke 

hosseinsh:
- adding timeouts to prevent the OCSP and curl lookups from hanging
- adding pagination for the query, to improve DB query performance
- adding metric to monitor OCSP failures (`unauthorized`)
- adding better logging about the progress of job
- adding a final report on OCSP/CRL failures

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -699,8 +699,13 @@ def check_revoked():
             # this must be tha last page
             there_are_still_certs = False
         else:
-            page = page + 1
-
+            metrics.send(
+                "certificate_revoked_progress",
+                "counter",
+                1,
+                metric_tags={"page":page}
+            )
+            page += 1
 
         for cert in certs:
             try:
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -755,7 +755,7 @@ def check_revoked():
     metrics.send(
         "certificate_revoked_checked",
         "gauge",
-        page * count + len(certs),
+        (page - 1) * count + len(certs),
     )
 
 
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -697,7 +697,7 @@ def check_revoked():
             # this must be tha last page
             there_are_still_certs = False
         else:
-            page =+ 1
+            page = page + 1
 
 
         for cert in certs:
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -687,37 +687,48 @@ def check_revoked():
         "function": f"{__name__}.{sys._getframe().f_code.co_name}",
         "message": "Checking for revoked Certificates"
     }
+    there_are_still_certs = True
+    page=1
+    count=1000
+    while there_are_still_certs:
+        certs = get_all_valid_certs(current_app.config.get("SUPPORTED_REVOCATION_AUTHORITY_PLUGINS", []),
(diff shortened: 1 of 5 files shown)


lemur/certificates/cli.py
```diff
@@ -687,37 +687,74 @@ def check_revoked():
...
+    )
+    metrics.send(
+        "certificate_revoked_crl_error",
+        "gauge",
+        crl_err_count,
+    )
+    metrics.send(
+        "certificate_revoked_checked",
+        "gauge",
+        page * count + len(certs),
```
charhate, line 756: Would this be `(page - 1) * count + len(certs)` since last page will have `len(certs)` number of certs

lemur/certificates/verify.py
```diff
@@ -55,21 +57,33 @@ def ocsp_verify(cert, cert_path, issuer_chain_path):
...
+        raise Exception(f"Got error when parsing OCSP url: {url}")
 
     elif "revoked" in p_message:
         current_app.logger.debug(
             "OCSP reports certificate revoked: {}".format(cert.serial_number)
         )
         return False
 
+    elif "unauthorized" in p_message:
+        # indicates the OCSP server does no longer no this certificate
```
charhate, line 80: if you can fix the comment

lemur/certificates/verify.py
```diff
@@ -55,21 +57,33 @@ def ocsp_verify(cert, cert_path, issuer_chain_path):
         stderr=subprocess.PIPE,
     )
 
-    message, err = p2.communicate()
+    try:
+        message, err = p2.communicate(timeout=15)
+    except TimeoutExpired:
+        p2.kill()
+        p2.communicate()
```
charhate, line 64: Is this a retry without any timeout?

lemur/certificates/verify.py
```diff
@@ -55,21 +57,33 @@ def ocsp_verify(cert, cert_path, issuer_chain_path):
...
+        return False
```
charhate, line 65: Returning False would mean cert is revoked (not valid), if so I think it is safer to retry in future run and return True in case of timeout.

lemur/certificates/verify.py
hosseinsh: it does indeed look like a try, though my understanding from the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) was that after kill, the communicate would finish communication.

> The child process is not killed if the timeout expires, so in order to cleanup properly a well-behaved application should kill the child process and finish communication


This might also be a MS artifact, from online [readings](https://github.com/python/cpython/commit/5fe153c21de020f37712910b0dcce93341bba272)
```
            if _mswindows:
                # Windows accumulates the output in a single blocking
                # read() call run on child threads, with the timeout
                # being done in a join() on those threads.  communicate()
                # _after_ kill() is required to collect that and add it
                # to the exception.
                exc.stdout, exc.stderr = process.communicate()
```


will remove it for now. 

hoping we can just rewrite this in Python, instead of using subprocesses.

lemur/certificates/verify.py
hosseinsh: excellent point, will raise exception, consistent with other errors.

lemur/certificates/cli.py
hosseinsh: good catch

charhate (approved):

PR #3599, opened 2021-05-25
put runtime oauth secret in flask session

havron:
I realized that in #3573, I inadvertently introduced a bug in the event that a state token secret must be generated at runtime: it's not stored in the Flask session, so subsequent use wouldn't work. 

In this PR, I write the runtime-made key into the Flask session explicitly, and add new tests for it. I also added an exception catch + test for `build_hmac()` in the case where the key set by the user isn't bytes-like, which is required.

### DIFF
--- a/lemur/auth/views.py
+++ b/lemur/auth/views.py
@@ -284,8 +284,18 @@ def update_user(user, profile, roles):
 
 
 def build_hmac():
-    key = current_app.config.get('OAUTH_STATE_TOKEN_SECRET', get_state_token_secret())
-    return hmac.HMAC(key, hashes.SHA256(), backend=default_backend())
+    key = current_app.config.get('OAUTH_STATE_TOKEN_SECRET', None)
+    if not key:
+        current_app.logger.warning("OAuth State Token Secret not discovered in config. Generating one.")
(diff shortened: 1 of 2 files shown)


lemur/auth/views.py
```diff
@@ -284,8 +284,18 @@ def update_user(user, profile, roles):
...
+    key = current_app.config.get('OAUTH_STATE_TOKEN_SECRET', None)
+    if not key:
+        current_app.logger.warning("OAuth State Token Secret not discovered in config. Generating one.")
+        key = get_state_token_secret()
+        current_app.config['OAUTH_STATE_TOKEN_SECRET'] = key  # store for remainder of Flask session
+
+    try:
+        h = hmac.HMAC(key, hashes.SHA256(), backend=default_backend())
+    except TypeError:
+        current_app.logger.error("OAuth State Token Secret must be bytes-like.")
```
havron, line 296: This case happens when the user doesn't set the key as bytes-like; I was considering making a new proper key in this case, but then it would override the user's choice.

hosseinsh (approved):
good catch @havron, and great new test cases. Thanks for raising the PR.

PR #3607, opened 2021-06-01
Add LEMUR_REISSUE_NOTIFICATION_EXCLUDED_DESTINATIONS

jtschladen:
Adds `LEMUR_REISSUE_NOTIFICATION_EXCLUDED_DESTINATIONS`, which allows configuration to specify a set of destinations for which we should _not_ notify about reissued certs with no endpoints. This is useful for cases where alternative (non-Lemur) automation handles certificate rotation on non-Lemur-managed endpoints.


hosseinsh (approved):
great PR. Thank you.

PR #3610, opened 2021-06-03
Autorotation for certs attached to destination(s)

havron:
This PR allows config to specify the default rotation status that Angular picks up during the certificate creation screen. I've set it as `False` here to be conservative, but having it configurable can be helpful to users who want to make cert rotation opt-out instead of opt-in!

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -176,6 +176,22 @@ def get_all_certs_attached_to_endpoint_without_autorotate():
     )
 
 
+def get_all_certs_attached_to_destination_without_autorotate():
+    """
+        Retrieves all certificates that are attached to a destination, but that do not have autorotate enabled.
+
+        :return: list of certificates attached to a destination without autorotate
+        """
+    return (
+        Certificate.query.filter(Certificate.destinations.any())
+        .filter(Certificate.rotation == false())
+        .filter(Certificate.revoked == false())
+        .filter(Certificate.not_after >= arrow.now())
+        .filter(not_(Certificate.replaced.any()))
+        .all()  # noqa
+    )
+
+
 def get_all_pending_cleaning_expiring_in_days(source, days_to_expire):
     """
     Retrieves all certificates that are available for cleaning, not attached to endpoint,
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -162,10 +162,10 @@ def get_all_pending_cleaning_expired(source):
 
 def get_all_certs_attached_to_endpoint_without_autorotate():
     """
-        Retrieves all certificates that are attached to an endpoint, but that do not have autorotate enabled.
+    Retrieves all certificates that are attached to an endpoint, but that do not have autorotate enabled.
 
-        :return: list of certificates attached to an endpoint without autorotate
-        """
+    :return: list of certificates attached to an endpoint without autorotate
+    """
     return (
         Certificate.query.filter(Certificate.endpoints.any())
         .filter(Certificate.rotation == false())
@@ -176,12 +176,24 @@ def get_all_certs_attached_to_endpoint_without_autorotate():
     )
 
 
-def get_all_certs_attached_to_destination_without_autorotate():
+def get_all_certs_attached_to_destination_without_autorotate(plugin_name=None):
     """
-        Retrieves all certificates that are attached to a destination, but that do not have autorotate enabled.
+    Retrieves all certificates that are attached to a destination, but that do not have autorotate enabled.
 
-        :return: list of certificates attached to a destination without autorotate
-        """
+    :param plugin_name: Optional destination plugin name to query. Queries certificates attached to any destination if not provided.
+    :return: list of certificates attached to a destination without autorotate
+    """
+    if plugin_name:
+        return (
+            Certificate.query.filter(Certificate.destinations.any(plugin_name=plugin_name))
+            .filter(Certificate.rotation == false())
+            .filter(Certificate.revoked == false())
+            .filter(Certificate.not_after >= arrow.now())
+            .filter(not_(Certificate.replaced.any()))
+            .all()  # noqa
+        )
+
+
     return (
         Certificate.query.filter(Certificate.destinations.any())
         .filter(Certificate.rotation == false())
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -193,7 +193,6 @@ def get_all_certs_attached_to_destination_without_autorotate(plugin_name=None):
             .all()  # noqa
         )
 
-
     return (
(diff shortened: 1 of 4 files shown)


havron:
Hm, I noticed that there are models where defaults are imported directly from lemur.common.defaults ([like here](https://github.com/Netflix/lemur/blob/master/lemur/certificates/models.py#L198)) but I wasn't sure if rotation defaults should also be included there since they're part of the defaults view, instead of the default module?

Also: in [lemur.certificates.models](https://github.com/Netflix/lemur/blob/master/lemur/certificates/models.py#L150), would it make sense to rewrite to something like:
```python
    rotation = Column(Boolean, default=current_app.config.get("LEMUR_ROTATION_DEFAULT", False))
```

I didn't see that pattern in other places, so I figured I should ask about it before committing.

lemur/manage.py
```diff
@@ -112,6 +112,7 @@
 LEMUR_DEFAULT_LOCATION = ''
 LEMUR_DEFAULT_ORGANIZATION = ''
 LEMUR_DEFAULT_ORGANIZATIONAL_UNIT = ''
+LEMUR_DEFAULT_ROTATION = False
```
hosseinsh, line 115: I like that we are not requiring this new variable, and just default to `False`

https://github.com/Netflix/lemur/blob/3fbcbc27174622b6075b492eccc1a42530afc5f0/lemur/manage.py#L61

docs/administration.rst
```diff
@@ -304,6 +304,14 @@ and are used when Lemur creates the CSR for your certificates.
         LEMUR_DEFAULT_AUTHORITY = "verisign"
 
 
+.. data:: LEMUR_DEFAULT_ROTATION
+    :noindex:
+
+    ::
+
```
hosseinsh, line 311: shall we add a note, that this defaults to False, if not defined?

hosseinsh (commented):
Thanks for your continuous contributions @havron to Lemur.

Adding the flexibility to define the default rotation value would potentially be of value to certain use-cases.

We have noticed that not all certificates can be rotated by Lemur, e.g., if the endpoint is not visible to Lemur, which would require manual rotation by the respective team. 
Also, a team might end up creating multiple certificates in the process, and deploy only one, in which case not defaulting to `auto-rotate` would also be relevant.

To ensure Lemur does not miss on any rotations, where the endpoint is known to Lemur, we have created the celery task, which just turns on `auto-rotate`, where applicable: 
https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L902

havron:
> Thanks for your continuous contributions @havron to Lemur.
> 
> Adding the flexibility to define the default rotation value would potentially be of value to certain use-cases.
> 
> We have noticed that not all certificates can be rotated by Lemur, e.g., if the endpoint is not visible to Lemur, which would require manual rotation by the respective team.
> Also, a team might end up creating multiple certificates in the process, and deploy only one, in which case not defaulting to `auto-rotate` would also be relevant.
> 
> To ensure Lemur does not miss on any rotations, where the endpoint is known to Lemur, we have created the celery task, which just turns on `auto-rotate`, where applicable:
> https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L902

Thanks @hosseinsh for feedback, I really appreciate it! 🙂  I see your point about being careful about creating this default if it's difficult to guarantee for all certs.

Honestly, the main use case I'd been thinking of are certificates with an Issuer and Destination(s), but no Endpoints. Certs in that category can still benefit from celery's [certificate_reissue()](https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L599) task (which reuses `rotation` for eligibility checks). To your point on manual rotation, I actually had written a new Notification `reissue` event type, with the premise being that teams creating certificates with Destinations (but no Endpoints) may still want their certs automatically reissued and re-uploaded to their Destinations, and then get notified when they need to trigger a re-deployment to their final endpoints.

If that above use case makes sense, maybe I could rework this PR (or open a new one) to instead:
1.  Add a new celery task for `enable_autorotate_for_certs_attached_to_destination`
2. Add a new `reissue` event type notification sent by [certificate_reissue()](https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L599) with either a PendingCertificate or Certificate in the notification data (depending on whether it's issued synchronously or not - this is where https://github.com/Netflix/lemur/pull/3612 was going!)

hosseinsh:
Having a celery job to enable auto-rotate for certs with destinations sounds  certainly promising. We could optionally add blocked/allowed type of destinations, like enable auto-rotate for any S3 type of destinations, just as an example.  

@jtschladen has been super productive with adding new types of notifications. The "manual action required" after rotation, due to missing endpoints, is my favorite so far.

havron:
> Having a celery job to enable auto-rotate for certs with destinations sounds certainly promising. We could optionally add blocked/allowed type of destinations, like enable auto-rotate for any S3 type of destinations, just as an example.
> 
> @jtschladen has been super productive with adding new types of notifications. The "manual action required" after rotation, due to missing endpoints, is my favorite so far.

Yeah that new notification is great - I've been working on a Slack equivalent, but that's for another PR 😉 it definitely bolsters the case for turning on autorotate for certs without an endpoints, since they still can autorenew+upload to destinations.

I just updated the PR to reflect the new approach of a celery task; let me know if this is more along the lines of what we're looking for!

lemur/certificates/service.py
```diff
@@ -176,6 +176,22 @@ def get_all_certs_attached_to_endpoint_without_autorotate():
     )
 
 
+def get_all_certs_attached_to_destination_without_autorotate():
```
hosseinsh, line 179: what do you think about adding an optional destination `plugin_name`? this would allow for use-cases, where for instance only one type of destination should be auto-rotated. For instance, auto-rotate only `aws-s3`.
In case no `plugin_name` is provided, all destinations could be retrieved. 

the cli could then read the plugin_name from `ENABLE_AUTO_ROTATE_DESTINATION_TYPE`

hosseinsh (commented):
I like this new approach, thanks for the adjustment. I think I would like to make use of this, scoped to certain destination types.

Btw, would be great to add an example config [here](https://github.com/Netflix/lemur/blob/3fbcbc27174622b6075b492eccc1a42530afc5f0/docker/src/lemur.conf.py#L127).
Noticing, documentation about this feature would also make it more discoverable.

lemur/certificates/service.py
havron: Perfect, added!

havron:
This all sounds good to me @hosseinsh; I just added each of your suggestions 🎉

hosseinsh (approved):
Awesome PR @havron. Thanks for adding this feature to Lemur.
Merging shortly.

PR #3611, opened 2021-06-03
ACME External Account Binding Support

hosseinsh:
historically, the ACME plugin like other plugins has been sourcing the API credentials via the config file. The ACME plugin, however, could be used with multiple CAs. Hence, with this PR, we support adding the API credentials in the CA options. We encrypt the credentials before storage in Lemur. 

More specifically, when adding a new ACME issuer, one can now set `acme_regr` and `acme_private_key` (encrypted by Lemur).

**External Account Binding (EAB):**
Lemur now supports EAB, and can create a new account based on provisioned EAB credentials (`kid`, and `hmac_key`).

### DIFF
--- a/lemur/plugins/lemur_acme/plugin.py
+++ b/lemur/plugins/lemur_acme/plugin.py
@@ -88,6 +88,20 @@ class ACMEIssuerPlugin(IssuerPlugin):
             "type": "str",
             "required": False,
             "helpMessage": "HMAC key for the external account.",
+        },
+        {
+            "name": "acme_private_key",
+            "type": "textarea",
+            "default": "",
+            "required": False,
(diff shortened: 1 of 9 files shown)


lemur/plugins/lemur_acme/plugin.py
```diff
@@ -358,7 +416,12 @@ def create_authority(options):
         :param options:
         :return:
         """
-        role = {"username": "", "password": "", "name": "acme"}
+        name = "acme"
+        print(options)
```
charhate, line 420: Probably added for debugging?

hosseinsh: good catch, removed

charhate (approved):

PR #3612, opened 2021-06-05
Fix immediate (synchronous) acme creation

havron:
A seldom-used, but useful feature is the ability to create certificates immediately with `ACMEIssuerPlugin` (specifically for `AcmeDnsChallenge`), rather than [passing the work asynchronously](https://github.com/Netflix/lemur/blob/master/lemur/certificates/service.py#L476) to celery. I noticed that this feature exists already in Lemur, but somewhere along the way, it drifted from changes to the method signatures of `get_authorizations()` and `finalize_authorizations()` and appeared to no longer work.

In addition to updating calls to acme authorizations, I've added an `@retry` decorator to a new `create_immediately()` method, with the goal of roughly recreating the retry logic that celery uses for [`fetch_acme_cert()`](https://github.com/Netflix/lemur/blob/master/lemur/common/celery.py#L307).

### DIFF
--- a/lemur/plugins/lemur_acme/challenge_types.py
+++ b/lemur/plugins/lemur_acme/challenge_types.py
@@ -20,6 +20,8 @@ from lemur.plugins.base import plugins
 from lemur.destinations import service as destination_service
 from lemur.plugins.lemur_acme.acme_handlers import AcmeHandler, AcmeDnsHandler
 
+from retrying import retry
+
 
 class AcmeChallengeMissmatchError(LemurException):
     pass
@@ -192,16 +194,18 @@ class AcmeDnsChallenge(AcmeChallenge):
         authority = issuer_options.get("authority")
         create_immediately = issuer_options.get("create_immediately", False)
         acme_client, registration = self.acme.setup_acme_client(authority)
+        domains = self.acme.get_domains(issuer_options)
         dns_provider = issuer_options.get("dns_provider", {})
 
         if dns_provider:
-            dns_provider_options = dns_provider.options
+            for domain in domains:
+                # Currently, we only support specifying one DNS provider per certificate, even if that
+                # certificate has multiple SANs that may belong to different provid
+                self.acme.dns_providers_for_domain[domain] = [dns_provider]
+
             credentials = json.loads(dns_provider.credentials)
             current_app.logger.debug(
                 "Using DNS provider: {0}".format(dns_provider.provider_type)
-            )
-            dns_provider_plugin = __import__(
-                dns_provider.provider_type, globals(), locals(), [], 1
             )
             account_number = credentials.get("account_id")
             provider_type = dns_provider.provider_type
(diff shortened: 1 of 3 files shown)


lemur/plugins/lemur_acme/challenge_types.py
```diff
@@ -192,17 +195,19 @@ def create_certificate(self, csr, issuer_options):
         authority = issuer_options.get("authority")
         create_immediately = issuer_options.get("create_immediately", False)
         acme_client, registration = self.acme.setup_acme_client(authority)
+        domains = self.acme.get_domains(issuer_options)
         dns_provider = issuer_options.get("dns_provider", {})
 
         if dns_provider:
-            dns_provider_options = dns_provider.options
+            for domain in domains:
+                # Currently, we only support specifying one DNS provider per certificate, even if that
```
havron, line 203: This is lifted from [`get_ordered_certificate()`](https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_acme/plugin.py#L85), the roughly analogous (albeit it's for pending certs) method that celery calls.

lemur/plugins/lemur_acme/challenge_types.py
```diff
@@ -213,39 +218,40 @@ def create_certificate(self, csr, issuer_options):
...
-        pem_certificate, pem_certificate_chain = self.acme.request_certificate(
-            acme_client, authorizations, csr
+        pem_certificate, pem_certificate_chain = self.create_immediately(
+            acme_client, dns_authorization, csr
         )
         # TODO add external ID (if possible)
         return pem_certificate, pem_certificate_chain, None
 
+    @retry(stop_max_attempt_number=ACME_ADDITIONAL_ATTEMPTS, wait_fixed=5000)
+    def create_immediately(self, acme_client, order_info, csr):
```
havron, line 243: Since there's no pending certificate as we're issuing immediately, should we be sending metrics on attempts sort of like pending certs do?

havron: Added!

hosseinsh:
> Thanks for the extra context @hosseinsh, this all makes sense! Given that HTTP01 is now a supported acme challenge type for sync issuing, and also given the inherent limitations of DNS challenges for sync use: is there even a need to continue supporting the create_immediately parameter from issuer options anymore? It's not called anywhere else in Lemur (and wouldn't have worked if it was.) Perhaps this PR should implement its removal, rather than repair?

good point, the only use-case that comes to mind is testing and developing. Any thoughts on keeping your revised code for this purpose?


> I like the idea of mixing validation methods, at least to delegate wildcards to DNS challenges. Not sure if that would work best in sync or in get_ordered_certificate()

yeah, I haven't started thinking about it yet. Maybe when creating a certificate, the SANs could be checked to see if
(a) already validated or (b) supporting http01 validation.
In such cases, we could move to sync issuance. 
Otherwise, perform all applicable http01 validations, and move to async model, i.e., create a pending certificate. 
In this case, our current flow needs to be adjusted to skip all already validated domains.

This could become a new hybrid acme plugin type.

havron:
That sounds good to me! We can keep these for testing and development, at least. I added an explicit test here: https://github.com/Netflix/lemur/blob/87bbeaf4a79558bdfabeaa11627fba59cf6d4036/lemur/plugins/lemur_acme/tests/test_acme_dns.py#L311-L313

I also added a small docs update to clarify the intended use of `create_immediately`.

havron:
Also, I added those metrics mentioned above!

hosseinsh (approved):
Thanks @havron for raising this PR and spending cycles revising it.
merging it shortly.

PR #3635, opened 2021-06-18
Concatenate enum name instead of enum itself to avoid the error

mizzy:
This error shows up when revoking certificates with CFSSL plugin:

```
Failed to revoke: can only concatenate str (not "CRLReason") to str
```

So I concatenate enum name instead of enum itself to avoid this error.


hosseinsh (approved):
good catch @mizzy. Thanks for raising the PR, appreciated.
Will be merging shortly.

PR #3645, opened 2021-06-23
Create three new indexes which make searches faster

csine-nflx:


charhate (approved):

hosseinsh (approved):
Great PR, helping to reduce the search overhead significantly.
Thanks for adding the changes on `model.py`

Issue #3633, opened 2021-06-16
adding rotation toggle+filtering to frontend


havron:
👋 In my fork, I recently added the ability to toggle and filter a Certificate's rotation field directly on the certificates list (table) view - I wanted to ask here if it would be of value to raise a PR for it? Since it does add another column to the frontend, I didn't want to clutter the certificates view if there isn't enough general interest in this. For me, it's very useful to be able to quickly see+filter which certs aren't using autorotate and which are without going into the edit view for each cert!

Here's the code: https://github.com/havron/lemur/commit/3390c898ec35de2b9baff502284e3e1f69ca0f7a

It builds on @charhate's certificate table update API in https://github.com/Netflix/lemur/pull/3176 🙂

hosseinsh:
Hi @havron, this seems a nice addition to the Lemur UI. From experience, I notice the overhead of multiple clicks to see if a certificate has auto-rotate enabled, and this is one of the common operations. 
I will seek the opinion of the other members, and report back on the general interest.

hosseinsh:
@havron looks like there is popular demand for this feature. looking forward to a PR, whenever you get around to it.

havron:
Yay, happy to hear that! Just opened a PR: https://github.com/Netflix/lemur/pull/3646.

havron:
Closing now that it's merged

PR #3646, opened 2021-06-23
Add rotation toggle and filtering to table view

havron:
This PR tackles Issue https://github.com/Netflix/lemur/issues/3633 🙂 it surfaces the `rotation` bool switch for all certificates to the certificate table view in Angular, and provides filtering support to let users quickly see which certs have autorotation on, and which do not!

It builds on @charhate's new API from PR https://github.com/Netflix/lemur/pull/3176 for updating switches in the cert table view.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -934,20 +934,20 @@ class Certificates(AuthenticatedResource):
         """
         .. http:post:: /certificates/1/update/switches
 
-           Update certificate notification
+           Update certificate boolean switches for notification or rotation
 
            **Example request**:
 
            .. sourcecode:: http
 
-              POST /certificates/1/update/notify HTTP/1.1
+              POST /certificates/1/update/switches HTTP/1.1
               Host: example.com
               Accept: application/json, text/javascript
               Content-Type: application/json;charset=UTF-8
 
               {
-                 "notify": false,   <!-- Optional -->
-                 "rotation": false  <!-- Optional -->
+                 "notify": false,
+                 "rotation": false
               }
 
            **Example response**:
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -352,6 +352,7 @@ class CertificatesList(AuthenticatedResource):
         parser.add_argument("owner", type=inputs.boolean, location="args")
         parser.add_argument("id", type=str, location="args")
         parser.add_argument("active", type=inputs.boolean, location="args")
+        parser.add_argument("rotation", type=inputs.boolean, location="args")
         parser.add_argument(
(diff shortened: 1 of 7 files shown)


lemur/certificates/views.py
```diff
@@ -1516,7 +1519,7 @@ def put(self, certificate_id, data=None):
     Certificates, "/certificates/<int:certificate_id>", endpoint="certificate"
 )
 api.add_resource(
-    Certificates, "/certificates/<int:certificate_id>/update/notify", endpoint="certificateUpdateNotify"
+    Certificates, "/certificates/<int:certificate_id>/update/switches", endpoint="certificateUpdateSwitches"
```
jtschladen, line 1522: I don't think it's likely that anyone's using this endpoint programmatically (i.e. not via the UI) right now -- and therefore I think this change is acceptable -- but I do at least want to call out that this is a breaking/backwards incompatible change if they are.

havron: Thanks @jtschladen, that's a great point; we could bump the version in the changelog for it if desired! Although as you noted, I also suspect this isn't a highly used API endpoint given that `PUT /certificates/<id>` has been around for much longer, and offers more flexibility to programmatic callers looking to update certs.

I'd be happy to preserve `/update/notify` and add a new, separate `/update/rotation` endpoint, if we want to be cautious. I _think_ I'd just need to add a new `Resource` class for it, split up the service funcs, and use a customPOST in the Angular service call.

charhate: Thanks @havron for contributing the change. Code changes look good to me. We do suspect that it's not a highly used endpoint, so consolidating new changes in existing API makes sense. Given backward incompatibility, bumping version and adding changelog calling out this change will be helpful.

havron: Sounds good to me! I just added a changelog in https://github.com/Netflix/lemur/pull/3646/commits/3e94c84137a6a323deb955f54c6031e852a299a1, following the [new release process](https://lemur.readthedocs.io/en/latest/doing-a-release.html). Let me know if I can improve the release notes 🙂 @charhate

charhate: Thanks @havron, really appreciate your efforts in compiling detailed release notes.
Given all the change going in next release and specially API change that is not backward compatible, the minor version change seems more appropriate (so 0.10.0 instead of 0.9.1). I have taken liberty to commit this https://github.com/Netflix/lemur/pull/3646/commits/7edee9f444e21b6463f55a4aa2e84c005fde80d4, hope you do not mind.
I'll merge the PR soon and do a release.

charhate (approved):

PR #3654, opened 2021-06-29
Adds async reissue notifications

havron:
This PR allows celery to handle reissue notifications (`send_reissue_no_endpoints_notification` and `send_reissue_failed_notification`) for `PendingCertificate`s upon their resolution to a full `Certificate`, as opposed to sending reissue notifications while a cert is still in a pending state. This is a subtle change that unlocks, for example, the ability to further customize notifications to include more details about the final certificate, such as its expiration date, rotation settings, and so forth.

### DIFF
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -28,7 +28,11 @@ from lemur.endpoints import cli as cli_endpoints
 from lemur.extensions import metrics
 from lemur.factory import create_app
 from lemur.notifications import cli as cli_notification
-from lemur.notifications.messaging import send_pending_failure_notification
+from lemur.notifications.messaging import (
+    send_pending_failure_notification,
+    send_reissue_no_endpoints_notification,
+    send_reissue_failed_notification
+)
 from lemur.pending_certificates import service as pending_certificate_service
 from lemur.plugins.base import plugins
 from lemur.sources.cli import clean, sync, validate_sources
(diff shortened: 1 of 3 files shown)


lemur/common/celery.py
```diff
@@ -294,6 +299,8 @@ def fetch_acme_cert(id):
             pending_certificate_service.update(
                 cert.get("pending_cert").id, resolved=True
             )
+            if notify_reissue:
+                send_reissue_no_endpoints_notification(final_cert)
```
jtschladen, line 303: This method takes both the old and new cert in order to check endpoints from the old cert, but include information about the new cert (since in the synchronous case the endpoints aren't yet moved over when the notification is sent). I _think_ that written like this, you're going to get an error trying to dump the new certificate data. 

That said, I don't actually know the ACME flow very well, so I'm not sure whether you'd need to fetch the old cert or if you could write a second version that exclusively uses the new cert.

havron: Oh wow, I totally missed its method signature, whoops! Thanks for pointing that out. I'd written a different notification in my fork that accepted just the new cert and had mostly copied it over here 😅  perhaps I could propagate the old cert's `id` to celery (or the entire `Certificate` object) so that it could call the reissue notification correctly with both the old and final (new) certs?

FWIW, my main interest in letting certs resolve first is to allow the new cert's expiration date and id/final URL be included in the final notification - not sure if that changes anything, but figured I'd mention it.

jtschladen: Yeah, updated version looks good to me.

That motivation makes sense - I didn't test the notifications with pending certs when I added them, so it sounds like a gap I missed. Thanks for fixing it!

jtschladen (approved):

jtschladen:
This has been merged. Thanks for the contribution @havron !

PR #3657, opened 2021-07-01
Use directory object fetched from ACME server

bobmshannon:
This change updates the ACME client to configure itself using the [directory](https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.1) object [fetched](https://github.com/certbot/certbot/blob/master/acme/acme/client.py#L824) from the ACME server when the client is initialized.

Without this change, issuing certificates from some ACME authorities with external account binding support enabled will fail. I'm able to reproduce this issue using DigiCert's ACME API, but it should also be reproducible using a test server such as [Pebble](https://github.com/letsencrypt/pebble) (which is what DigiCert appears to be using for their ACME server) as well:

```json
{"detail":"the 'url' field in external account binding differs from outer JWS","status":400,"type":"malformed"}
```


hosseinsh (approved):
Hey @bobmshannon, thanks for raising this PR. Testing looks good on my side. Going to Merge shortly.

PR #3677, opened 2021-07-12
Deduplicate domains table

csine-nflx:
Background: Whenever a new certificate is minted, every domain in the CN and SANs fields are written as new entries in the domains table.  This has let to hundreds and even thousands of duplicate domains listed in the domains table which slows down queries and leads to table bloat.

This change changes the way that new entries are added to the domains table.  First, it is searched for the current domain, and if found, an association is created with the lowest id.  This change will work even if the next step is not performed, but is not ideal. A new db migration script has been created which will "de-dup" the domains table and update the certificate associations.  This script, however, runs in series and if the table contains many entries, may take days or even weeks to complete.  A better migration/update script may be needed for larger Lemur installations.

### DIFF
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -169,7 +169,10 @@ class Certificate(db.Model):
     sources = relationship(
         "Source", secondary=certificate_source_associations, backref="certificate"
     )
+
+    # This is set by default, do we need to update it?
     domains = association_proxy('certificate_associations', 'domain')
+
     roles = relationship("Role", secondary=roles_certificates, backref="certificate")
     replaces = relationship(
         "Certificate",
@@ -242,7 +245,11 @@ class Certificate(db.Model):
         self.dns_provider_id = kwargs.get("dns_provider_id")
 
         for domain in defaults.domains(cert):
+            # Currently, this creates a new Domain object for each domain, and a new association for each.
             self.domains.append(Domain(name=domain))
+
+            # New Logic
+            # Create a new association with existing Domain if it exists, otherwise, create a new Domain.
 
         # Check integrity before saving anything into the database.
         # For user-facing API calls, validation should also be done in schema validators.
--- a/lemur/certificates/models.py
+++ b/lemur/certificates/models.py
@@ -245,11 +245,15 @@ class Certificate(db.Model):
         self.dns_provider_id = kwargs.get("dns_provider_id")
 
         for domain in defaults.domains(cert):
-            # Currently, this creates a new Domain object for each domain, and a new association for each.
-            self.domains.append(Domain(name=domain))
+            # Avoid circular import.
+            from lemur.domains import service as domain_service
 
-            # New Logic
-            # Create a new association with existing Domain if it exists, otherwise, create a new Domain.
+            # Create an association with an existing Domain, otherwise, create a new Domain.
+            existing_domain = domain_service.get_by_name_sorted_by_id(domain, "asc")
+            if existing_domain:
+                self.domains.append(existing_domain[0])
+            else:
+                self.domains.append(Domain(name=domain))
 
         # Check integrity before saving anything into the database.
         # For user-facing API calls, validation should also be done in schema validators.
--- a/lemur/domains/service.py
+++ b/lemur/domains/service.py
@@ -42,6 +42,7 @@ def get_by_name(name):
     """
     return database.get_all(Domain, name, field="name").all()
 
+
 def get_by_name_sorted_by_id(name, sort_dir="asc"):
     """
     Fetches domain by its name
--- a/lemur/domains/service.py
+++ b/lemur/domains/service.py
@@ -42,6 +42,16 @@ def get_by_name(name):
     """
     return database.get_all(Domain, name, field="name").all()
 
+def get_by_name_sorted_by_id(name, sort_dir="asc"):
+    """
+    Fetches domain by its name
+
+    :param name:
+    :param sort_dir:
+    :return:
+    """
+    return database.get_all_sorted(Domain, name, field="name", sort_by="id", sort_dir=sort_dir).all()
+
 
 def is_domain_sensitive(name):
     """
@@ -65,6 +75,10 @@ def create(name, sensitive):
     :param sensitive:
     :return:
     """
+    domain = get_by_name(name)
+    if domain:
+        return domain
+
     domain = Domain(name=name, sensitive=sensitive)
     return database.create(domain)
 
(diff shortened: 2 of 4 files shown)


lemur/certificates/models.py
```diff
@@ -242,7 +245,15 @@ def __init__(self, **kwargs):
         self.dns_provider_id = kwargs.get("dns_provider_id")
 
         for domain in defaults.domains(cert):
-            self.domains.append(Domain(name=domain))
+            # Avoid circular import.
+            from lemur.domains import service as domain_service
+
+            # Create an association with an existing Domain, otherwise, create a new Domain.
+            existing_domain = domain_service.get_by_name_sorted_by_id(domain, "asc")
```
jtschladen, line 252: Would it be helpful to add a limit to the query to avoid getting potentially thousands of results back here (if the migration hasn't been run)?

lemur/domains/service.py
```diff
@@ -65,6 +76,10 @@ def create(name, sensitive):
     :param sensitive:
     :return:
     """
+    domain = get_by_name(name)
```
hosseinsh, line 79: shall we revise the comments to clarify the revised behavior of `create`?

lemur/migrations/versions/614f0f51b553_.py
```diff
@@ -0,0 +1,91 @@
...
+                        f"Fix the sensitivity conflicts and run again.")
+                    return
+
+                # Update association to point to single domain entry
+                conn.execute(
+                    text(
+                        "UPDATE certificate_associations SET domain_id=:domain_id WHERE domain_id=:cur_id"
+                    ).bindparams(domain_id=domain_id, cur_id=cur_id)
+                )
+                # Delete current domain entry
```
hosseinsh, line 80: should we log which domains (id, name, cert_id) are being deleted?

lemur/certificates/models.py
charhate: Since this is intended to get domain with lowest id for the matching name, it can fetch only first record than fetching all and using `existing_domain[0]`
e.g. `return database.get_all_sorted(Domain, name, field="name", sort_by="id", sort_dir=sort_dir).first()`
But to Jasmine's point, does this run in a reasonable time for a large dataset if migration hasn't been run?

lemur/migrations/versions/614f0f51b553_.py
```diff
@@ -0,0 +1,91 @@
...
+    # Do Upgrade
+    dedup_domains_table()
+
+    log.info("---Upgrade Complete - Total %s seconds ---\n" % (time.time() - start_time))
+
+
+def downgrade():
+    log.info("\n*** Request to downgrade at (%s) ***\n" % datetime.datetime.now())
+    log.info("It is not possible to downgrade this change since duplicated domains were already deleted.")
+    log.info("If you need to restore the duplicated domains for some reason, you will need to restore from backup.")
```
charhate, line 46: Could be worth adding this context to the comment from top of the migration script

lemur/migrations/versions/614f0f51b553_.py
```diff
@@ -0,0 +1,91 @@
...
+                        "UPDATE certificate_associations SET domain_id=:domain_id WHERE domain_id=:cur_id"
+                    ).bindparams(domain_id=domain_id, cur_id=cur_id)
+                )
+                # Delete current domain entry
+                conn.execute(
+                    text(
+                        "DELETE FROM domains WHERE id=:cur_id"
+                    ).bindparams(cur_id=cur_id)
+                )
+                commit()
```
charhate, line 86: Probably move this `commit` outside the loop (inside outer for).
Consider a case of domain with 5 duplicate entries. If the sensitivity conflict occurs at 4th entry, it will stop migration and display error, but 2nd and 3rd entry would be deleted by then.

lemur/migrations/versions/614f0f51b553_.py
```diff
@@ -0,0 +1,91 @@
...
+    log.info("It is not possible to downgrade this change since duplicated domains were already deleted.")
+    log.info("If you need to restore the duplicated domains for some reason, you will need to restore from backup.")
+
+
+def dedup_domains_table():
+    conn = op.get_bind()
+
+    # Loop through all domains and get a list of them by count
+    for name, count in conn.execute(text(
+            "SELECT name,count(*) FROM domains GROUP BY name HAVING COUNT(*) > 1 ORDER BY count(*) ASC")):
```
charhate, line 54: Not sure if this buys much improvement, but how about removing `ORDER BY count(*) ASC`. I could've missed it, the script does not use sorting specifically.

lemur/migrations/versions/614f0f51b553_.py
```diff
@@ -0,0 +1,91 @@
...
+    log.info("\n*** Request to downgrade at (%s) ***\n" % datetime.datetime.now())
+    log.info("It is not possible to downgrade this change since duplicated domains were already deleted.")
+    log.info("If you need to restore the duplicated domains for some reason, you will need to restore from backup.")
+
+
+def dedup_domains_table():
+    conn = op.get_bind()
+
+    # Loop through all domains and get a list of them by count
+    for name, count in conn.execute(text(
```
charhate, line 53: Info log in the beginning and end of this loop can be useful, like  `starting dedup for {domain}` and `finished dedup for {domain}`

lemur/certificates/models.py
csine-nflx: Search by domain is an indexed search, so should be fast.  However, there may be many duplicated records returned (if db migration hasn't been done), so returning just one is probably preferred here.

lemur/domains/service.py
csine-nflx: Good call

lemur/migrations/versions/614f0f51b553_.py
csine-nflx: Adding more context in top comment.  However, it will likely be updated again as the migration script is updated/optimized.

lemur/migrations/versions/614f0f51b553_.py
csine-nflx: That is correct.  It is done this way to start with the domains with less duplicates which made debugging the code easier.  There is no tangible reason to leave it in now, but the the improvement would be negligible.

lemur/migrations/versions/614f0f51b553_.py
csine-nflx: That would potentially lead to millions of entries.  Instead, it might be better to log something like the following for each domain.  Would give some tangible info about what is being de-duped.
"Removed {quantitiy} duplicate entries for {domain}"

lemur/migrations/versions/614f0f51b553_.py
csine-nflx: That would be helpful in that use case.  For some domains, there could be thousands of changes though.  Is there an issue with committing thousands of changes at one time?  Need to research a bit.

hosseinsh:
closing for new, to revisit in the near future.

PR #3679, opened 2021-07-14
Add the ability to exclude certs belonging to specific owners for the deployed expiring cert check

jtschladen:
In addition to the existing ability to exclude specific domains, this adds the ability to exclude certs based on owner. This is an alternative way to skip this check if specific owner need to be excluded.


hosseinsh (approved):
we might to also update the tests

https://github.com/Netflix/lemur/blob/80ea20923d7358e5642d32a283697a2326b7f32b/lemur/tests/test_certificates.py#L1486

PR #3680, opened 2021-07-15
Make some changes to the asynchronous reissue notification logic to avoid passing a Certificate directly

jtschladen:
With the existing code, we were seeing errors `TypeError: Object of type Certificate is not JSON serializable`. It looks like passing the `Certificate` directly to Celery wasn't working as intended. Instead, I'm refactoring the code to look up the replaced cert at notification time (when the `PendingCertificate` is resolved).

### DIFF
--- a/lemur/common/celery.py
+++ b/lemur/common/celery.py
@@ -232,13 +232,12 @@ def report_revoked_task(**kwargs):
 
 
 @celery.task(soft_time_limit=600)
-def fetch_acme_cert(id, notify_reissue_cert=None):
+def fetch_acme_cert(id):
     """
     Attempt to get the full certificate for the pending certificate listed.
 
     Args:
         id: an id of a PendingCertificate
-        notify_reissue_cert: existing Certificate to use for reissue notifications, if supplied
     """
     task_id = None
     if celery.current_task:
@@ -299,8 +298,8 @@ def fetch_acme_cert(id, notify_reissue_cert=None):
             pending_certificate_service.update(
                 cert.get("pending_cert").id, resolved=True
             )
-            if notify_reissue_cert is not None:
-                send_reissue_no_endpoints_notification(notify_reissue_cert, final_cert)
+            if real_cert.replaces is not None:
+                send_reissue_no_endpoints_notification(real_cert.replaces, final_cert)
             # add metrics to metrics extension
             new += 1
         else:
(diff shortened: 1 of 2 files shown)


jtschladen:
Hi @havron - just wanted to let you know about this PR since this is an iteration on the code you recently added. Please let me know if you have concerns or other ideas!

lemur/common/celery.py
```diff
@@ -316,7 +315,7 @@ def fetch_acme_cert(id, notify_reissue_cert=None):
                 send_pending_failure_notification(
                     pending_cert, notify_owner=pending_cert.notify
                 )
-                if notify_reissue_cert is not None:
+                if pending_cert.notify and pending_cert.replaces is not None:
```
jtschladen, line 318: Does anyone know if `pending_cert.notify` will be set at this point? I haven't dug into the code to check yet.

hosseinsh: my understanding is that it would follow what was provided in the UI or API request.

charhate (approved):

lemur/common/celery.py
jtschladen: I'm interested specifically in the case of reissued certs - at what point in time is the `notify` field of the new cert copied over from the prior cert? I'm reasonable sure it's already set at this point (since line 316 uses `.notify` as well, which I didn't initially notice), but I also didn't yet identify the code that takes care of setting it.

havron: Thanks for tagging me @jtschladen! Ugh, sorry it errored, admittedly I hadn't tested passing the full cert object -- I'll be more careful in the future. I believe the cert copy happens as follows

Reissue task dumps old cert's data (I believe this includes the `notify` field):

https://github.com/Netflix/lemur/blob/31774a4f03383b9afa69494c6d63830c84663c95/lemur/certificates/service.py#L866

Then after adjustments to fields that aren't `notify`, it reaches the creation call:

https://github.com/Netflix/lemur/blob/31774a4f03383b9afa69494c6d63830c84663c95/lemur/certificates/service.py#L935

Then the created cert is committed to Postgres (either as `PendingCertificate` or full `Certificate`):

https://github.com/Netflix/lemur/blob/31774a4f03383b9afa69494c6d63830c84663c95/lemur/certificates/service.py#L489

All that should be happening before the [acme fetch task runs](https://github.com/Netflix/lemur/blob/31774a4f03383b9afa69494c6d63830c84663c95/lemur/certificates/service.py#L518).

Hope that seems to be correct! Thanks for catching this.

hosseinsh (approved):

lemur/common/celery.py
jtschladen: Thanks - that sounds right to me. Appreciate the input!

Issue #3628, opened 2021-06-09
Installation error (lemur 0.9.0, ubuntu 20.04, python 3.7)


mced:
Hello,

I'm trying installing lemur 0.9.0 (from GitHub release) on Ubuntu 20.04 & python 3.7 but I have the following error when running `make release` which seems legit to me:

```
                  [16:35:06] 'dev:styles' errored after 16 ms
                  [16:35:06] Error: File not found with singular glob: /www/lemur/lemur-0.9.0/bower_components/angular-loading-bar/src/loading-bar.css (if this was purposeful, use `allowEmpty` option)
                      at Glob.<anonymous> (/www/lemur/lemur-0.9.0/node_modules/glob-stream/readable.js:84:17)
                      at Object.onceWrapper (events.js:286:20)
                      at Glob.emit (events.js:198:13)
                      at Glob.EventEmitter.emit (domain.js:448:20)
                      at Glob._finish (/www/lemur/lemur-0.9.0/node_modules/glob/glob.js:195:8)
                      at done (/www/lemur/lemur-0.9.0/node_modules/glob/glob.js:180:14)
                      at Glob._processSimple2 (/www/lemur/lemur-0.9.0/node_modules/glob/glob.js:686:12)
                      at /www/lemur/lemur-0.9.0/node_modules/glob/glob.js:674:10
                      at Glob._stat2 (/www/lemur/lemur-0.9.0/node_modules/glob/glob.js:770:12)
                      at lstatcb_ (/www/lemur/lemur-0.9.0/node_modules/glob/glob.js:762:12)
                      at RES (/www/lemur/lemur-0.9.0/node_modules/inflight/inflight.js:31:16)
                      at f (/www/lemur/lemur-0.9.0/node_modules/once/once.js:25:25)
                      at FSReqWrap.oncomplete (fs.js:153:21)
                  [16:35:06] 'build:inject' errored after 17 ms
                  [16:35:06] 'build:html' errored after 17 ms
                  [16:35:06] 'build' errored after 126 ms
```

```
$ ls -alh /www/lemur/lemur-0.9.0/bower_components/angular-loading-bar/src/loading-bar.css
ls: cannot access '/www/lemur/lemur-0.9.0/bower_components/angular-loading-bar/src/loading-bar.css': No such file or directory
$ ls -alh /www/lemur/lemur-0.9.0/bower_components/angular-loading-bar/src/
[...]

PR #3681, opened 2021-07-16
Make installation passed with Makefile

mced:
Signed-off-by: Cédric Menassa <cedric.menassa@gmail.com>

related to https://github.com/Netflix/lemur/issues/3628


charhate:
This will not be an appropriate change. The issue is fixed though, additional details in issue #3628

PR #3707, opened 2021-08-04
Digicert Session re-init after wrong header issues

hosseinsh:
This PR also improves the retry strategy. Before retries applied only to failed DNS lookups, socket connections and connection timeouts, but not to requests where data has made it to the server.
With the new retry strategy we also covers HTTP status code 400, 406, 500, 502, 503, 504.

This is also applied to Entrust Plugin

https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/

### DIFF
--- a/lemur/plugins/lemur_digicert/plugin.py
+++ b/lemur/plugins/lemur_digicert/plugin.py
@@ -347,6 +347,13 @@ class DigiCertIssuerPlugin(IssuerPlugin):
             }
         )
 
+        # max_retries applies only to failed DNS lookups, socket connections and connection timeouts,
+        # never to requests where data has made it to the server.
+        # we Retry we also covers HTTP status code 400, 406, 500, 502, 503, 504
+        retry_strategy = Retry(total=3, backoff_factor=0.1, status_forcelist=[400, 406, 500, 502, 503, 504])
+        adapter = requests.adapters.HTTPAdapter(max_retries=retry_strategy)
+        self.session.mount("https://", adapter)
+
         self.session.hooks = dict(response=log_status_code)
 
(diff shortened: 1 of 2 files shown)


lemur/plugins/lemur_digicert/plugin.py
```diff
@@ -327,6 +347,13 @@ def __init__(self, *args, **kwargs):
             }
         )
 
+        # max_retries applies only to failed DNS lookups, socket connections and connection timeouts,
+        # never to requests where data has made it to the server.
+        # we Retry we also covers HTTP status code 400, 406, 500, 502, 503, 504
+        retry_strategy = Retry(total=3, backoff_factor=0.1, status_forcelist=[400, 406, 500, 502, 503, 504])
```
charhate, line 353: Can you share more context behind retry specifically for 400? Since 4xx is a bad request (exception of 406, where we are trying to fix the header), will retry of same request help?

lemur/plugins/lemur_digicert/plugin.py
```diff
@@ -215,7 +216,23 @@ def handle_response(response):
     return response.json()
 
 
-def handle_cis_response(response):
+def reset_cis_session(session):
+    """
+    The current session might be in a bad state with wrong headers. Let's attempt to update the session back the initial
+    state.
```
charhate, line 222: `back the initial state.` --> `back to the initial state.`

charhate (approved):

lemur/plugins/lemur_digicert/plugin.py
hosseinsh: I had noticed the DigiCert 406 being accompanied with a 400, hence the thinking there. But we can remove it for now, and see how the state recovery after 406 helps.

PR #3708, opened 2021-08-05
server options only for sub-plugin options, not for option of type plugin

charhate:
Creating a destination with Export Plugin is resulting in below error.
```
KeyError: 'plugin_object'
  at lemur.destinations.service.create(lemur/destinations/service.py:32)
```
I believe the changes #3570 are relevant to options of type other than plugin, otherwise later in the code it errors on missing plugin details like one above.

### DIFF
--- a/lemur/schemas.py
+++ b/lemur/schemas.py
@@ -214,25 +214,26 @@ class PluginInputSchema(LemurInputSchema):
                         f"Unable to load plugin options. Slug: {data['slug']} Option {option_name}"
                     )
                 option["value"] = sub_data
+                plugin_options_validated.append(option)
+            else:
+                # validate user inputs for sub-plugin options and only accept "value" field from user
+                if data["plugin_object"]:
+                    try:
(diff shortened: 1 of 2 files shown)


jtschladen (approved):

lemur/schemas.py
```diff
@@ -214,25 +214,26 @@ def get_object(self, data, many=False):
...
-                        continue
-                    server_options_with_user_value["value"] = option_value
-                    plugin_options_validated.append(server_options_with_user_value)
-
-                except (ValueError, ValidationError) as e:
-                    raise ValidationError(
-                        f"Unable to validate plugin options. Slug: {data['slug']} Option {option_name}: {e}"
-                    )
+                plugin_options_validated.append(option)
+            else:
```
hosseinsh, line 218: minor: should this be an else if case for data["plugin_object"]?

hosseinsh (approved):

PR #3709, opened 2021-08-06
Membership plugin

charhate:
Changing implementation from #3414 to use a plugin allowing it to be generic. As mentioned in #3414, we are unaware of the usage of this code across the community. Hopefully current change makes it more flexible. This PR also cleans up configs specifically introduced for #3414.

The membership plugin currently offers 3 APIs
```
    def does_principal_exist(self, principal_email):
        raise NotImplementedError

    def does_group_exist(self, group_email):
        # check if a group (Team DL) exists

    def retrieve_user_memberships(self, user_id):
        # get a list of groups a user belongs to
```

If provided `USER_MEMBERSHIP_PROVIDER `, the API `retrieve_user_memberships` is used to retrieve user memberships details to store in Lemur. This is the same old behavior which is now driven via plugin.

Additionally, certificate creation now validated owner, if provided `USER_MEMBERSHIP_PROVIDER`. Currently, it has form validation to ensure it is an email. The UI suggests it to be a team email, however there is no validation that the email is valid. Now the validation can be added `does_group_exist` API from membership plugin.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -480,6 +480,9 @@ class CertificatesList(AuthenticatedResource):
            :statuscode 403: unauthenticated
 
         """
+        if not service.is_valid_owner(data["owner"]):
+            return dict(message=f"Invalid owner: check if email {data['owner']} exists as a Team DL"), 412
+
         role = role_service.get_by_name(data["authority"].owner)
 
         # all the authority role members should be allowed
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -481,7 +481,7 @@ class CertificatesList(AuthenticatedResource):
 
         """
         if not service.is_valid_owner(data["owner"]):
-            return dict(message=f"Invalid owner: check if email {data['owner']} exists as a Team DL"), 412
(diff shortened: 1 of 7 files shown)


lemur/certificates/views.py
```diff
@@ -480,6 +480,9 @@ def post(self, data=None):
            :statuscode 403: unauthenticated
 
         """
+        if not service.is_valid_owner(data["owner"]):
+            return dict(message=f"Invalid owner: check if email {data['owner']} exists as a Team DL"), 412
```
hosseinsh, line 484: shall revise the language to group?

"Invalid owner: check if {data['owner']} is a valid group. individuals cannot be certificate owners."

hosseinsh (approved):
awesome work. Just minor comment around the error message.

lemur/certificates/views.py
charhate: Thanks for suggestion, I'm not sure how commonly terminology "group" is used over team dl or email. So I'll do a minor tweak in `Invalid owner: check if {data['owner']} is a valid group email. Individuals cannot be certificate owners.`

PR #3716, opened 2021-08-09
Entrust deactivate resiliency

hosseinsh:
In Entrust test environment, one can issue test certificates. However, each certificate needs to be deactivate to free up the certificate issuance quota.
This improvement catches the corner-cases of a certificate already having been deactivated, or when the certificate cannot be found.

### DIFF
--- a/lemur/plugins/lemur_entrust/plugin.py
+++ b/lemur/plugins/lemur_entrust/plugin.py
@@ -301,8 +301,8 @@ class EntrustIssuerPlugin(IssuerPlugin):
 
         # Let's first check the status of the certificate
         base_url = current_app.config.get("ENTRUST_URL")
-        deactivate_url = f"{base_url}/certificates/{certificate.external_id}"
-        response = self.session.get(deactivate_url)
+        status_url = f"{base_url}/certificates/{certificate.external_id}"
+        response = self.session.get(status_url)
 
         try:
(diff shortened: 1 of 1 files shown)


lemur/plugins/lemur_entrust/plugin.py
```diff
@@ -293,12 +293,33 @@ def revoke_certificate(self, certificate, reason):
...
+            "external_id": f"{certificate.external_id}"
+        }
+
+        # Let's first check the status of the certificate
         base_url = current_app.config.get("ENTRUST_URL")
-        deactivate_url = f"{base_url}/certificates/{certificate.external_id}/deactivations"
-        response = self.session.post(deactivate_url)
-        metrics.send("entrust_deactivate_certificate", "counter", 1)
-        return handle_response(response)
+        deactivate_url = f"{base_url}/certificates/{certificate.external_id}"
```
jtschladen, line 304: Is this more like a `status_url`?

jtschladen (approved):

lemur/plugins/lemur_entrust/plugin.py
hosseinsh: oh good catch, I think we can designate it as such. It fetches the entire certificate at this point, to avoid failures on deactivation.

PR #3743, opened 2021-09-11
cert status check fix

charhate:
1. The unauthorized error from OCSP says `Responder Error: unauthorized (6)` which makes it enter the first if looking for `error`. Changing order to catch unauthorized ones separately.
2. When status could not be determined, [verify()](https://github.com/Netflix/lemur/blob/master/lemur/certificates/verify.py#L196) returns None (all exceptions are caught already, none thrown). Thus `cert.status = "valid" if status else "revoked"` does not look sufficient. This PR updates the status to unknown `if status is None:`

### DIFF
--- a/lemur/certificates/verify.py
+++ b/lemur/certificates/verify.py
@@ -69,7 +69,14 @@ def ocsp_verify(cert, cert_path, issuer_chain_path):
         raise Exception(f"OCSP lookup timed out: {url}, certificate serial number {cert.serial_number:02X}")
 
     p_message = message.decode("utf-8")
-    if "error" in p_message or "Error" in p_message:
+
+    if "unauthorized" in p_message:
+        # indicates the OCSP server does not know this certificate
+        metrics.send("check_revocation_ocsp_verify", "counter", 1, metric_tags={"status": "unauthorized", "url": url})
+        current_app.logger.warning(f"OCSP unauthorized error:{url}")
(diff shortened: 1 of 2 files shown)


lemur/certificates/verify.py
```diff
@@ -69,7 +69,14 @@ def ocsp_verify(cert, cert_path, issuer_chain_path):
         raise Exception(f"OCSP lookup timed out: {url}, certificate serial number {cert.serial_number:02X}")
 
     p_message = message.decode("utf-8")
-    if "error" in p_message or "Error" in p_message:
+
+    if "unauthorized" in p_message:
+        # indicates the OCSP server does not know this certificate
+        metrics.send("check_revocation_ocsp_verify", "counter", 1, metric_tags={"status": "unauthorized", "url": url})
+        current_app.logger.warning(f"OCSP unauthorized error:{url}")
```
jtschladen, line 76: Does it make sense to add serial number to this log as well? It doesn't seem very useful without that.

jtschladen (approved):

lemur/certificates/verify.py
charhate: Oops, let me add it

charhate: Actually the caller logs return exception as warning, so I'll remove this log. For other conditions too we do the same

Issue #3760, opened 2021-09-21
Reduce strictness of Vault URL field validator


bobmshannon:
Currently, the Vault destination plugin uses a [regexp validator](https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_vault_dest/plugin.py#L41) to validate the inputted vault URL:

```
^https?://[a-zA-Z0-9.:-]+$
```

The problem with this regular expression is that it does not allow Vault URLs with additional subpaths, e.g. `http://localhost:2845/vault` for example. This capability is required in environments where Vault is accessed through a reverse proxy service, for example.

PR #3763, opened 2021-09-22
Ensure all valid URLs are validated in Vault plugin

bobmshannon:
Closes https://github.com/Netflix/lemur/issues/3760.


lemur/plugins/lemur_vault_dest/plugin.py
```diff
@@ -22,6 +22,8 @@
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 
+from validators.url import regex as url_regex
```
wdauchy, line 25: nitpick: if I understood the import logic correctly, grouping third parties libraries would be better here than creating a group for each one

hosseinsh (approved):
Thanks @bobmshannon for the PR. The validators.url is certainly practical. The PR looks good to me.

PR #3764, opened 2021-09-22
Update hvac module version to 0.11.2

### LINKED ISSUE #3761: Add X-Vault-Request: true header to all requests issued to Vault
If the Lemur Vault plugins are configured to use the URL of a Vault Agent instead of a Vault Server, then requests may fail if they are missing the `X-Vault-Request: true` header. This is because Vault Agents may be configured with the `require_request_header` [option](https://www.vaultproject.io/api#the-x-vault-request-header) which causes all requests that are missing this header to be rejected. Barring a change to the default behavior of `hvac` (our Vault client library) I would propose always adding this header by default when we instantiate the plugin in Lemur for maximum compatibility. Given that the official Go client sets this header [by default](https://github.com/hashicorp/vault/blob/master/api/client.go#L459) this should be a safe and forward compatible change.

bobmshannon:
Closes #3761 by picking up https://github.com/hvac/hvac/pull/762 introduced in `hvac` 0.11.2. In this new version of `hvac`, the `X-Vault-Request: true` header is added to all requests which provides additional protections against SSRF vulnerabilities when interacting with a Vault Agent.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -210,7 +210,7 @@ gitpython==3.1.17
     #   bandit
 gunicorn==20.1.0
     # via -r requirements-docs.in
-hvac==0.11.0
+hvac==0.11.2
     # via -r requirements-docs.in
 idna==2.10
     # via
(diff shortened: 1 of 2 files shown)


hosseinsh (approved):

hosseinsh:
Thanks for the PR  @bobmshannon. Merging now.

Issue #3656, opened 2021-06-30
OAUTH2 with Azure issues


rvaneerdewijk:
Hi, I'm struggling to get OAUTH2 authentication working with Azure. It's possible I've left out important fields or are using the wrong values.

Config file options (with values changed for security reasons):

ACTIVE_PROVIDERS = ["oauth2"]
OAUTH2_SECRET = 'e0~r14.IIpG~-daPyAR2.QjsDgxaWjfo91'
OAUTH2_ACCESS_TOKEN_URL = "https://login.microsoftonline.com/1ede324e-872d-4f4b-8e56-ce3ed381b23c/oauth2/token"
OAUTH2_NAME = "Azure"
OAUTH2_CLIENT_ID = "12354aa4-116b-4d79-9b4a-0f4d3941f87b"
OAUTH2_REDIRECT_URI = "https://lemur.ourdomain.com/api/1/auth/oauth2"
OAUTH2_AUTH_ENDPOINT = "https://login.microsoftonline.com/1ede324e-872d-4f4b-8e56-ce3ed381b23c/oauth2/authorize"
OAUTH2_VERIFY_CERT = True

This adds an option to the login page. When I click the new button for logging into Azure, it opens up a pop-up verifying which Azure account I want to use. I select the account, and I immediately get the warning "Whoa there ... The supplied credentials are invalid". The log file has the following entries:

2021-06-30 14:43:15,772 DEBG 'lemur-web' stdout output:
Exception on /api/1/auth/oauth2 [POST]
Traceback (most recent call last):
  File "/www/lemur/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/www/lemur/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/www/lemur/lib/python3.8/site-packages/flask_restful/__init__.py", line 467, in wrapper
    resp = resource(*args, **kwargs)
  File "/www/lemur/lib/python3.8/site-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/www/lemur/lib/python3.8/site-packages/flask_restful/__init__.py", line 582, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/www/lemur/lemur/auth/views.py", line 548, in post
    id_token, access_token = exchange_for_access_token(
[...]

PR #3768, opened 2021-09-24
Add OAuth2 & Ping Config Params

JoeMcRobot:
Relates to the following issue - https://github.com/Netflix/lemur/issues/3656

By adding the `PING_URL` and `OAUTH2_URL` params this addresses applications that have a separate front-end and back-end and need the `url` and `redirectUri` to have different parameter values for OAuth2 to function properly. Both params are defaulted to `PING_REDIRECT_URI` and `OAUTH2_REDIRECT_URI` to ensure backwards compatibility.

Also, added the `TOKEN_AUTH_HEADER_CASE_SENSITIVE` param to fix the issue if the oauth provider has implemented the access token request authorization header in a case-sensitive way. This has also been implemented to ensure backwards compatibility.

### DIFF
--- a/docs/administration.rst
+++ b/docs/administration.rst
@@ -858,6 +858,13 @@ For more information about how to use social logins, see: `Satellizer <https://g
 
             PING_CLIENT_ID = "client-id"
 
+.. data:: PING_URL
+    :noindex:
+
+        ::
+
+            PING_URL = "https://<yourlemurserver>"
+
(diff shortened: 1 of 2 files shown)


hosseinsh:
thanks @JoeMcRobot for the thoughtful PR. Merging shortly.

hosseinsh (approved):

PR #3784, opened 2021-10-04
Another attempt at fixing decorator conflict

jtschladen:


hosseinsh (approved):
very strange, I was hoping the pinning to the highest minimum version would do the trick. 

```
The conflict is caused by:
    lemur[tests] develop depends on decorator==5.1.0
    lemur[tests] develop depends on decorator==4.4.2
```

hosseinsh:
do we need to remember to unpin this in the future?

jtschladen:
I was hoping that as well; unfortunately, it still seems to be failing pinned at >= 5.1.0, so I'm resorting to the other direction.

I did get the following error locally with the >= 5.1.0, but as I don't see it in Travis I'm unsure if it's ultimately the same problem or a different one:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
networkx 2.5.1 requires decorator<5,>=4.3, but you have decorator 5.1.0 which is incompatible.
```

At any rate, let's try this and go from there.

jtschladen:
Reverted that change and instead upgraded both `decorator` and `networkx` in the `.txt` dependency files. I tried this out in #3785 (which I don't intend to merge) and it seems to work.

PR #3807, opened 2021-10-13
Cloudfront endpoint support

douglasc-nflx:
Add support for CloudFront distributions as endpoints.

- Discover Distributions with IAM certs as an endpoint; concatenate multiple DNS names for the dnsname field.
- Allow limiting AWS source plugin by IAM certificate path for parity with destination plugin.
- Limit AWS source plugin instance to discover either ELBs or Distributions.
- Users must configure additional AWS source and destination plugins on the /cloudfront/ path for using IAM certificates with CloudFront.
- Remove the "-{path}" suffix being added to non-root path destinations. Lemur will expect IAM server certificate names to match the Lemur constructed name.
- Bonus: avoid unnecessary calls to AssumeRole during IAM syncs.


douglasc-nflx:
Superseded by PR #3835.

PR #3835, opened 2021-11-03
CloudFront distribution endpoint support.

douglasc-nflx:
This replaces PR #3807.  Mostly the same but using a separate table for Distribution DNS aliases and UI support for searching on it.

- Discover Distributions with IAM certs as an endpoint.
- Store CloudFront DNS name as the primary DNS name.
- Define a new table for endpoint alias, and make it searchable via Endpoint UI and API.
- Allow limiting AWS source plugin by IAM certificate path for parity with destination plugin.
- Limit AWS source plugin instance to discover either ELBs or Distributions.
- Users must configure additional AWS source and destination plugins on the /cloudfront/ path for using IAM certificates with CloudFront.
- Remove the "-{path}" suffix being added to non-root path destinations. Lemur will expect IAM server certificate names to match the Lemur constructed name.
- Bonus: avoid unnecessary calls to AssumeRole during IAM syncs.


lemur/plugins/lemur_aws/plugin.py
```diff
@@ -173,6 +173,77 @@ def get_elb_endpoints_v2(account_number, region, elb_dict):
...
+        return None
+
+    policy = dict(
+        name='cloudfront-none',
+        ciphers=[]
+    )
+    minimum_version = cert.get("MinimumProtocolVersion")
+    if minimum_version:
+        policy = dict(
+            name='cloudfront-%s' % minimum_version,
```
hosseinsh, line 228: readability

```
name=f'cloudfront-{minimum_version}'
```

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -195,10 +266,26 @@ class AWSSourcePlugin(SourcePlugin):
             "type": "str",
             "helpMessage": "Comma separated list of regions to search in, if no region is specified we look in all regions.",
         },
+        {
+            "name": "path",
+            "type": "str",
+            "helpMessage": "Only discover certificates with this path.",
```
hosseinsh, line 272: would validation add value here?

example usage:
https://github.com/Netflix/lemur/blob/fbf50b365cb0f7a0e9cae31dec1b853b958c45bb/lemur/plugins/lemur_aws/plugin.py#L425

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -211,6 +298,12 @@ def get_certificates(self, options, **kwargs):
         ]
 
     def get_endpoints(self, options, **kwargs):
+        if self.get_option("endpointType", options) == "cloudfront":
```
hosseinsh, line 301: I like how this is written to support backwards compatibility, in case `endpointType` is None.

hosseinsh (approved):
Awesome PR, and a new DB relation for Distribution Aliases is an excellent foresight.

lemur/endpoints/service.py
```diff
@@ -162,6 +175,11 @@ def render(args):
         else:
             query = database.filter(query, Endpoint, terms)
 
+        if terms[0] == "name":
+            alias_query = Endpoint.query.filter(
+                Endpoint.aliases.any(EndpointDnsAlias.alias.ilike("%{}%".format(terms[1]))))
```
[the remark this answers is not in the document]
douglasc-nflx: Done.

lemur/migrations/versions/189e5fda5bf8_.py
```diff
@@ -0,0 +1,32 @@
+"""empty message
```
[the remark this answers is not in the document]
douglasc-nflx: Added.

lemur/plugins/lemur_aws/cloudfront.py
```diff
@@ -0,0 +1,124 @@
...
+        config = response["DistributionConfig"]
+        viewer_cert = config["ViewerCertificate"]
+        if "IAMCertificateId" not in viewer_cert:
+            raise InvalidDistribution(distribution_id)
+        if iam_cert_id == viewer_cert["IAMCertificateId"]:
+            current_app.logger.warning(
+                "distribution {0} already assigned to IAM certificate {1}, not updated".format(
+                    distribution_id, iam_cert_id))
+            return
+        viewer_cert["IAMCertificateId"] = iam_cert_id
```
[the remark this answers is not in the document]
douglasc-nflx: viewer_cert here holds a reference to the whole sub-dictionary of "ViewerCertificate", which is also the one still referenced in the larger config. So when we change this field, it also changing what is contained within the whole config, that we pass back to the server below.

lemur/plugins/lemur_aws/plugin.py
douglasc-nflx: Missed these comments on the last PR, fixed now.

lemur/plugins/lemur_aws/plugin.py
douglasc-nflx: Added validation, a default, and consistency with the two plugins.  I also looked at making AWSSourcePlugin choose "/" or "/cloudfront/" default depending on the endpointType, but since we can't really do the same for AWSDestinationPlugin, I feel its better to make users be aware of this explicitly; I added to the helpMessage that this should be /cloudfront/ for CloudFront sources.

PR #3845, opened 2021-11-09
Make CN Optional

charhate:
As we are [moving away from use of CN](https://unmitigatedrisk.com/?p=381) and using only SAN for certificates, this PR updates lemur to treat CN as optional field, if SAN is provided
1. In Lemur UI, adding SAN removes validation on CN field making it optional
2. On the backend, CN is treated as optional field if authority has option `cn_optional` set to true in database. As of now, CN is mandatory for most of the public CA. Rather than relying on error message from third party in case of missing CN, Lemur will rely on options from authority to make CN a mandatory field. Example of option `{"name": "cn_optional", "type": "boolean", "value":true, "helpMessage": "Define if CN is an optional input when issuing certificates"}`

Today, Lemur uses CN in certificate names. Thus for the case of missing CN, this PR makes use of first SAN instead.
Also support is added to store IPAddress SAN. Currently only DNSName SAN is stored in Domains table.
Storing IPAddress SAN in Domains table allows it to be displayed on certificate page by navigating to` More --> SAN`. The Domain tab is renamed to SAN


hosseinsh (approved):
Awesome PR, with great testing coverage. 

I was only wondering, if the optional CN Authority flag `OPTIONAL_COMMON_NAME_AUTHORITIES , could be instead in the options of an Authority, instead of a new config parameter. This would also comes with certain pros and cons, specially considering migrations, and the effort of editing the option parameters of an existing Authority.

charhate:
@hosseinsh Great suggestion, I'll make the change.

lemur/certificates/service.py
```diff
@@ -458,7 +458,7 @@ def create(**kwargs):
         log_data = {
             "message": "Exception minting certificate",
             "issuer": kwargs["authority"].name,
-            "cn": kwargs["common_name"],
+            "cn": kwargs["common_name"] if "common_name" in kwargs else None,
```
douglasc-nflx, line 461: kwargs.get("common_name")

lemur/certificates/schemas.py
```diff
@@ -68,7 +68,7 @@ def default_notification(self, data):
 
 class CertificateInputSchema(CertificateCreationSchema):
     name = fields.String()
-    common_name = fields.String(required=True, validate=validators.common_name)
+    common_name = fields.String(validate=validators.common_name, missing="")
```
douglasc-nflx, line 71: The `missing=""` seems like its solving something that can't handle None? If something breaks without this, should be fixed to handle CN not present. If we do need this then a comment could be helpful.

lemur/static/app/angular/certificates/certificate/tracking.tpl.html
```diff
@@ -35,11 +35,11 @@
                ng-maxlength="64"
                ng-blur="certificate.attachCommonName()"
                ng-focus="certificate.removeCommonName()"
+               ng-change="certificate.checkCommonNameOptional(trackingForm)"
                required/>
 
         <p ng-show="trackingForm.commonName.$invalid && !trackingForm.commonName.$pristine" class="help-block">
-          You must
-          enter a common name and it must be less than 64 characters</p>
+          Common name is mandatory unless using Subject Alternate Names and it must be less than 64 characters</p>
```
douglasc-nflx, line 42: Suggestion:
Either common name or subject alternate name(s) are required, and they must be less than 64 characters

lemur/static/app/angular/certificates/view/view.tpl.html
```diff
@@ -173,7 +173,7 @@ <h2 class="featurette-heading">Certificates
                   </ul>
                 </uib-tab>
                 <uib-tab>
-                  <uib-tab-heading>Domains</uib-tab-heading>
+                  <uib-tab-heading>SAN</uib-tab-heading>
```
douglasc-nflx, line 176: I would do Subject Alternate Names if looks ok.

hosseinsh (approved):

lemur/static/app/angular/certificates/view/view.tpl.html
hosseinsh: **Subject Alternate Names** is definitely more readable than **SAN**.
The only problem is that the UI might not provide enough space for the not abbreviated version, though it might still work.

Right now, when expanding a certificate in Lemur by clicking on `MORE`, you see

Basic Info | Endpoints | Notifications | Roles | Destinations | **Domains** | Replaces

view

lemur/certificates/schemas.py
charhate: Current code assumes common name is always present in many places, since we had `required=True`. So I preferred to keep ` missing=""` avoid risk of not handling CN not present. I'll add comment as suggested.

lemur/static/app/angular/certificates/view/view.tpl.html
charhate: +1, rendering it in UI is the only issue given longer string `Subject Alternate Names`, was not the first choice for me as well otherwise.

PR #3854, opened 2021-11-10
Allow search by serial number via API

charhate:
Allow serial number as a filter get certs API, `http://host:port/api/1/certificates?serial=<serial_number>`

The serial number could be a decimal value (as stored in DB) or hex value either prefixed with `0x` or using `:` separated byte notation.


hosseinsh (commented):
Awesome PR, adding a much desired search capability to Lemur!

This is just a thought, would it add value to also support the hex format for the serial number?  we could then check for the format, and convert it on the fly to decimal format for the lookup, if needed.

lemur/certificates/views.py
```diff
@@ -263,7 +263,7 @@ def __init__(self):
     @validate_schema(None, certificates_list_output_schema_factory)
     def get(self):
         """
-        .. http:get:: /certificates
+        .. http:get:: /certificates?serial=82311058732025924142789179368889309156
```
hosseinsh, line 266: shall we add the other examples too: 

```
serial=218243997808053074560741989466015229225
serial=0xA43043DAB7F6F8AE115E94854EEB6529
serial=a4:30:43:da:b7:f6:f8:ae:11:5e:94:85:4e:eb:65:29
```

hosseinsh (approved):

lemur/certificates/views.py
charhate: Added (most)other supported filters too

PR #3857, opened 2021-11-11
Remove source associations from certs no longer discovered.

douglasc-nflx:
Add a step in sync_source to clean source and destination associations
from valid certificates which are no longer reported by the plugin
instance.

This will get certificates into the correct state after setting up a
separate cloudfront plugin, as existing /cloudfront/ certificates would
be discovered by a new source plugin, but the existing associations still need
to be cleaned up.

### DIFF
--- a/lemur/sources/service.py
+++ b/lemur/sources/service.py
@@ -203,9 +203,9 @@ def sync_certificates(source, user):
                  "gauge", len(certificates),
                  metric_tags={"source": source.label})
 
-    existing_by_id = {}
+    existing_certificates_with_source_by_id = {}
     for e in certificate_service.get_all_valid_certificates_with_source(source.id):
-        existing_by_id[e.id] = e
+        existing_certificates_with_source_by_id[e.id] = e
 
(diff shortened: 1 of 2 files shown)


lemur/sources/service.py
```diff
@@ -203,6 +203,10 @@ def sync_certificates(source, user):
                  "gauge", len(certificates),
                  metric_tags={"source": source.label})
 
+    existing_by_id = {}
```
hosseinsh, line 206: this is just a readability suggestion; 
I was wondering, if we can use a more meaning full variable name, since `exists` is also used, but with a different meaning. 

for example: 
`existing_certificates_with_source_by_id`

and: 
```
for cert in certificate_service.get_all_valid_certificates_with_source(source.id)
   existing_certificates_with_source_by_id[cert.id] = cert
```

hosseinsh (approved):

lemur/sources/service.py
douglasc-nflx: Done.

PR #3867, opened 2021-11-16
Ignore ELBs by account or tag

douglasc-nflx:
* If config option 'AWS_ELB_IGNORE_TAG' is set, ignore any ELB or ELBv2
reporting a tag with this name. Lemur will not show such ELBs in the UI or
attempt to rotate certificates on them.

* Add endpointType option 'none' for a source, to disable ELB discovery
for an entire account, while leaving certificate discovery active.


lemur/plugins/lemur_aws/elb.py
```diff
@@ -87,6 +87,48 @@ def get_all_elbs(**kwargs):
...
+        return elbs
+    ignore_tag = current_app.config.get("AWS_ELB_IGNORE_TAG")
+    if not ignore_tag:
+        return elbs
+    try:
+        keys = [elb[key_field] for elb in elbs]
+        client = kwargs.pop("client")
+        tags_list = client.describe_tags(**{arg_name: keys})["TagDescriptions"]
+        ignored_keys = {}
+        for tags in tags_list:
```
hosseinsh, line 115: This is where I get a bit lost, due to all the formats AWS endpoint tags can have. 
If it is not much trouble, an example formatting of `tags` and `elb` would it easier to follow the code.

hosseinsh (approved):

lemur/plugins/lemur_aws/elb.py
douglasc-nflx: Added some more docs and the response format, hope this helps.

PR #3869, opened 2021-11-20
Display endpoint source, add type tag to endpoint metrics

douglasc-nflx:
* Show the source an endpoint was discovered from in the UI.
* Add the endpoint 'type' as a tag to metrics.
* Optimize queries for Endpoints and Authorities pages to pre-load instead of lazy-load dependencies.


lemur/authorities/models.py
```diff
@@ -45,7 +45,6 @@ class Authority(db.Model):
         secondary=roles_authorities,
         passive_deletes=True,
         backref=db.backref("authority"),
-        lazy="dynamic",
```
charhate, line 48: May be just a speculation, could this affect loading of certificates page or something else if it is going to pre-load additional data?

douglasc-nflx: I don't think so.  This option still leaves the relation as lazy-loaded, but will use the default value of 'select' which means traversing an Authority.roles invokes the query and returns a result, whereas 'dynamic' would return a pre-composed query that offers further filtering. I don't see anything making use of that feature, and removing this option is needed to use Authority.roles in the joinedload().

charhate (approved):

douglasc-nflx:
Sorry this PR was forgotten for a while; some of the query changes caused an incorrect count to be returned.  I fixed those and also added filtering endpoints by source.  PTAL.

PR #3905, opened 2021-12-14
Call DescribeTags in batches of 20.

douglasc-nflx:


lemur/plugins/lemur_aws/elb.py
```diff
@@ -126,7 +126,12 @@ def _filter_ignored_elbs(elbs, key_field, arg_name, response_key_field, **kwargs
         keys = [elb[key_field] for elb in elbs]
         client = kwargs.pop("client")
         # {'TagDescriptions': [{'ResourceArn': 'string','Tags': [{'Key': 'string','Value': 'string'},]}]}
-        tags_list = client.describe_tags(**{arg_name: keys})["TagDescriptions"]
+        tags_list = []
+        # Restrict to 20 tags per call per elbv1 limits.
+        while len(keys):
+            next_keys = keys[:20]
+            keys = keys[20:]
+            tags_list += client.describe_tags(**{arg_name: next_keys})["TagDescriptions"]
```
hosseinsh, line 134: any chance we can catch the validation error more gracefully, and for example return an all elbs, without filtering?

douglasc-nflx: If an ignore tag was set, then it seems better to fail-closed and not potentially return ELBs we should have been ignoring if the tags API had worked.

I just sent separately PR #3906 which I think should address the core issue that we want to avoid expiring when we're not able to read for some reason, thoughts welcome.

hosseinsh: I like the idea of avoiding to expire certs from a given source, when we are failing to read from that source .

hosseinsh (approved):

PR #3906, opened 2021-12-15
Expire endpoints as part of source sync.

douglasc-nflx:
* Expiring endpoints depends on successful syncing of the source. If there
are unexpected issues reading from the source, we want to suspend expiration
and not assume the endpoints are truly gone yet. Provide for this dependency
by moving expiration within the sync celery task.
* Retain the TTL behavior and config variable.
* Propagate exceptions reading from the source and abort before expiration.


lemur/plugins/lemur_aws/plugin.py
```diff
@@ -325,13 +321,9 @@ def get_load_balancers(self, options, **kwargs):
                     capture_exception()
                     continue
 
+            continue
```
hosseinsh, line 324: not sure if `continue` here is intentional, since the remainder of the code would never be reached

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -306,17 +306,13 @@ def get_load_balancers(self, options, **kwargs):
...
         for region in regions:
-            try:
-                elbs = elb.get_all_elbs(account_number=account_number, region=region)
-                current_app.logger.info({
-                    "message": "Describing classic load balancers",
-                    "account_number": account_number,
-                    "region": region,
-                    "number_of_load_balancers": len(elbs)
-                })
-            except Exception as e:  # noqa
```
hosseinsh, line 317: is my understanding correct, that we are removing the try and catch, to handle the exception all the way through ` sync_endpoints` 

https://github.com/Netflix/lemur/blob/a69c9e01e846a4fb86f8351b1164608d19b7f344/lemur/sources/service.py#L75

in `sync`?

This would be a failure at the account level, which would break the entire syn cert, sync endpoints, and expire endpoints. I believe this is the intended behavior to fail close.

lemur/sources/cli.py
```diff
@@ -108,26 +108,34 @@ def execute_clean(plugin, certificate, source):
...
+    help="Time in hours, after which endpoint has not been refreshed, to remove endpoints from the source.",
+)
+def sync(source_strings, ttl):
     sources = validate_sources(source_strings)
     for source in sources:
         status = FAILURE_METRIC_STATUS
 
         start_time = time.time()
-        print("[+] Staring to sync source: {label}!\n".format(label=source.label))
+        print("[+] Staring to sync source: {label} expire ttl={ttl}h\n".format(label=source.label, ttl=ttl))
```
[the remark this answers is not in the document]
douglasc-nflx: Good suggestion, done.

lemur/sources/service.py
```diff
@@ -164,6 +168,24 @@ def sync_endpoints(source):
     return new, updated, updated_by_hash
 
 
+def expire_endpoints(source, ttl_hours):
```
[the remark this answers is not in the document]
douglasc-nflx: This was dead code and has now been removed, or you can consider it moved to sources/service.py.

lemur/common/celery.py
```diff
@@ -650,42 +650,6 @@ def certificate_rotate(**kwargs):
     return log_data
 
 
-@celery.task(soft_time_limit=3600)
```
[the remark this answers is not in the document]
douglasc-nflx: If you agree then we'll remove it.

lemur/plugins/lemur_aws/plugin.py
douglasc-nflx: Your understanding is correct, we want exceptions here to abort all syncing for the source before we do any DB writes.

lemur/plugins/lemur_aws/plugin.py
douglasc-nflx: Thanks for finding this bit of debugging. To catch something similar in the future I added a unit test to make sure plugin is always returning both types of ELBs.

lemur/plugins/lemur_aws/tests/test_plugin.py
```diff
@@ -87,3 +87,79 @@ def test_upload_acme_token(app):
...
+@mock_ec2()
+@mock_elbv2()
+def test_get_all_elb_and_elbv2s(app, aws_credentials):
+    from copy import deepcopy
+    from lemur.plugins.lemur_aws.elb import get_load_balancer_arn_from_endpoint
+    from lemur.plugins.base import plugins
+    from lemur.plugins.utils import set_plugin_option
+
+    cert_arn_prefix = "arn:aws:iam::123456789012:server-certificate/"
+    cert_arn = cert_arn_prefix + "plugin-test.netflix.net-R3-20211023-20220121"
```
hosseinsh, line 103: nit; a more generic domain that is not company specific might better suited here. We have mainly used `example.com` in the past.

hosseinsh (approved):
The new test is awesome, thank you.

lemur/plugins/lemur_aws/tests/test_plugin.py
douglasc-nflx: Good point, done.

PR #3924, opened 2022-01-05
Add Python 3.9 to Travis tests

hosseinsh:
Adding Python 3.9 to tests, and dropping Python 3.7, since it is breaking some of our package updates that require a minimum of Python 3.8, and Postgres 9.4, as it reached EOL.


.travis.yml
```diff
@@ -4,20 +4,10 @@ node_js:
...
-        postgresql: "9.4"
-        chrome: stable
-      services:
-        - xvfb
-    - name: "python3.7-postgresql-10-bionic"
-      dist: bionic
-      language: python
-      python: "3.7"
+      python: "3.9"
       env: TOXENV=py37
```
jtschladen, line 11: Should this be `TOXENV=py39`?

hosseinsh: ahh, good catch thanks.

jtschladen (approved):

jtschladen:
If you're trying to cover everything, the mean [README](https://github.com/Netflix/lemur/blob/master/README.rst) has a mention of 3.7.

.readthedocs.yml
```diff
@@ -15,7 +15,7 @@ formats: all
 
 # Set the version of Python and requirements required to build the docs
 python:
-   version: 3.7
+   version: 3.9
```
jtschladen, line 18: Apparently not officially supported yet, unless we switch to the beta image:
```
Problem in your project's configuration. Invalid "python.version": expected one of (2, 2.7, 3, 3.5, 3.6, 3.7, 3.8, pypy3.5), got 3.9
```

https://github.com/readthedocs/readthedocs.org/issues/7554

hosseinsh: then 3.8 should do it, which is actually the min version needed for Bumping flake8 from 3.9.2 to 4.0.1

PR #3931, opened 2022-01-06
Fix typo in expiring certificates periodic task name

bobmshannon:
When configuring Lemur tasks, I noticed that the task `identity_expiring_deployed_certificates` had a misspelling when searching for the corresponding metric which monitors how often this task runs. While this change is purely cosmetic, I figured it wouldn't be a bad idea to clean it up. I've renamed in such a way that existing users won't be broken until the old task is removed completely in a future release with breaking changes.


jtschladen:
Thanks @bobmshannon -- good catch. Appreciate the backwards compatibility.

jtschladen (approved):

PR #3944, opened 2022-01-20
refactoring code with List Comprehension

zjzh:
refactoring code with List Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？


hosseinsh:
I am going to decline this PR, since it is not adding value to our code-base. 

I refer to Philip Guyton on their elaborate comment on why such PRs are not helping open source projects.
https://github.com/rockstor/rockstor-core/pull/2351#issuecomment-1019463178

Issue #3947, opened 2022-01-22
Support AWS GovCloud partitions


bobmshannon:
When using the Lemur AWS plugin to manage endpoints which reside in AWS GovCloud, the ARNs (when assuming an IAM role, for example) must be prefixed with `arn:aws-us-gov` instead of `arn:aws`. It should be possible to support such a use-case by adding a new option  `LEMUR_AWS_PARTITION` (similar to `LEMUR_AWS_REGION`) and updating the code paths which currently hard code using `arn:aws` accordingly.

PR #3948, opened 2022-01-22
Add support for additional AWS partitions

bobmshannon:
This PR adds support for allowing the Lemur AWS plugin to be used with other [AWS partitions](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/Partitions.html), including `aws-us-gov` and `aws-cn`.

To achieve this a new configuration option called `LEMUR_AWS_PARTITION` has been added which modifies the ARNs the AWS plugin uses. Additionally, the STS wrapper may now optionally default to the region specified via `LEMUR_AWS_REGION`. This is to ensure that the credentials fetched by the `boto3` client use regional endpoints located within the AWS partition that the credentials were fetched from.

Closes https://github.com/Netflix/lemur/issues/3947.

### DIFF
--- a/lemur/plugins/lemur_aws/iam.py
+++ b/lemur/plugins/lemur_aws/iam.py
@@ -84,27 +84,27 @@ def get_registry_type_from_arn(arn):
     :param arn: IAM TLS certificate arn
     :return: iam or acm or unkown
     """
-    if arn.startswith("arn:aws:iam"):
+    if arn.startswith("arn:aws:iam") or arn.startswith("arn:aws-us-gov:iam") or arn.startswith("arn:aws-cn:iam"):
         return 'iam'
-    elif arn.startswith("arn:aws:acm"):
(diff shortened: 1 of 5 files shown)


lemur/plugins/lemur_aws/iam.py
```diff
@@ -84,27 +84,27 @@ def get_registry_type_from_arn(arn):
...
         return 'iam'
-    elif arn.startswith("arn:aws:acm"):
+    elif arn.startswith("arn:aws:acm") or arn.startswith("arn:aws-us-gov:acm") or arn.startswith("arn:aws-cn:acm"):
         return 'acm'
     else:
         return 'unknown'
 
 
-def create_arn_from_cert(account_number, region, certificate_name, path=''):
+def create_arn_from_cert(account_number, partition, certificate_name, path=''):
```
hosseinsh, line 95: good idea to remove region, since cert ARNs are per account

hosseinsh (approved):
Thanks for the great PR @bobmshannon, and the companying changes.

jtschladen (approved):

PR #3953, opened 2022-01-31
Use a unique role for each authority

jtschladen:
This PR modifies the authority creation flow to create a brand new, unique role for each authority, rather than a sharing a single issuer plugin role across all authorities using the same issuer plugin.

This change does NOT affect existing authorities. If desired, existing authority roles would need to be manually updated.

This also adds validation that an authority owner is valid (only if `USER_MEMBERSHIP_PROVIDER` is configured).

### DIFF
--- a/lemur/tests/test_certificates.py
+++ b/lemur/tests/test_certificates.py
@@ -153,8 +153,12 @@ def test_cleanup_after_revoke(session, issuer_plugin, crypto_authority):
     assert not revoked_cert.destinations
 
 
-def test_get_by_attributes(session, certificate):
-    from lemur.certificates.service import get_by_attributes
+def test_get_by_attributes(session, authority, user, certificate):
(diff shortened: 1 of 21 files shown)


hosseinsh (approved):
Great PR, with awesome test coverage.

lemur/tests/test_certificates.py
```diff
@@ -153,8 +153,12 @@ def test_cleanup_after_revoke(session, issuer_plugin, crypto_authority):
     assert not revoked_cert.destinations
 
 
-def test_get_by_attributes(session, certificate):
-    from lemur.certificates.service import get_by_attributes
+def test_get_by_attributes(session, authority, user, certificate):
```
jtschladen, line 156: Most of the changes in this file relate to making it able to run independently of all other tests. Without these changes, the test was failing because it depended on state established by `test_authorities.py`; it is now independent, and passes on its own.

PR #3959, opened 2022-02-02
Restrict multiple accounts on a certificate, by plugin

douglasc-nflx:
This clause only intends to prevent conflicts within the same plugin.
Using the same destination account across plugin types should be allowed.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -459,14 +459,15 @@ def create(**kwargs):
     """
     Creates a new certificate.
     """
-    # Validate destinations do not overlap accounts
+    # Validate destinations do not overlap accounts for the same plugin
     if "destinations" in kwargs:
-        dest_accounts = {}
+        dest_plugin_accounts = {}
         for dest in kwargs["destinations"]:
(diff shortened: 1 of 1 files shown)


lemur/certificates/service.py
```diff
@@ -459,14 +459,15 @@ def create(**kwargs):
     """
     Creates a new certificate.
     """
-    # Validate destinations do not overlap accounts
+    # Validate destinations do not overlap accounts for the same plugin
     if "destinations" in kwargs:
-        dest_accounts = {}
+        dest_plugin_accounts = {}
         for dest in kwargs["destinations"]:
+            plugin_accounts = dest_plugin_accounts.setdefault(dest.plugin_name, {})
```
jtschladen, line 466: Optional style suggestion: other code in Lemur uses `defaultdict(dict)` to achieve similar behavior in a slightly more concise way.

jtschladen (approved):

lemur/certificates/service.py
douglasc-nflx: Didn't know about defaultdict, nice. I'll leave as-is, setdefault pattern is fairly common too.

PR #3963, opened 2022-02-04
Support rotating endpoints with non-unique names via CLI

bobmshannon:


douglasc-nflx (approved):
Looks good to me, thank you for the submission.

lemur/endpoints/service.py
```diff
@@ -51,6 +51,20 @@ def get_by_name(name):
     return database.get(Endpoint, name, field="name")
 
 
+def get_by_name_and_source(name, source):
+    """
+    Retrieves and endpoint by it's name and source.
```
hosseinsh, line 56: minor language: 
`Retrieves an endpoint by it's name and source`

lemur/certificates/cli.py
```diff
@@ -239,7 +263,14 @@ def rotate(endpoint_name, new_certificate_name, old_certificate_name, message, c
...
         new_cert = validate_certificate(new_certificate_name)
-        endpoint = validate_endpoint(endpoint_name)
+        if source:
+            endpoint = validate_endpoint_from_source(endpoint_name, source)
+        else:
+            try:
+                endpoint = validate_endpoint(endpoint_name)
+            except MultipleResultsFound as e:
+                print("[!] Multiple endpoints found with name {0}, try narrowing the search down to an endpoint from a specific source by re-running this command with the --source flag.".format(endpoint_name))
+                sys.exit(1)
```
hosseinsh, line 273: shall we add a logging here, to ensure this is easier to debug, when run via automation and not manually via cli?

Also wondering if `sys.exit(1)` is a good option here. while totally legit for manual CLI call, when called from celery, we are not catching `SystemExit Exception`, and this would break all the remaining rotations.

hosseinsh (commented):
Thanks @bobmshannon for always making the effort to contributing upstream. Appreciated. 
Left one comment around how are handling termination, and also better logging.

lemur/endpoints/service.py
bobmshannon: Fixed :)

lemur/certificates/cli.py
bobmshannon: That's a good point that I missed when making this change.

What I've done to address this concern in the most recent commit is:
(1) emit logs via the application logged
(2) re-raise the `MultipleResultsFound` exception to ensure that the exception is captured by sentry and that a metric indicating that the `endpoint_rotation_job` failed is also emitted.

Hopefully this should match the existing behavior that is more friendly to automation like you've mentioned.

PR #3965, opened 2022-02-07
ACME Authority Validation Error

JoeMcRobot:
When trying to create a certificate using multiple ACME Authorities, the following error occurs when the `Certificate Authority` is changed via the select dropdown:

`{"message": "Validation Error.", "reasons": {"Schema": {"rule": ["Validity end must not be after 2022-12-29T23:59:59+00:00"]}}}`

The error is caused during the `create` function, which calls the `populateValidityDateAsPerDefault` function that sets the `certificate.validityEnd` end date past the `Certificate Authority` certificate expiration date.

A check has been added to skip the `populateValidityDateAsPerDefault` function if the `acme-issuer` plugin is being used.


douglasc-nflx (approved):
This is consistent with other UI code switching on acme-issuer slug.

hosseinsh:
thank you for the PR @JoeMcRobot

PR #3986, opened 2022-03-21
Issuer: option to drop cert from Chain

hosseinsh:
This option allows to drop a certificate from a chain. for example, the ECC-based chain from Let's Encrypt returns 
```
server cert
leaf cert
ISRG X2 (ECC root - Cross-signed)
ISRX X1 (RSA root)
```
it might be desirable to drop X1, if needed. Let's Encrypt is not offering an alternative chain presently for ECC, but likely in the future this could be supported via the alternative chain


lemur/common/utils.py
```diff
@@ -493,3 +494,21 @@ def is_json(json_input):
...
+    if full_chain == '':
+        return ''
+    full_chain_certs = CERT_PEM_REGEX.findall(full_chain.encode())
+    pem_certificate = OpenSSL.crypto.dump_certificate(
+        OpenSSL.crypto.FILETYPE_PEM,
+        OpenSSL.crypto.load_certificate(
+            OpenSSL.crypto.FILETYPE_PEM, ''.join(cert.decode() for cert in full_chain_certs[:-1])
+        ),
+    ).decode()
+    return pem_certificate
```
jtschladen, line 514: Is there any pre-existing PEM decoding in the flow(s) where this is used? I'm wondering if it could present any new decoding errors in places that previously wouldn't have noticed them (though that's probably for the best, as valid PEMs seems ideal).

lemur/plugins/lemur_acme/challenge_types.py
```diff
@@ -243,6 +251,13 @@ def create_certificate(self, csr, issuer_options):
         pem_certificate, pem_certificate_chain = self.create_certificate_immediately(
             acme_client, dns_authorization, csr
         )
+
+        if "drop_last_cert_from_chain" in authority.options \
+                and authority.options.get("drop_last_cert_from_chain") is True \
+                and pem_certificate_chain.count("BEGIN CERTIFICATE") > 1:
```
jtschladen, line 257: Does it make sense to move `pem_certificate_chain.count("BEGIN CERTIFICATE") > 1` into `drop_last_cert_from_chain` if it's common behavior?

jtschladen (approved):

charhate (approved):

lemur/common/utils.py
hosseinsh: thanks for pointing this out, looks like since this treatment is running already in the ACME handler, we should be good
https://github.com/Netflix/lemur/blob/dbb4894de82606a2adadacc66dfd07dffec8eff4/lemur/plugins/lemur_acme/acme_handlers.py#L139-L144

lemur/plugins/lemur_acme/challenge_types.py
hosseinsh: good point, revision will follow shortly

PR #3989, opened 2022-03-28
Dependency update

jtschladen:


hosseinsh (approved):

lemur/plugins/lemur_aws/tests/test_elb.py
```diff
@@ -54,7 +54,7 @@ def test_get_all_elbs(app, aws_credentials):
 @mock_sts()
 @mock_ec2()
 @mock_elbv2()
-def test_get_all_elbs_v2():
+def test_get_all_elbs_v2(app):
```
jtschladen, line 57: Unrelated to this change, but this test started failing due to a `moto` upgrade I reverted. This change is required to run this test individually.

PR #3990, opened 2022-03-30
Fix drop cert from chain

hosseinsh:
minor bug fix, and improved logging

### DIFF
--- a/lemur/plugins/lemur_acme/challenge_types.py
+++ b/lemur/plugins/lemur_acme/challenge_types.py
@@ -136,10 +136,13 @@ class AcmeHttpChallenge(AcmeChallenge):
             for authz in validationError.failed_authzrs:
                 for chall in authz.body.challenges:
                     if chall.error:
-                        current_app.logger.error(
-                            "ValidationError occured of type {}, with message {}".format(chall.error.typ,
-                                                                                         ERROR_CODES[chall.error.code]))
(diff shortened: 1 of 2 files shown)


lemur/plugins/lemur_acme/challenge_types.py
```diff
@@ -136,17 +136,20 @@ def create_certificate(self, csr, issuer_options):
...
-                            "ValidationError occured of type {}, with message {}".format(chall.error.typ,
-                                                                                         ERROR_CODES[chall.error.code]))
-            raise Exception('Validation error occured, can\'t complete challenges. See logs for more information.')
+                        current_app.logger.error(f"ValidationError occurred of type: {chall.error.typ}, "
+                                                 f"with message: {ERROR_CODES[chall.error.code]}, "
+                                                 f"detail: {chall.error.detail}")
+
+            raise Exception(f"ValidationError occurred of type: {chall.error.typ}, "
+                            f"with message: {ERROR_CODES[chall.error.code]}, "
+                            f"detail: {chall.error.detail}")
```
charhate, line 145: nit, the message could be in a variable that gets reused, makes it easier to modify and be consistent.

charhate (approved):

lemur/plugins/lemur_acme/challenge_types.py
hosseinsh: good advise, thanks

PR #4041, opened 2022-05-06
Sync endpoint: capture error and continue

charhate:
If encouraged any error in an endpoint sync, log and continue with the next endpoint

### DIFF
--- a/lemur/sources/service.py
+++ b/lemur/sources/service.py
@@ -168,6 +168,7 @@ def sync_endpoints(source):
         except Exception as e:
             current_app.logger.warning(f"Error in processing sync endpoint "
                                        f"dnsname={endpoint['dnsname']} port={endpoint['port']}: {format(e)}")
+            capture_exception()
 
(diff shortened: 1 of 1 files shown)


lemur/sources/service.py
```diff
@@ -86,84 +86,88 @@ def sync_endpoints(source):
...
-        else:
-            current_app.logger.debug("Endpoint Updated: {}".format(endpoint))
-            endpoint_service.update(exists.id, **endpoint)
-            updated += 1
+            else:
+                current_app.logger.debug("Endpoint Updated: {}".format(endpoint))
+                endpoint_service.update(exists.id, **endpoint)
+                updated += 1
+        except Exception as e:
+            current_app.logger.warning(f"Error in processing sync endpoint "
```
jtschladen, line 169: We use `capture_exception()` in many (but not all) situations like this; seems like we may want it here?

jtschladen (approved):

hosseinsh (approved):

Issue #3943, opened 2022-01-19
Add metric to monitor expiration dates of certificates in Lemur


bobmshannon:
Metric(s) which monitor the seconds until each certificate expires would be useful for quickly answering questions like "which certificates are expiring in 3 months?" and to build alert rules to detect deployed certificates that are expiring for example.

At a high level, this could potentially be implemented as another periodical task unless there are other recommended ways for doing so. Ideally, such a metric would provide tags to (1) quickly filter down on specific certificates based on `name` and (2) quickly filter down on certificates which have endpoints associated with them versus ones that do not.

Relates to #3870.

hosseinsh:
Hi @bobmshannon, thanks for initiating this conversation. 

Summarizing the mechanisms we have at hand today, and which we could expand or revise:

a) the `DASHBOARD` page of Lemur UI, gives some high level statistics, one of them being about expiring certs. 
https://github.com/Netflix/lemur/blob/master/lemur/certificates/service.py#L824-L833
The Dashboard could certainly benefit from some updates to make it more useful, for instance, being able to control the expiration window, and potentially new statistics. 
The dashboard won't however show the individual expiring certs. 

b) expiration notification
This is currently set two 30 days, and usually an opportunity to alert the individual teams, or the secop team
https://github.com/Netflix/lemur/blob/dbc60e749d4cca2d4ab2ef42d8b57563ee738fa2/docker/src/lemur.conf.py#L125-L138

To make easier for the security team to follow all the expirations, we also created a summary of upcoming expirations
https://github.com/Netflix/lemur/blob/dbc60e749d4cca2d4ab2ef42d8b57563ee738fa2/docker/src/lemur.conf.py#L139-L144

b) Notify_expiring_deployed_certificates
Lemur can also detect and alert when identifying an expiring cert is still deployed. This, however, only works for non wildcard certs, and when common ports are used.

https://github.com/Netflix/lemur/blob/7a80ae629b86a396887e89132ee5a00c6ffc2ecc/lemur/common/celery.py#L967

hosseinsh:
Regarding certificates which have endpoints associated with, there is a celery job to enable auto-rotation by default. We want to also simplify that, and enable auto-rotate when an endpoint with certificate is discovered.

bobmshannon:
Thanks for the overview @hosseinsh. These are all great notification mechanisms when used within the right context.

I think the value provided by augmenting these mechanisms with metrics is centered around two major use-cases:

[...]

PR #4084, opened 2022-06-01
Add metrics for certificate expiry

mitchcail:
Background:
The purpose of this PR is to introduce a task that grabs all non-expired and non-revoked certificates in lemur and emits a metric for the number of days until each certs expiration. This will give users the ability to fine tune their notifications and alerts around expiring certificates. For a bit more context see issue #3943.


lemur/common/celery.py
```diff
@@ -1047,3 +1047,37 @@ def identify_expiring_deployed_certificates():
...
+        cli_certificate.expirations_metrics()
+    except SoftTimeLimitExceeded:
+        log_data["message"] = "Time limit exceeded."
+        current_app.logger.error(log_data)
+        capture_exception()
+        metrics.send("celery.timeout", "counter", 1, metric_tags={"function": function})
+        return
+
+    metrics.send(f"{function}.success", "counter", 1)
+    return log_data
```
bobmshannon, line 1083: nit: newline

lemur/common/celery.py
```diff
@@ -1047,3 +1047,37 @@ def identify_expiring_deployed_certificates():
...
+        "message": "sending metrics for expiring certificates",
+        "task_id": task_id,
+    }
+
+    if task_id and is_task_active(function, task_id, None):
+        log_data["message"] = "Skipping task: Task is already active"
+        current_app.logger.debug(log_data)
+        return
+
+    current_app.logger.debug(log_data)
```
bobmshannon, line 1071: nit: remove unused debug logging

lemur/certificates/service.py
```diff
@@ -1274,3 +1274,44 @@ def allowed_issuance_for_domain(common_name, extensions):
...
+
+    for certificate in certificates:
+        try:
+            days_until_expiration = _get_cert_expiry_in_days(certificate.not_after)
+            has_active_endpoints = len(certificate.endpoints) > 0
+            is_replacement = len(certificate.replaces) > 0
+
+
+            metrics.send(
+                f"{certificate.name}.days_until_expiration",
```
bobmshannon, line 1295: nit: let's use a generic metric name here e.g. `lemur.certificate.days_until_expiry`

lemur/certificates/cli.py
```diff
@@ -1118,3 +1119,25 @@ def identify_expiring_deployed_certificates(exclude_domains, exclude_owners, com
...
+def expirations_metrics():
+    """
+    Iterates over all certificates and emits a metric for the days remaining for a certificate to expire.
+    This is used for building custom dashboards and alerts for certificate expiry.
+    """
+    try:
+        print("Starting to publish metrics for time left until cert expirations")
+        success, failure = send_certificate_expiration_metrics()
+        print(
+            f"Finished publishing metrics for time left until cert expirations! Sent: {success} "
```
bobmshannon, line 1134:
```suggestion
            f"Finished publishing metrics for time left until cert expirations! Sent: {success}"
```

lemur/certificates/service.py
```diff
@@ -1274,3 +1274,43 @@ def allowed_issuance_for_domain(common_name, extensions):
...
         is_authorized_for_domain(common_name)
+
+
+def send_certificate_expiration_metrics():
+    """
+    Iterate over each certificate and emit a metric for how many days until expiration.
+    """
+    success = failure = 0
+    # get eligible certificates
+    certificates = get_certificates()
```
bobmshannon, line 1285: This seems to filter out certificates which have notifications disabled and are not expiring within 30 days. We want this to monitor all non-expired certificates so this behavior is probably not what we want here. So I'd suggest creating another function here which does something similar as  `get_certificates()` but has the required filtering constraints.

mitchcail: Understood, so maybe something like,
```
certificates = database.db.session.query(Certificate).filter(
        Certificate.expired == false(),
        Certificate.revoked == false()
    )
```

bobmshannon (approved):

lemur/certificates/cli.py
```diff
@@ -1118,3 +1119,25 @@
...
+def expiration_metrics():
+    """
+    Iterates over all certificates and emits a metric for the days remaining for a certificate to expire.
+    This is used for building custom dashboards and alerts for certificate expiry.
+    """
+    try:
+        print("Starting to publish metrics for time left until cert expirations")
+        success, failure = send_certificate_expiration_metrics()
+        print(
+            f"Finished publishing metrics for time left until cert expirations! Sent: {success}"
```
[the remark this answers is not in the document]
mitchcail: Please let me know if anything needs to be actioned around this. I was trying to follow a similar pattern to this [method](https://github.com/Netflix/lemur/blob/master/lemur/notifications/cli.py#L52-L59).

lemur/certificates/service.py
```diff
@@ -1274,3 +1274,46 @@ def allowed_issuance_for_domain(common_name, extensions):
...
+def send_certificate_expiration_metrics():
+    """
+    Iterate over each certificate and emit a metric for how many days until expiration.
+    """
+    success = failure = 0
+
+    certificates = database.db.session.query(Certificate).filter(
+        Certificate.expired == false(),
+        Certificate.revoked == false()
+    )
```
jtschladen, line 1288: Would it make sense to include a (perhaps optional) parameter limiting the query to a maximum expiry period (probably in days)? Iterating over all valid certs could potentially be a very large data set in a large Lemur deployment, and I'm guessing this metric might be most valuable for understanding certs expiring soon -- or is getting visibility into the full set of certs more the goal?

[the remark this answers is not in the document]
mitchcail: I updated this query to filter out already rotated certs in commit [c415dca](https://github.com/Netflix/lemur/pull/4084/commits/c415dcaf746e76365592de9446a8e3e768b5234c)

mitchcail: Ability to pass in an expiry window was added as part of commit [e31ae99](https://github.com/Netflix/lemur/pull/4084/commits/e31ae99463418aab1a7ddbf0c76308e2d7129dac). The default of None will query for all active certs, while passing in a value such as 90 will only query for active certs expiring over the next 90 days. As part of this commit I moved the query logic to a separate method to make testing that logic easier.

jtschladen (approved):
Thanks for the contribution @mitchcail !

PR #4102, opened 2022-06-17
compare actual SAN when determining duplicates

charhate:
San is a boolean property in cert model. This change compares actual DNS SANs for equality.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -1080,8 +1080,14 @@ def is_duplicate(matching_cert, compare_to):
     matching_destinations = [destination.label for destination in matching_cert.destinations]
     compare_to_destinations = [destination.label for destination in compare_to.destinations]
 
-    return (len(matching_destinations) == len(compare_to_destinations)
-            and set(matching_destinations) == set(compare_to_destinations))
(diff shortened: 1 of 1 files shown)


lemur/certificates/cli.py
```diff
@@ -1080,8 +1080,12 @@ def is_duplicate(matching_cert, compare_to):
...
     compare_to_destinations = [destination.label for destination in compare_to.destinations]
 
-    return (len(matching_destinations) == len(compare_to_destinations)
-            and set(matching_destinations) == set(compare_to_destinations))
+    if (len(matching_destinations) == len(compare_to_destinations)
+            and set(matching_destinations) == set(compare_to_destinations)):
+        matching_sans = [domain.name for domain in matching_cert.domains]
+        compare_to_sans = [domain.name for domain in compare_to.domains]
+
+        return len(matching_sans) == len(compare_to_sans) and set(matching_sans) == set(compare_to_sans)
```
hosseinsh, line 1088: should we also add an else case and `return False`

hosseinsh (approved):

lemur/certificates/cli.py
charhate: Thanks for catching

PR #4174, opened 2022-08-23
De-activate endpoint (Entrust)

hosseinsh:
Entrust staging environment gives the option to deactivate provisioned certificates to regain used certificate tokens.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -1650,6 +1650,78 @@ class CertificateRevoke(AuthenticatedResource):
             return dict(message=f"Failed to revoke: {str(e)}"), 400
 
 
+class CertificateDeactivate(AuthenticatedResource):
+    def __init__(self):
+        self.reqparse = reqparse.RequestParser()
+        super(CertificateDeactivate, self).__init__()
+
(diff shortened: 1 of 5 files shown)


lemur/certificates/views.py
```diff
@@ -1650,6 +1650,78 @@ def put(self, certificate_id, data=None):
...
+           :reqheader Authorization: OAuth token to authenticate
+           :statuscode 200: no error
+           :statuscode 403: unauthenticated or cert attached to LB
+           :statuscode 400: encountered error, more details in error message
+
+        """
+        cert = service.get(certificate_id)
+
+        if not cert:
+            return dict(message="Cannot find specified certificate"), 404
```
charhate, line 1693: I would recommend using `412 = Precondition Failed` instead of 404, as it gets confusing whether the endpoint url is wrong giving 404 or an actual error thrown by API is giving 404

charhate (approved):

lemur/certificates/views.py
hosseinsh: shall we do this consistently across all endpoints checking for the certificate?

Issue #4233, opened 2022-10-19
OpenSSL PKCS12 export is prepopulating the Passphrase field


neilschelly:
When I go to export a certificate/key as a PKCS12 file, the passphrase prompt isn't empty. It's getting a default value equivalent to whatever the last user to export put in this field.
![image](https://user-images.githubusercontent.com/3343289/196695446-96cea564-c524-4d6b-8243-c69e270f59c7.png)

When the UI makes the request for export plugins, it receives the following response:
```
GET https://LEMUR/api/1/plugins?type=export
{
  "total": 4,
  "items": [
    <SNIP>
    {
      "slug": "openssl-export",
      "pluginOptions": [
        {
          "name": "type",
          "type": "select",
          "required": true,
          "available": [
            "PKCS12 (.p12)"
          ],
          "helpMessage": "Choose the format you wish to export",
          "value": "PKCS12 (.p12)"
        },
        {
          "name": "passphrase",
          "type": "str",
          "required": false,
          "helpMessage": "If no passphrase is given one will be generated for you, we highly recommend this.",
          "validation": "",
          "value": "PREPOPULATED_FIELD"
        },
        {
          "name": "alias",
          "type": "str",
          "required": false,
          "helpMessage": "Enter the alias you wish to use for the keystore.",
          "value": ""
        }
      ],
      "description": "Is a loose interface to openssl and support various formats",
      "title": "OpenSSL"
    }
  ]
}
```

PR #4234, opened 2022-10-20
Clear passphrase "memory" in the export plugin options

neilschelly:
Resolves https://github.com/Netflix/lemur/issues/4233

After an export operation is done, if any of the plugin options fields have the passphrase as their value, the value will be deleted. Since the plugin object will stay in use for future exports, that new "default" for the option will stick around for the next use otherwise.

I considered putting similar functionality in the actual openssl plugin where I saw the behavior, but it makes sense to me that this should exist in the certificate export functionality. Any plugin that stores the passphrase in any of its options will now have that cleared before the next use of the plugin.

One side effect of this... I can't be sure that all plugins will label the option holding a passphrase the same if they have one. For that reason, I didn't just look for an option called `passphrase` to clear. That would have worked for the openssl plugin at least. I opted to look for any options where the stored option value is equal to the passphrase returned above by the call to `plugin.export()`. That does mean that if I submit the passphrase for several of the fields, they will all have their values cleared out. I feel like that's probably a good thing if folks are setting all the options to the same thing as the passphrase.

### DIFF
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -1555,6 +1555,11 @@ class CertificateExport(AuthenticatedResource):
             cert.body, cert.chain, cert.private_key, options
         )
 
+        # Clear memory for last passphrase if it's in plugin.options
+        for option in plugin.options:
+            if 'value' in option and option['value'] == passphrase:
+                del option['value']
+
         # we take a hit in message size when b64 encoding
         return dict(
             extension=extension,


hosseinsh:
thanks @neilschelly for the thoughtful PR. 

looks good to me. going to merge it shortly.

hosseinsh (approved):

Issue #4259, opened 2022-11-22
Feature Request - being able to control the subject of email notifications


neilschelly:
Currently, the email subjects received from Lemur are basically hard-set to "Lemur: Template Name Notification" where `templates/template_name.html` becomes `Template Name` in the above subject. For our purposes, it would be very useful to change those subjects. We can override the actual _contents_ of the emails by specifying our own template files, but the subjects are essentially hardcoded in the lemur_email plugin.

I propose that the subject of the message be set by the `<title>SUBJECT</title>` found inside the template, if present. If that's not present, we can default to the current behavior. I will follow up with a PR soon.

PR #4260, opened 2022-11-22
Custom email subjects

neilschelly:
This is my proposal for solving my feature request at https://github.com/Netflix/lemur/issues/4259. If there's a <title> tag found in the email notification template, then that will be used for the email subject. Failing that, the current default subjects will apply.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,10 +115,10 @@ endif
 	@echo "--> Updating Python requirements"
 	pip install --upgrade pip
 	pip install --upgrade pip-tools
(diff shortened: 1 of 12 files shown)


hosseinsh:
thank you @neilschelly for raising the PR, and also documenting the issue. 
The PR looks good to me. 

Though for the sake of completeness, do you think you might be able to add test cases for the title parser, for example here:
https://github.com/Netflix/lemur/blob/a4719cb200ff89457f0372df52d3daa4181de29e/lemur/plugins/lemur_email/tests/test_email.py

PR #4322, opened 2023-02-10
Implement clean for S3DestinationPlugin, add confirm on dest remove

jmcrawford45:


CHANGELOG.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+Unreleased
```
hosseinsh, line 4: good practice to keep adding to the list vs. our current approach of checking manually for notable updates :D

lemur/plugins/lemur_aws/plugin.py
```diff
@@ -669,6 +669,12 @@ def delete_acme_token(self, token_path, options, **kwargs):
                                                                "filename": filename})
         return response
 
+    def clean(self, certificate, options, **kwargs):
+        prefix = self.get_option("prefix", options)
+        s3.delete(bucket_name=self.get_option("bucket", options),
+                  prefixed_object_name=f"{prefix}/{certificate.name}.pem",
```
hosseinsh, line 675: we might need a better handling of prefix, since it may also contain a `/` at the end, or be empty ""

hosseinsh: we can also a dedicated test case here
https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_aws/tests/test_plugin.py#L14

jmcrawford45: > we might need a better handling of prefix, since it may also contain a `/` at the end, or be empty ""

os.path.join handles both of these cases, I'll replace this and other usages with it

jmcrawford45: > we can also a dedicated test case here https://github.com/Netflix/lemur/blob/master/lemur/plugins/lemur_aws/tests/test_plugin.py#L14

Thanks for the pointer, added!

hosseinsh (approved):

charhate (approved):

PR #4344, opened 2023-02-27
Lemur: Destination - fix S3 bucket path and add logs for successful upload

jmcrawford45:


lemur/plugins/lemur_aws/plugin.py
```diff
@@ -577,7 +577,9 @@ class S3DestinationPlugin(ExportDestinationPlugin):
             "name": "prefix",
             "type": "str",
             "required": False,
+            "validation": check_validation("[^/].*"),
             "helpMessage": "Must be a valid S3 object prefix!",
+            "default": "lemur"
```
hosseinsh, line 582: do we want to set the default prefix empty?

hosseinsh (approved):

lemur/plugins/lemur_aws/plugin.py
charhate: +1, empty is potentially safer and may align with current behavior, if code could handle None path anywhere.

charhate (approved):

lemur/plugins/lemur_aws/plugin.py
jmcrawford45: sure thing, the validation regex gets a bit more complex, but it's nothing too unmanageable

PR #4393, opened 2023-04-17
Replaced get_random_secret logic with more secure version

ikkebr:
In the `get_random_secret` function, the generated secret key is divided into four equal parts, each containing a specific set of characters (uppercase letters, special characters, lowercase letters, and digits). While this approach ensures that the secret key has a mix of character types, it might lead to lower entropy in some cases.

For example, if the length of the secret key is small, the rounding operation (`round(length / 4)`) could result in an uneven distribution of characters from different sets. This might decrease the entropy of the generated secret key, making it more predictable and less secure.

To improve entropy in this function, it is better to randomly pick characters from a combined set of uppercase letters, special characters, lowercase letters, and digits. This will ensure that each character has an equal probability of being chosen at each position in the secret key, leading to higher entropy and a more secure random secret string.

### DIFF
--- a/docker/src/lemur.conf.py
+++ b/docker/src/lemur.conf.py
@@ -18,10 +18,8 @@ debug = os.environ.get("DEBUG") == "True"
 
 
 def get_random_secret(length):
-    secret_key = ''.join(secrets.choice(string.ascii_uppercase) for x in range(round(length / 4)))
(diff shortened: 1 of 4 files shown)


hosseinsh (approved):
Thanks Henrique for the great PR.

PR #4408, opened 2023-05-05
Enable dependabot updates for GitHub Actions

jtschladen:
I noticed that our GitHub action for CodeQL scanning is out of date. I also noticed that Dependabot now supports [updating actions](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot), so I'd like to let Dependabot take care of updating CodeQL for us.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,18 @@
   version: 2
   updates:
-    - directory: "/"
+    - package-ecosystem: "pip"
+      directory: "/"
       schedule:
         interval: "weekly"
         day: "monday"
         time: "08:00"
         timezone: "America/Los_Angeles"
-      package-ecosystem: "pip"
-      reviewers:
-        - "hosseinsh"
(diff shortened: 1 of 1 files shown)


hosseinsh (approved):

hosseinsh (approved):

.github/dependabot.yml
```diff
@@ -1,16 +1,18 @@
...
-    - directory: "/"
+    - package-ecosystem: "pip"
+      directory: "/"
       schedule:
         interval: "weekly"
         day: "monday"
         time: "08:00"
         timezone: "America/Los_Angeles"
-      package-ecosystem: "pip"
-      reviewers:
```
jtschladen, line 10: I don't believe reviewers should be necessary, as we have [CODEOWNERS](https://github.com/Netflix/lemur/blob/master/.github/CODEOWNERS) set up.

PR #4459, opened 2023-06-07
Add type checking to lemur CI

jmcrawford45:


lemur/common/utils.py
```diff
@@ -504,7 +504,7 @@ def drop_last_cert_from_chain(full_chain: str) -> str:
     pem_certificate = OpenSSL.crypto.dump_certificate(
         OpenSSL.crypto.FILETYPE_PEM,
         OpenSSL.crypto.load_certificate(
-            OpenSSL.crypto.FILETYPE_PEM, ''.join(cert.decode() for cert in full_chain_certs[:-1])
+            OpenSSL.crypto.FILETYPE_PEM, ''.join(cert.decode() for cert in full_chain_certs[:-1]).encode()
```
hosseinsh, line 507: That is the only part I would appreciate additional clarification or context on.

hosseinsh (approved):

lemur/common/utils.py
jmcrawford45: `load_certificate` takes a byte string. Before we were passing a unicode string, but that hasn't caused any issues since PEM only uses ASCII characters. `encode` converts a str to a bytes in python so the type matches

PR #4467, opened 2023-06-13
Deduplication logic decoupled from cert naming convention

charhate:
Certificated name could be different than the default naming done by Lemur, which is roughly `CN-Issuer-ValidFrom-ValidTo-SerialNumber`. Typically duplicate certs share same prefix and differ in their names by using serial number. However relying on the naming in deduping of the certs may not always work.

This code change groups the candidate certs by common name, instead of relying on shared prefix (which included common name anyway). The rest of the logic remains unchanged.

### DIFF
--- a/lemur/certificates/cli.py
+++ b/lemur/certificates/cli.py
@@ -31,8 +31,8 @@ from lemur.certificates.service import (
     get_all_certs_attached_to_endpoint_without_autorotate,
     get_all_certs_attached_to_destination_without_autorotate,
     revoke as revoke_certificate,
-    list_duplicate_certs_by_authority,
-    get_certificates_with_same_prefix_with_rotate_on,
+    list_recent_valid_certs_issued_by_authority,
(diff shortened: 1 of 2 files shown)


lemur/certificates/cli.py
```diff
@@ -1082,6 +1080,7 @@ def is_duplicate(matching_cert, compare_to):
         or matching_cert.key_type != compare_to.key_type
         or matching_cert.not_before.date() != compare_to.not_before.date()
         or matching_cert.not_after.date() != compare_to.not_after.date()
+        or matching_cert.authority_id != compare_to.authority_id
```
hosseinsh, line 1083: shall we also include comparison on destiations?

hosseinsh (approved):

lemur/certificates/cli.py
charhate: Yup, destination and SAN comparison is further below in this same method

PR #4492, opened 2023-06-30
Rename click commands to allow for programmatic calls

jmcrawford45:


jmcrawford45:
tl;dr is that click commands are expect to be called without args, unlike our previous cli manager. By creating a `func_name_command` alias for the click command, we can continue calling the underlying functions programmatically without unexpected extra argument errors

jtschladen (approved):

PR #4507, opened 2023-07-12
Add listener certification for SNI

charhate:
If cert is used as Listener cert for SNI, lemur cert rotation does not configure new cert the same way. Along with `modify_listener`, it needs to do `add_listener_certificates`. 
Currently this code relies on `describe_listener_certificates` to see if there is a cert added with default value as both - true and false.

### DIFF
--- a/lemur/plugins/lemur_aws/elb.py
+++ b/lemur/plugins/lemur_aws/elb.py
@@ -6,6 +6,7 @@
 .. moduleauthor:: Kevin Glisson <kglisson@netflix.com>
 """
 import botocore
+import click
 from flask import current_app
 
 from retrying import retry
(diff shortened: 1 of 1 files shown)


lemur/plugins/lemur_aws/elb.py
```diff
@@ -421,11 +422,67 @@ def attach_certificate_v2(listener_arn, port, certificates, **kwargs):
...
     try:
-        return kwargs["client"].modify_listener(
+        needs_cert_update_for_sni = has_listener_cert_for_sni(listener_arn, kwargs["client"])
+
+        modified_listener = kwargs["client"].modify_listener(
             ListenerArn=listener_arn, Port=port, Certificates=certificates
         )
+
+        if needs_cert_update_for_sni:
+            click.echo("[+] Adding cert as listener cert for SNI")
```
jmcrawford45, line 432: should we be using current_app.logger instead? it seems that's mostly what we use today for plugins and click.echo is mainly for cli commands

lemur/plugins/lemur_aws/elb.py
```diff
@@ -421,11 +422,67 @@ def attach_certificate_v2(listener_arn, port, certificates, **kwargs):
...
     """
     try:
-        return kwargs["client"].modify_listener(
+        needs_cert_update_for_sni = has_listener_cert_for_sni(listener_arn, kwargs["client"])
+
+        modified_listener = kwargs["client"].modify_listener(
             ListenerArn=listener_arn, Port=port, Certificates=certificates
         )
+
+        if needs_cert_update_for_sni:
```
jmcrawford45, line 431: Can we add a test case for this, or do we already have a case internally that we can verify post-deployment?

jmcrawford45 (approved):
modulo open comments

jtschladen (approved):

lemur/plugins/lemur_aws/elb.py
charhate: Though seemed straightforward, [describe_listener_certificates](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2/client/describe_listener_certificates.html) output for mocked client is missing the `IsDefault` flag which should've been present as per the docs. Implementation depends on the flag being present.
The code is tested manually rotating cert for an LB. And we do have a case to internally verify post-deployment

PR #4509, opened 2023-07-14
Add admin-only authority creation option

jtschladen:
Adds `ADMIN_ONLY_AUTHORITY_CREATION` to allow Lemur to be configured to expose authority creations to admins only.


hosseinsh (approved):

lemur/authorities/views.py
```diff
@@ -227,6 +228,11 @@ def post(self, data=None):
            :statuscode 403: unauthenticated
```
charhate, line 228: nit, this can use update too, calling out 401 as unauthenticated and 403 as unauthorized/forbidden

charhate (approved):

lemur/authorities/views.py
jtschladen: Thanks, good point. There are 53 other places in the code that define 403 as unauthenticated, which is never true as far as I am aware 😬 I won't touch those now, though.

Issue #4518, opened 2023-07-18
Access Control Vulnerabilities in the Read-Only Role


joeinfor7685:
Currently, there is an issue where users assigned the read-only role have the ability to create domains and import certificates, which goes against the intended access restrictions. This presents a security vulnerability as unauthorized actions can be performed by users who are not supposed to have such privileges. The issues are in the following sections:

**Sections:**
- Admin -> Domains
- Certificates -> Import

This situation poses the following risks:

**Unauthorized Domain Creation:** 
Users with read-only access should not have the capability to create domains. Allowing this action could result in the creation of unauthorized domains that may be used for malicious purposes or cause confusion within the system.

**Unauthorized Certificate Imports:**
Importing certificates should be restricted to users with the appropriate permissions and responsibilities. Allowing read-only users to import certificates opens up the potential for unauthorized or unverified certificates to be added to the system, compromising its security and integrity.

To address this issue, it is crucial to rectify the access control configuration to align with the intended roles and permissions:

**Restrict Domain Creation:** 
Modify the access control settings to ensure that only users with the necessary permissions, such as administrators or specific roles designated for domain creation, can create domains within the system. This prevents unauthorized users from creating domains.

**Control Certificate Imports:** 
Limit the privilege to import certificates to users who have been explicitly granted the appropriate permissions. This ensures that only authorized users, such as certificate administrators, can import valid and trusted certificates into the system.

By resolving these access control flaws, the system can effectively enforce proper restrictions, reducing the risk of unauthorized actions and maintaining the integrity and security of the application.

**References:**
- [OWASP C7: Enforce Access Controls](https://owasp.org/www-project-proactive-controls/v3/en/c7-enforce-access-controls)
- [OWASP's Authorization Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authorization_Cheat_Sheet.html)
- [CWE-284](https://cwe.mitre.org/data/definitions/284.html)

---

The referenced issue was found via a pen test conducted in collaboration with [Infor](https://www.infor.com) and [Cobalt.io](https://www.cobalt.io)

PR #4538, opened 2023-08-02
Access Control Vulnerabilities - Issue #4518

joeinfor7685:
Update to address Issue #4518. 

Currently users assigned the `read-only` role are still able to create/modify resources. 

The `LEMUR_STRICT_ROLE_ENFORCEMENT` config param has been added, along with a new `StrictRolePermission` permission. 

When `LEMUR_STRICT_ROLE_ENFORCEMENT` is set to True, this property enforces the default Lemur role functionality. The default Lemur roles are `admin`, `operator`, and `read-only`. If this is set to True, users are required to have a role assigned upon creation and the `operator` and `read-only` role are strictly enforced. Users assigned the `read-only` role will not be able to create/update resources.

### DIFF
--- a/lemur/authorities/views.py
+++ b/lemur/authorities/views.py
@@ -14,6 +14,7 @@ from lemur.common.schema import validate_schema
 from lemur.auth.service import AuthenticatedResource
 from lemur.auth.permissions import AuthorityCreatorPermission
 from lemur.auth.permissions import AuthorityPermission
+from lemur.auth.permissions import operator_permission
 
 from lemur.certificates import service as certificate_service
 
@@ -118,6 +119,7 @@ class AuthoritiesList(AuthenticatedResource):
         args["user"] = g.current_user
         return service.render(args)
 
+    @operator_permission.require(http_exception=403)
     @validate_schema(authority_input_schema, authority_output_schema)
     def post(self, data=None):
         """
@@ -299,6 +301,7 @@ class Authorities(AuthenticatedResource):
         """
         return service.get(authority_id)
 
+    @operator_permission.require(http_exception=403)
     @validate_schema(authority_update_schema, authority_output_schema)
     def put(self, authority_id, data=None):
         """
--- a/lemur/authorities/views.py
+++ b/lemur/authorities/views.py
@@ -119,8 +119,8 @@ class AuthoritiesList(AuthenticatedResource):
         args["user"] = g.current_user
         return service.render(args)
 
-    @operator_permission.require(http_exception=403)
     @validate_schema(authority_input_schema, authority_output_schema)
+    @operator_permission.require(http_exception=403)
     def post(self, data=None):
         """
         .. http:post:: /authorities
@@ -301,8 +301,8 @@ class Authorities(AuthenticatedResource):
         """
         return service.get(authority_id)
 
-    @operator_permission.require(http_exception=403)
     @validate_schema(authority_update_schema, authority_output_schema)
+    @operator_permission.require(http_exception=403)
     def put(self, authority_id, data=None):
         """
         .. http:put:: /authorities/1
--- a/lemur/authorities/views.py
+++ b/lemur/authorities/views.py
@@ -12,9 +12,7 @@ from lemur.common import validators
 from lemur.common.utils import paginated_parser
 from lemur.common.schema import validate_schema
 from lemur.auth.service import AuthenticatedResource
-from lemur.auth.permissions import AuthorityCreatorPermission
-from lemur.auth.permissions import AuthorityPermission
-from lemur.auth.permissions import operator_permission
+from lemur.auth.permissions import AuthorityCreatorPermission, AuthorityPermission, StrictRolePermission
 
 from lemur.certificates import service as certificate_service
 
@@ -230,9 +228,8 @@ class AuthoritiesList(AuthenticatedResource):
            :statuscode 403: unauthorized
            :statuscode 200: no error
         """
-
         permission = AuthorityCreatorPermission()
-        if not permission.can():
+        if not permission.can() or not StrictRolePermission().can():
             return dict(message="You are not allowed to create a new authority."), 403
 
         if not validators.is_valid_owner(data["owner"]):
@@ -414,16 +411,16 @@ class Authorities(AuthenticatedResource):
         roles = [x.name for x in authority.roles]
         permission = AuthorityPermission(authority_id, roles)
 
-        if permission.can():
-            return service.update(
-                authority_id,
-                owner=data["owner"],
-                description=data["description"],
-                active=data["active"],
-                roles=data["roles"],
-            )
+        if not permission.can() or not StrictRolePermission().can():
+            return dict(message="You are not authorized to update this authority."), 403
 
-        return dict(message="You are not authorized to update this authority."), 403
+        return service.update(
+            authority_id,
+            owner=data["owner"],
+            description=data["description"],
+            active=data["active"],
+            roles=data["roles"],
+        )
 
 
 class CertificateAuthority(AuthenticatedResource):
--- a/lemur/authorities/views.py
+++ b/lemur/authorities/views.py
@@ -120,7 +120,6 @@ class AuthoritiesList(AuthenticatedResource):
         return service.render(args)
 
     @validate_schema(authority_input_schema, authority_output_schema)
-    @operator_permission.require(http_exception=403)
     def post(self, data=None):
         """
         .. http:post:: /authorities
@@ -302,7 +301,6 @@ class Authorities(AuthenticatedResource):
         return service.get(authority_id)
 
     @validate_schema(authority_update_schema, authority_output_schema)
-    @operator_permission.require(http_exception=403)
     def put(self, authority_id, data=None):
         """
         .. http:put:: /authorities/1
--- a/lemur/certificates/views.py
+++ b/lemur/certificates/views.py
@@ -20,7 +20,7 @@ from lemur.common.schema import validate_schema
 from lemur.common.utils import paginated_parser
 
 from lemur.auth.service import AuthenticatedResource
-from lemur.auth.permissions import AuthorityPermission, CertificatePermission, operator_permission
+from lemur.auth.permissions import AuthorityPermission, CertificatePermission, StrictRolePermission
 
 from lemur.certificates import service
 from lemur.certificates.models import Certificate
@@ -388,7 +388,6 @@ class CertificatesList(AuthenticatedResource):
         return service.render(args)
 
     @validate_schema(certificate_input_schema, certificate_output_schema)
-    @operator_permission.require(http_exception=403)
     def post(self, data=None):
         """
         .. http:post:: /certificates
@@ -504,6 +503,9 @@ class CertificatesList(AuthenticatedResource):
            :statuscode 403: unauthenticated
 
         """
+        if not StrictRolePermission().can():
+            return dict(message="You are not authorized to create a new certificate."), 403
+
         if not validators.is_valid_owner(data["owner"]):
             return dict(message=f"Invalid owner: check if {data['owner']} is a valid group email. Individuals cannot be certificate owners."), 412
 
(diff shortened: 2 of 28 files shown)


lemur/authorities/views.py
```diff
@@ -119,6 +120,7 @@ def get(self):
         return service.render(args)
 
     @validate_schema(authority_input_schema, authority_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 123: this is currently controlled via the ADMIN_ONLY_AUTHORITY_CREATION property, so this isn't needed

lemur/authorities/views.py
```diff
@@ -300,6 +302,7 @@ def get(self, authority_id):
         return service.get(authority_id)
 
     @validate_schema(authority_update_schema, authority_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 305: this already requires role admin so it shouldn't be necessary https://github.com/Netflix/lemur/blob/1c096f1028a237a6f58d181e5aaabb0f8f1e0953/lemur/auth/permissions.py#L80

lemur/certificates/views.py
```diff
@@ -542,6 +543,7 @@ def __init__(self):
         super(CertificatesUpload, self).__init__()
 
     @validate_schema(certificate_upload_input_schema, certificate_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 546: this restriction doesn't make much sense to me. I don't understand the risk of a user uploading a cert they already own

lemur/certificates/views.py
```diff
@@ -388,6 +388,7 @@ def get(self):
         return service.render(args)
 
     @validate_schema(certificate_input_schema, certificate_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 391: for some of our users, the self service nature of lemur is an important feature. We'd likely want this gated with a documented configuration property e.g, ADMIN_ONLY_CERTIFICATE_CREATION

lemur/certificates/views.py
```diff
@@ -817,6 +819,7 @@ def get(self, certificate_id):
         return service.get(certificate_id)
 
     @validate_schema(certificate_edit_input_schema, certificate_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 822: this restriction doesn't make much sense to me. I don't understand the risk of a user updating notifications / ownership of a cert they already own

lemur/certificates/views.py
```diff
@@ -968,6 +971,7 @@ def put(self, certificate_id, data=None):
         return cert
 
     @validate_schema(certificate_edit_input_schema, certificate_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 974: ibid

lemur/certificates/views.py
```diff
@@ -1074,6 +1078,7 @@ def post(self, certificate_id, data=None):
         log_service.create(g.current_user, "update_cert", certificate=cert)
         return cert
 
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 1081: we have ALLOW_CERT_DELETION config property to disallow deletion across the board, not sure if there's a case where we want to prevent owners from deleting their certs, but we'd want operators to have that ability

lemur/certificates/views.py
```diff
@@ -1133,6 +1138,7 @@ def __init__(self):
         super(CertificateUpdateOwner, self).__init__()
 
     @validate_schema(certificate_edit_input_schema, certificate_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 1141: same comment as update endpoint w.r.t. owners being able to update metadata for their certs

lemur/certificates/views.py
```diff
@@ -1456,6 +1462,7 @@ def __init__(self):
         super(CertificateExport, self).__init__()
 
     @validate_schema(certificate_export_input_schema, None)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 1465: Post is deceptive here, an export is functionally a get of a cert that's reformatted

lemur/certificates/views.py
```diff
@@ -1583,6 +1590,7 @@ def __init__(self):
         super(CertificateRevoke, self).__init__()
 
     @validate_schema(certificate_revoke_schema, None)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 1593: same comment as above w.r.t. owners being able to manage certificates they own

lemur/certificates/views.py
```diff
@@ -1669,6 +1677,7 @@ def __init__(self):
         self.reqparse = reqparse.RequestParser()
         super(CertificateDeactivate, self).__init__()
 
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 1680: ibid

lemur/common/schema.py
```diff
@@ -159,7 +160,10 @@ def decorated_function(*args, **kwargs):
             except Exception as e:
                 capture_exception()
                 current_app.logger.exception(e)
-                return dict(message=str(e)), 500
+                code = 500
```
jmcrawford45, line 163: We may be able to avoid this change by switching the order of validate_schema and permission require decorators. I'm concerned that some users may implement schema validation that relies on a remote service in which case it would be appropriate to report as a 500 even if remote returned a 4xx

lemur/domains/views.py
```diff
@@ -172,6 +173,7 @@ def get(self, domain_id):
         return service.get(domain_id)
 
     @validate_schema(domain_input_schema, domain_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 176: this is gated by admin already https://github.com/Netflix/lemur/blob/1c096f1028a237a6f58d181e5aaabb0f8f1e0953/lemur/auth/permissions.py#L25

lemur/domains/views.py
```diff
@@ -86,6 +86,7 @@ def get(self):
         return service.render(args)
 
     @validate_schema(domain_input_schema, domain_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 89: for some of our users, the self service nature of lemur is an important feature. We'd likely want this gated with a documented configuration property e.g, ADMIN_ONLY_DOMAIN_CREATION

lemur/roles/views.py
```diff
@@ -257,6 +257,7 @@ def get(self, role_id):
         )
 
     @validate_schema(role_input_schema, role_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 260: this is gated by admin already https://github.com/Netflix/lemur/blob/1c096f1028a237a6f58d181e5aaabb0f8f1e0953/lemur/auth/permissions.py#L58

lemur/notifications/views.py
```diff
@@ -113,6 +114,7 @@ def get(self):
         return service.render(args)
 
     @validate_schema(notification_input_schema, notification_output_schema)
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 117: this resource is actually the most concerning to me that we don't have any access controls on. But, even for this case we'd likely want to start with a default false config property ADMIN_ONLY_NOTIFICATION_WRITE with the eventual goal of defaulting to True

lemur/pending_certificates/views.py
```diff
@@ -213,6 +213,7 @@ def get(self, pending_certificate_id):
     @validate_schema(
         pending_certificate_edit_input_schema, pending_certificate_output_schema
     )
+    @operator_permission.require(http_exception=403)
```
jmcrawford45, line 216: we can gate put with ADMIN_ONLY_CERTIFICATE_CREATION property, I don't see a strong reason for delete / post to be gated though

jmcrawford45 (commented):
Thank you @joeinfor7685 for your contribution! My apologies that it took a while to review this; although the diff is small, it has pretty large implications on the security and usability of lemur. I've flagged a few places where the admin check is redundant, where we'd want these enforcements to be configurable, and where additional access controls don't seem necessary. Overall, introducing a few new configuration properties will likely be the best way to allow for different consumers to make informed decisions on the tradeoffs between more fine grained access controls and self service models.

lemur/notifications/views.py
jmcrawford45: another alternative that allows for self service management of notifications while preventing some of the attacks I'm concerned about is to introduce the concept of notification ownership (similar to what we have for roles, authorities, and certificates today). but this is likely to be a somewhat involved change

lemur/certificates/views.py
joeinfor7685: we are currently not enforcing the default `read-only` role. a user assigned the `read-only` role is still able to upload certificates. we are thinking if an admin creates a new user with the `read-only` role. they shouldn't be able to create/upload any certificates just read them, unless they are promoted to the `operator` or `admin` role.

also, if a user was initially assigned the `operator` role, then changed to the `read-only` role. this would ensure the user would not be able to create/upload any certificates.

lemur/certificates/views.py
joeinfor7685: same as Line 546

lemur/certificates/views.py
joeinfor7685: i agree with having a new configuration property. i noticed that when users are initially created a role is not required. so the current updates would restrict users that are not assigned a role.

for the new configuration property, this would allow the `operator` and `admin` role to access the endpoints and enforce the `read-only` role. instead of `ADMIN_ONLY_*`, what if we added a property called `LEMUR_STRICT_ROLE_ENFORCEMENT=(True|False)` (or something similar). 

if this is set it will activate/enforce the default system roles and require admins to assign a role when creating users. it will default to `False` to ensure backward compatibly.

joeinfor7685:
Added a new `LEMUR_STRICT_ROLE_ENFORCEMENT` config param, along with a new `StrictRolePermission` permission.

When `LEMUR_STRICT_ROLE_ENFORCEMENT` is set to `True`, it enforces the default Lemur role functionality for `admin`, `operator`, and `read-only`. Users are required to have a default role assigned upon creating/updating and the `operator` and `read-only` role are strictly enforced. Users assigned the `read-only` role will not be able to create/update resources.

The `LEMUR_STRICT_ROLE_ENFORCEMENT` defaults to `False` and if not set everything will be backwards compatible.

joeinfor7685:
Also, removed all of the checks where  the`allow creators` permission is being used. i agree, these checks don't make much sense and the certificate creator should be able to modify the certificate.

lemur/users/service.py
```diff
@@ -14,6 +14,9 @@
 from lemur.users.models import User
 
 
+DEFAULT_ROLES = ["admin", "operator", "read-only"]
```
jmcrawford45, line 17:
```suggestion
STRICT_ENFORCEMENT_DEFAULT_ROLES = ["admin", "operator", "read-only"]
```
to reduce confusion for consumers that haven't configured strict enforcement

lemur/users/service.py
```diff
@@ -26,6 +29,10 @@ def create(username, password, email, active, profile_picture, roles):
     :param roles:
     :return:
     """
+    strict_role_enforcement = current_app.config.get("LEMUR_STRICT_ROLE_ENFORCEMENT", False)
+    if strict_role_enforcement and not any(role.name in DEFAULT_ROLES for role in roles):
+        return dict(message="Default role required: admin, operator, read-only"), 400
```
jmcrawford45, line 34: nit: may be worth rewording to clarify that at least one are required (not necessarily all)

jmcrawford45:
Thanks again for working on this! Dropped a couple minor nit comments, but once those are resolved and the build is green we should be good to merge

joeinfor7685:
Resolved remaining comments and build should be good to go. Let us know if you see anything else. Thanks!

PR #4554, opened 2023-08-14
Using newer versions of Python and nodejs to fix Docker build failures

Tamerz:
Running `docker compose run test` would fail. One of the required Python libraries is `celery[redis]==5.3.1` which is not available on Python 3.7 or older.

The `7.x` repository from NodeSource does not work with newer versions of Debian that the `python` containers are based on. I also had to install `nodejs` instead of just `npm` or it would fail. The `nodejs` package also installs `npm`.

I chose Python 3.8 since it is the lowest working version that supports that version of celery and Node.js 18 since it is the latest LTS release, and the tests pass when using it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM python:3.7
+FROM python:3.8
 RUN apt-get update
 RUN apt-get install -y make software-properties-common curl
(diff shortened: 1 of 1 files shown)


hosseinsh (approved):
thanks @Tamerz for the PR. this aligns the docker config with the Lemur requirements in terms of min. Python version.

PR #4600, opened 2023-09-14
Handle token exchange failure

Ebag333:
Small PR to catch a case where token exchange can fail, this provides some error catching/handling around either the request directly failing or the ID/token missing (previously would just throw a generic KeyError if one of these were missing).


Ebag333:
@hosseinsh Fixed the linting error if you want to rerun the tests.

hosseinsh (approved):

PR #4609, opened 2023-09-20
Handle a missing field

Ebag333:
Add a bit of error handling if we get thrown a key error. An example of this happening is in a new environment when populating Authorities if you missed the serial number or some sort of validity range it'd throw a generic error (no meaningful indification as to why it would fail to create), this way we got told specifically which field we're having issues with.


Ebag333:
Closing as this was started from the wrong branch and is a bit messy.

PR #4610, opened 2023-09-20
Requests Logging

Ebag333:
Add in functionality to log requests that come in.  This is useful both for debugging purposes as well as if we want to be able to audit usage logs.

We sanitize the URLs that touch /api/ by default to avoid exposing any sort of query parameters that could contain sensitive information.  This can be disabled for debugging purposes (not recommended in a prod env).

We also skip /metrics and /healthcheck, as these tend to be noisy and don't really do much for logging/debugging purposes.  This is set in an app config variable so that we can add endpoints to skip or otherwise modify what we're logging.

We log this at info level it can be turned off simply by increasing the log level.

This functionality defaults to off to avoid spamming existing folks who don't want this level of logging turned on.

### DIFF
--- a/lemur/__init__.py
+++ b/lemur/__init__.py
@@ -9,7 +9,9 @@
 .. moduleauthor:: Hossein Shafagh <hshafagh@netflix.com>
 
 """
+import socket
 import time
+import urllib.parse
 from flask import g, request
 
 from lemur import factory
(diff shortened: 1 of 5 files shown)


Ebag333:
Closing as this is a bit messy (started with the wrong branch)

PR #4611, opened 2023-09-20
Requests Logging

Ebag333:
Add in functionality to log requests that come in. This is useful both for debugging purposes as well as if we want to be able to audit usage logs.

We sanitize the URLs that touch /api/ by default to avoid exposing any sort of query parameters that could contain sensitive information. This can be disabled for debugging purposes (not recommended in a prod env).

We also skip /metrics and /healthcheck, as these tend to be noisy and don't really do much for logging/debugging purposes. This is set in an app config variable so that we can add endpoints to skip or otherwise modify what we're logging.

We log this at info level it can be turned off simply by increasing the log level.

This functionality defaults to off to avoid spamming existing folks who don't want this level of logging turned on.


lemur/manage.py
```diff
@@ -126,7 +126,9 @@ def cli(script_info, config):
 LOG_LEVEL = "DEBUG"
 LOG_FILE = "lemur.log"
 LOG_UPGRADE_FILE = "db_upgrade.log"
-
+LOG_REQUEST_HEADERS = False
+LOG_SANITIZE_REQUEST_HEADERS = True
+LOG_REQUEST_HEADERS_SKIP_ENDPOINT = ["/metrics", "/healthcheck"]  # These endpoints are noisy so skip them by default
```
charhate, line 131: Can you please document these configs under Lemur Docs https://lemur.readthedocs.io/en/latest/administration.html

charhate (approved):

lemur/manage.py
Ebag333: @charhate great call out, thanks for catching that. I've updated the docs, let me know if any need more clarification or additional verbage.

PR #4617, opened 2023-09-22
Mapping of IDP roles against Lemur roles

Ebag333:
Functionally this does two things:
1) Takes a fixed mapping in our config and any users that have a group we give that specified Lemur role (net new functionality)
2) Takes all the groups assigned to a user (profile['roles'] or profile['googleGroups']) and creates a Lemur group if it doesn't already exist, then assigns it to the user (reworked from existing functionality)

Overall the goal of this is to allow for configuration from various IDPs. For example in our environment we have `roles` that we need to pull from and not `googleGroups`.  If you have multiple IDPs (or just have multiple places to look for assignments), you can chain any keys found under `profile` that has a list of strings.

Major changes:
- dropped hard setting the third party flag. If it's being created then it's already set and redundant, if it's not set than there's a risk that we're overwriting something with a third party flag that we presumably shouldn't.
- added a gate to the role creation per user, so if someone wants to disable creating a role for each user they have that choice.  It defaults to on to mimic the existing behavior, but I can see use cases as creating a role per user is quite noisy.
- Added protecting the built-ins.  The issue is that if someone creates a group named `admin` in the IDP (which is not only simple to do but common) then all those users get added as Lemur admins. This is more an issue of accidental over permissioning than malicious (as if your IDP is compromised you're screwed anyway).  This defaults to enabled, which is a semi-breaking change from previous.  This will _only_ be an issue for someone who has that particular use case (IE: my group in Google/LDAP is named "admin" and I want that same user set to be admin in Lemur).  Mapping the users via IDP_ROLES_MAPPING does NOT get stopped by the protect builtins, to give a way to still assign those specific groups should that behavior be desired (eg IDP_ROLES_MAPPING = {"admin": "admin"}). 

While there is a lot of additional configuration to toggle bits on/off or control how they work, they _should_ all default to the existing behavior if not set. The one exception is the one mentioned above with protecting the built-ins.

Few other minor changes: 
- added some type hints
- dedupe the roles list at the end
- some comments/logging/documentation


lemur/auth/views.py
```diff
@@ -183,56 +183,95 @@ def retrieve_user_memberships(user_api_url, user_membership_provider, access_tok
...
+        idp_group_to_role_map = current_app.config.get("IDP_ROLES_MAPPING", {})
+        matched_roles = [
+            idp_group_to_role_map[role] for role in profile.get(idp_groups_key, []) if role in idp_group_to_role_map
+        ]
+        roles.extend(matched_roles)
+
+        # Automatically create and assign roles from the user profile.
+        if current_app.config.get("IDP_ASSIGN_ROLES_FROM_USER_GROUPS", True):
+            idp_roles_description = current_app.config.get("IDP_ROLES_DESCRIPTION",
+                                                           "Identity provider role (generated by Lemur)")
```
charhate, line 222: nit, perhaps helpful to include name of the IDP - `idp_groups_key` in default description here to make the source obvious. This would also get the description closer to earlier hardcoded one ` description="This is a google group based role created by Lemur",`

charhate (approved):

charhate:
Thanks for a thorough description and for carefully crafting the change. I would appreciate if you can fix the linting errors to get the built going and hopefully we can merge soon.
```
--> Linting Python files
PYFLAKES_NODOCTEST=1 flake8 lemur
lemur/manage.py:151:78: W291 trailing whitespace
lemur/manage.py:164:28: W291 trailing whitespace
make: *** [Makefile:96: lint-python] Error 1
Error: Process completed with exit code 2.
```

lemur/default.conf.py
```diff
@@ -16,3 +16,13 @@
 
 LOG_LEVEL = "DEBUG"
 LOG_FILE = "lemur.log"
+
+# Set of controls to use around ingesting user group information from the IDP
```
hosseinsh, line 20: It would also be great to document these new configurations in the Lemur docs for better visibility and understanding. Thanks.

Ebag333:
> Thanks for a thorough description and for carefully crafting the change. I would appreciate if you can fix the linting errors to get the built going and hopefully we can merge soon.
> 
> ```
> --> Linting Python files
> PYFLAKES_NODOCTEST=1 flake8 lemur
> lemur/manage.py:151:78: W291 trailing whitespace
> lemur/manage.py:164:28: W291 trailing whitespace
> make: *** [Makefile:96: lint-python] Error 1
> Error: Process completed with exit code 2.
> ```

Thanks for catching this.  I'd thought I'd gotten all the linting, but apparently not. :)

Ebag333:
I've addressed the three comments above, let me know if there's anything else you'd like to see.

Ebag333:
I imported typing hints (annotations) from __future__ that should make it compatible with Python 3.7 and 3.8, so that should take care of the stacktrace we're seeing on the 3.8 tests.

PR #4674, opened 2023-10-24
Add support for multiple versions of lemur token secret

jmcrawford45:


lemur/tests/conf.py
```diff
@@ -34,6 +34,7 @@ def get_random_secret(length):
 
 # You should consider storing these separately from your config (should be URL-safe)
 LEMUR_TOKEN_SECRET = "test"
+LEMUR_TOKEN_SECRETS = [LEMUR_TOKEN_SECRET]
```
jtschladen, line 37: For these files where we control the config, is there any reason to keep `LEMUR_TOKEN_SECRET`? No need for backwards compatibility here.

jtschladen (approved):

PR #4678, opened 2023-10-27
Update get_all_valid_certs service.py

Lounifitz:
Add "oder_by" so that the result is always in the same order. Currently, the query returns records in a random order, making pagination useless. Some certificates are not processed, others are processed several times. We currently have over 18,000 active certificates in our database and have identified and reproduced this problem.

### DIFF
--- a/lemur/certificates/service.py
+++ b/lemur/certificates/service.py
@@ -134,12 +134,12 @@ def get_all_valid_certs(authority_plugin_name, paginate=False, page=1, count=100
         query = query.outerjoin(Authority, Authority.id == Certificate.authority_id)\
             .filter(Certificate.not_after > arrow.now().format("YYYY-MM-DD"))\
             .filter(Authority.plugin_name.in_(authority_plugin_name))\
-            .filter(Certificate.revoked.is_(False))
-
+            .filter(Certificate.revoked.is_(False))\
+            .order_by(Certificate.id)
     else:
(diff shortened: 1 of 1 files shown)


hosseinsh (approved):

jtschladen:
Hi @Lounifitz ! Thanks for your contribution. #4141 was a prior change intended to fix the issue you describe around pagination; when pagination is enabled, the `sort_and_page` call on [line 148](https://github.com/Netflix/lemur/pull/4678/files#diff-6cdb5b288be1b6c86d89080b40e8d2398ce8f85c355c7393cbd6017f00744864R148) sorts prior to paging. I do agree that results returned with pagination disabled won't be sorted, though.

I'm curious if you're still running into the sorting issue with pagination enabled on a version of Lemur after #4141? If so, I'd like to fix the underlying issue in `sort_and_page`, as it affects other endpoints as well.

Lounifitz:
Hello,
Thank you for your answers.
Yes, it's my fault. We're still using version 1.1 for a few months, which doesn't include this correction, and I didn't pay any attention to this later correction.
We're going to integrate this correction immediately while waiting for the version upgrade scheduled for early 2024.
Sorry for the inconvenience.

hosseinsh:
No worries [Lounifitz](https://github.com/Lounifitz) and good like with the upgrade.

PR #4703, opened 2023-11-14
Add digicert source plugin

pinmarva:
Added Digicert source plugin


lemur/plugins/lemur_digicert/plugin.py
```diff
@@ -317,10 +319,71 @@ def __init__(self, *args, **kwargs):
...
+        # we Retry we also covers HTTP status code 406, 500, 502, 503, 504
+        retry_strategy = Retry(total=3, backoff_factor=0.1, status_forcelist=[406, 500, 502, 503, 504])
+        adapter = requests.adapters.HTTPAdapter(max_retries=retry_strategy)
+        self.session.mount("https://", adapter)
+
         super().__init__(*args, **kwargs)
 
-    def get_certificates(self):
-        pass
+    def get_certificates(self, options, **kwargs):
```
jmcrawford45, line 331: for backwards compatibility, can we gate this behavior via a config variable e.g. DIGICERT_SOURCE_ENABLED = False and document it in docs/administration.rst as well as adding a note to the changelog on this change?

lemur/plugins/lemur_digicert/plugin.py
```diff
@@ -317,10 +319,71 @@ def __init__(self, *args, **kwargs):
...
+                        base_url,
+                        c["certificate"]["id"],
+                        29
+                    )
+
+                    pem_all = self.session.get(download_url)
+
+                    certificates = x509.load_pem_x509_certificates(pem_all.content)
+                    certificate = certificates[0].public_bytes(serialization.Encoding.PEM).decode()
+                    chains = certificates[-(len(certificates) - 1):]
```
jmcrawford45, line 368:
```suggestion
                    chains = certificates[1:]
```

lemur/plugins/lemur_digicert/plugin.py
```diff
@@ -317,10 +319,71 @@ def __init__(self, *args, **kwargs):
...
-        pass
+    def get_certificates(self, options, **kwargs):
+        """Fetch all Digicert certificates."""
+        base_url = current_app.config.get("DIGICERT_URL")
+
+        # make request
+        search_url = f"{base_url}/services/v2/order/certificate"
+
+        certs = []
+        offset = 0
```
jmcrawford45, line 339: Can we use len(certs) instead of offset? I'm concerned that the API may return less than the limit. In this case, the source plugin would skip over some certs

lemur/plugins/lemur_digicert/plugin.py
pinmarva: Done!
- Added DIGICERT_SOURCE_ENABLED
- Added documentation
- Added information in CHANGELOG file

lemur/plugins/lemur_digicert/plugin.py
pinmarva: Done!

PR #4738, opened 2023-12-12
Allow filtering with SAN in certificates/valid endpoint

charhate:
Added supported to filter with a SAN. This is useful for the certs without CN, as CN could be optional.

Lookup could be extended to support lookup by multiple hostnames/SAN instead of just one. However, given the fact that domain table does not hold unique entries, performance will degrade when filtered with many SANs. Thus keeping it simple for now.

### DIFF
--- a/lemur/tests/test_certificates.py
+++ b/lemur/tests/test_certificates.py
@@ -1827,6 +1827,31 @@ def test_query_common_name(session):
     assert len(cn2_valid_certs) == 1
 
 
+def test_query_san(session):
+    from lemur.tests.factories import CertificateFactory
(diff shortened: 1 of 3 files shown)


lemur/tests/test_certificates.py
jtschladen, line 1: Any additional tests we can add to validate querying by SAN works as expected?

jtschladen (approved):

lemur/tests/test_certificates.py
charhate: Thanks, added https://github.com/Netflix/lemur/pull/4738/commits/8b944fdcdacf57bcde25119e7ba79605f1183923

PR #4803, opened 2024-03-28
fixed dependencies and small build issues

odopertchouk:
Numerous small fixes

- Fixed the problem where if a `hooks` directory doesn't exist in the project root, then setup-git step would create a symlink named `*`
- `lint-python` now scans the lemur package and not all python files in the repo, because that could also, for example, include installed python packages, if the virtual env directory is inside the repo directory.
- pip-compile now has verbose output, for ease of troubleshooting dependency issues.
- pip-compile steps reordered in the order of their dependency, otherwise they can cause spurious version conflicts.
- `celery[redis]>=5.3.5` (Instead of `== 5.3.5`) in requirements-tests.in to avoid version conflicts with requirements.txt, because  the latest verison of celery is now at least 5.3.6
- `moto[all] < 5` because lemur/plugins/lemur_aws/tests/ is not compatible with moto v5 . moto v5 no longer has mock_sts, mock_s3 and so on

Hi from Squarespace Product Security team!
We are avid users of Lemur and would like to contribute our fixes back to the community!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ reset-db:
 
 setup-git:
 	@echo "--> Installing git hooks"
-	if [ -d .git/hooks ]; then \
+	if [[ -d .git/hooks && -d hooks ]]; then \
 		git config branch.autosetuprebase always; \
 		cd .git/hooks && ln -sf ../../hooks/* ./; \
 	fi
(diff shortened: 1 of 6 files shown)


jtschladen (approved):
Thanks for the contribution!

PR #4804, opened 2024-03-28
Updated quickstart docs

odopertchouk:
- The documentation was missing the database initialization step, this would result in 
```
ImportError: Can't find Python file migrations/env.py
```

when running `lemur init`

- Added `make up-reqs`  before `make release` to ensure all dependencies are up-to-date when installing from source.


jtschladen (approved):
Thanks for the contribution @odopertchouk !

Issue #4764, opened 2024-01-19
Error message with CLI command and "Working outside of application context"


Lounifitz:
Good morning,
We have just performed the merge between your version 1.6.0 and our custom version 1.1.16 based on version 1.1.0 dating back 2 years.
We were also able to see the issues presented below on version 1.6.0 before the merge.

**First problem**:
When we call the "lemur --help" or "lemur" or "lemur somethingwrong" command we get the following error message:

![image](https://github.com/Netflix/lemur/assets/149179506/f8293d04-91ca-4514-8994-ac9b210b7d21)

I modified line 67 of the factory.py file as follows, which seems to fix the problem. Is this normal?

![image](https://github.com/Netflix/lemur/assets/149179506/25d85f71-d916-4b57-8815-b705ef836da9)

**Second problem**:
Although we have respected your way of declaring our functions within our plugins and yours in the "cli.py" files, we obtain the following error when executing the "lemur command" command: "**RuntimeError: Working outside of application context**".
For this to work I had to add an import of "with_appcontext" and add the line "@with_appcontext" before the "cli" function definition.

![image](https://github.com/Netflix/lemur/assets/149179506/5bb45a92-3439-4706-9e53-27a79fba9d4b)

**Third problem**:
Additionally, it seems that some cli functions are not correctly declared to be used such as the certificate rotate function which does not appear when calling help. Isn't the call to "@cli.command()" missing for certain cli functions?

![image](https://github.com/Netflix/lemur/assets/149179506/4b47424c-3d78-4cf9-ac21-5b0300fdc7a9)

Best regards

Marc

ilyakruchinin:
I'm having the exact same issue, Ubuntu 22.04

Alexa-ia:
The same issue on Ubuntu 22.04 when trying to run lemur source sync, adding 
`from flask.cli import with_appcontext
@with_appcontext
`

in lemur/sources/cli.py solved the issue

andreisolo0:
The same issue on Ubuntu 22.04 when trying to run lemur source sync, adding
`from flask.cli import with_appcontext `
`@with_appcontext `

in lemur/sources/cli.py solved the issue

PR #4969, opened 2024-09-30
Fix "Working outside of application context" error, plus fix incorrect arg ordering for rotate by serial number

jtschladen:


jtschladen:
Relates to #4764; this should correct the missing `@with_appcontext` issue, but not other issues reported there.

charhate:
I think there is one more file that use the fix - https://github.com/Netflix/lemur/blob/main/lemur/api_keys/cli.py

charhate (approved):

PR #4970, opened 2024-09-30
Fix rotate command and add api_keys_cli

jtschladen:


jtschladen:
Additional fixes related to #4764

charhate (approved):

PR #5066, opened 2025-01-13
Convert CSR to string

jtschladen:


lemur/plugins/lemur_acme/plugin.py
```diff
@@ -192,7 +191,10 @@ def get_ordered_certificates(self, pending_certs):
                         self.acme.autodetect_dns_providers(domain)
 
                 try:
-                    order = acme_client.new_order(pending_cert.csr)
+                    csr = pending_cert.csr
+                    if isinstance(csr, str):
```
jtschladen, line 195: Replicates this logic: https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/crypto.py#L2420

To compensate for this change: https://github.com/certbot/certbot/commit/5dd898f56bc5fa9eded1926aba4cfaaee5ebe6fd#diff-89c4a6402b1423e96922f4964ff7c878b99441e22e1f9579f0b58590ded05bf3L124

hosseinsh (approved):

PR #5192, opened 2025-05-22
various fixups to disable autorotate task

jmcrawford45:


lemur/certificates/service.py
```diff
@@ -194,7 +194,6 @@ def get_all_certs_not_attached_to_endpoint_with_autorotate():
         .filter(Certificate.rotation == true())
         .filter(Certificate.revoked == false())
         .filter(Certificate.not_after >= arrow.now())
-        .filter(not_(Certificate.replaced.any()))
```
jtschladen, line 197: Why remove this?

jtschladen (approved):

lemur/certificates/service.py
jmcrawford45: This should be an improvement as it's confusing that we don't turn it off after it's already reissued. Also we still maintain audibility of manual vs automated reissuance via the description and replacement relation

PR #5284, opened 2025-08-27
Bump reqs

jmcrawford45:
trying this out as a way to optimize dependabot runtime. it probably won't work, in which case we may have to explore grouping


jtschladen (approved):

jmcrawford45:
trying #5285 as an alternative to optimize dependabot

PR #5312, opened 2025-09-16
Modernise python setup process (fix #5309)

### LINKED ISSUE #5309: Legacy editable install of lemur==1.3.dev0 Deprecation warning during installation
While following the installation steps described in [the doc](https://lemur.readthedocs.io/en/latest/quickstart/index.html#setting-up-an-environment) i get this warning:

```
$ make release
[...]
Installing collected packages: lemur
  Attempting uninstall: lemur
    Found existing installation: lemur 1.3.dev0
    Uninstalling lemur-1.3.dev0:
      Successfully uninstalled lemur-1.3.dev0
  DEPRECATION: Legacy editable install of lemur==1.3.dev0 from file:///Users/gijsm/Github/Netflix/lemur (setup.py develop) is deprecated. pip 25.3 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for lemur
Successfully installed lemur-1.3.dev0
```

This is on macos 15.7 with Python 3.13 installed with homebrew.

gijzelaerr:
this replaces the old setup.py with the modern pyproject.toml file.

The original setup.py had custom build commands for static assets. I created a build_backend.py file that could handle this, but for simplicity, the current solution uses the standard setuptools backend. The static asset building can still be done manually with npm run build_static or through the Makefile.


jmcrawford45 (approved):

gijzelaerr:
the problem seems to be that flake8 doesn't read the pyproject.toml config section. Installing [flake8-pyproject ](https://pypi.org/project/Flake8-pyproject/) solves that problem. But we can also switch to something like ruff

PR #5328, opened 2025-12-05
Evan/source dest

evan-datadog:
creating sources from dests always

also creating cli tool to recreate all sources.


jmcrawford45 (changes requested):
I'm assuming either this was opened against the wrong upstream or with a bad merge base

evan-datadog:
Yes sorry, thought I closed it.

PR #5332, opened 2025-12-10
Telegram notification plugin

fed1337:
This is very similar to the slack notification plugin. Tests are passed in a slightly different environment with python 3.10

### DIFF
--- a/lemur/plugins/lemur_telegram/plugin.py
+++ b/lemur/plugins/lemur_telegram/plugin.py
@@ -1,0 +1,109 @@
+"""
+.. module: lemur.plugins.lemur_telegram.plugin
+    :platform: Unix
+    :copyright: (c) 2025 by Fedor S
+    :license: Apache, see LICENSE for more details.
+
+.. moduleauthor:: Fedor S <fedorwww@gmail.com>
+"""
+import arrow
+import requests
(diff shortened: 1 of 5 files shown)


lemur/plugins/lemur_telegram/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+        elif notification_type == "rotation":
+            attachments = create_rotation_attachments(message)
+
+        if not attachments:
+            raise Exception("Unable to create message attachments")
+
+        data = {
+            "parse_mode": "MarkdownV2",
+            "chat_id": self.get_option("chat", options),
+            "text": "*Lemur {} Notification*\n\n{}".format(notification_type.capitalize(), *attachments),
```
jmcrawford45: Looking at the structure, it should probably be:

  # For expiration (list of strings)
  "text": "*Lemur {} Notification*\n\n{}".format(
      notification_type.capitalize(),
      ''.join(attachments)  # Join the list into one string
  )

  # Or for rotation (single string)
  "text": "*Lemur {} Notification*\n\n{}".format(
      notification_type.capitalize(),
      attachments  # Already a string, no unpacking
  )
 as currently unpacking will just yield the first element

PR #5343, opened 2026-02-16
Support manual certificate renewal process

### LINKED ISSUE #5342: Support manual certificate renewal process
While experimenting with Lemur, we've recently come across a specific usecase which I don't think is currently supported. We have this legacy third-party issuer which issues a small fraction of our certificates. This issuer does not support any sort of automation, and the only way to get a cert from them is to send a CSR by email, and we unfortunately don't have the option to drop them at the moment. 

Currently, the best way to manage this issuer is to manage the CSR generation and completion outside Lemur (using openssl CLI, for example), and import the completed cert afterwards, at which point the system manages the lifecycle as usual. But the largest drawback is we have to track the CSR information either outside Lemur, or by recording the creation process in the cert's description, which is pretty poor. 

We could easily leverage the pending certificates system to deal with this specific workflow by having an issuer that creates pending certs using the normal create certificate form, then uploading the signed public key to the pending cert to complete it. This would allow Lemur to manage the CSR and the private key creation from the start.

I'm currently working on implementing this for my org and I'm willing to submit a PR here, after a little bit of additional testing

desmaraisp:
Implements and closes #5342

This PR does three things:

- Creates a new, barebones plugin. This plugin has a field to insert the associated root CA's public cert in the issuer creation form. Unlike some of the other plugins, the CA cert is a user-provided value. Once the issuer is created, the plugin simply issues a pending certificate, and it's up to the user to complete it with their own process

- Adds a new property to the base plugin class that allows async issuer plugins to bypass the acme auto resolve process. This property defaults to the current behavior, but is overriden for the new issuer to avoid errors if ACME_DISABLE_AUTORESOLVE isn't True

- Creates unit tests and documentation for the above items

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ~~~~~~~~~~~~~~~~~~~~
+Added manual issuer plugin. This plugin allows you to manage certificates issued by third party issuers that do not support automation.
 Added ENABLE_AUTOROTATION_FILTER: a configurable, plugin-independent callback that can be used to skip enabling autorotate
 based on your specific business logic. For example, you could disallow enabling autorotate on certs with notifications disabled.
 Added REISSUE_FILTER: a configurable, plugin-independent callback that can be used to reject reissuance requests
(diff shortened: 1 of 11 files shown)


desmaraisp:
There _is_ one thing I wasn't quite sure of and was hoping one of the maintainers could weigh in on. Each plugin can declare additional fields in a couple forms here and there. This PR uses one of those (public_certificate) but there doesn't seem to be a clear naming convention for those, some plugins use camel case, others snake case. Which is preferred?

PR #5415, opened 2026-04-15
Catch per-region errors in endpoint discovery to allow partial progress

jmcrawford45:
Previously a timeout or error in any region (e.g. get_all_elbs) would abort discovery for the entire account. Now each region is wrapped independently so a bad region is skipped and the rest continue.

### DIFF
--- a/lemur/plugins/lemur_aws/plugin.py
+++ b/lemur/plugins/lemur_aws/plugin.py
@@ -324,6 +324,8 @@ class AWSSourcePlugin(SourcePlugin):
                     "region": region,
                 })
                 capture_exception()
+                metrics.send("source_sync_fail", "counter", 1,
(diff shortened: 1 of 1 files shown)


lemur/plugins/lemur_aws/plugin.py
jtschladen, line 1: Should a failure in one region ultimately be reported as a failed sync? I worry that converting these failures to logs rather than metrics means we won't easily identify ongoing failures. Looking through the code, it does look like there be similar pre-existing issues, but I just don't want to lose any observability here.

jtschladen (approved):

lemur/plugins/lemur_aws/plugin.py
jmcrawford45: Added a sync fail tag, I'll leave the existing issues for a separate PR

PR #5210, opened 2025-06-06
New GCS destination plugin for uploading certificates to Google Cloud Storage

odopertchouk:
- New GCS destination plugin for uploading certificates to Google Cloud Storage
- Modular architecture with separate methods for testing and maintainability
- Comprehensive error handling for GCS API exceptions
- Support for certificate variable expansion (CN, OU, O, L, S, C)
- Wildcard certificate handling with automatic asterisk replacement
- Complete test suite
- Added google-cloud-storage dependency to requirements
- All code follows PEP 8 standards and passes lint checks

Features:

- Configurable bucket names and object paths
- Certificate and private key upload support
- Robust credential validation using GOOGLE_APPLICATION_CREDENTIALS
- Detailed logging for debugging and monitoring
- Input validation to prevent security issues

PR #5440, opened 2026-04-27
Fix PyPI release workflow to use python -m build instead of setup.py

jtschladen:


requirements-tests.txt
jtschladen, line 1: https://github.com/Netflix/lemur/pull/5210 merged with failing builds on accident, this should fix it

charhate (approved):

PR #5458, opened 2026-05-13
chore: declare contents: read on the CI workflow

arpitjain099:
CI workflow does Python lint + tests across the matrix. Three other workflows already declare per-job permissions; `ci.yml` gets a workflow-level cap as the simpler form (all jobs read-only).


arpitjain099:
Closing this for now. Thanks for the review consideration; may revisit in the future if needed.

